PR #21, opened 2026-06-21
DHS pass through basics

dsainati1:
I did a pass through Basics. Making a PR to merge in these changes with a review process.


LF/Basics.lean
```diff
@@ -292,85 +271,28 @@ BCP: Can we add some line breaks to that?  What is idiomatic?
...
+We can also ask Lean to _compile_ our definitions to efficient
 native code.
 
 Lean compiles to C, which is then compiled to machine code by a
 standard C compiler.  This facility is very useful, since it gives
 us a path from proved-correct algorithms written in Lean to
 efficient executables. We'll come back to this topic in later
 chapters.
-
-## Running Lean
```
dsainati1, line 336: Claude's comment here is valid, so I just removed this.

LF/Basics.lean
```diff
@@ -156,33 +157,38 @@ inference_ -- but we'll generally include them to make reading
...
-`.` syntax is a compromise that lets us know we're qualifying a name
-without having to type too much.
+important to avoid confusion and headaches when writing proofs.
+If we wish to be a bit more concise, we can use a syntax with just a `.`,
+like `.monday`, that lets us know we're qualifying a name
+without having to type too much. A more concise (but equivalent) definition
+of `nextWorkingDay` is given below.
 
-:::dev
-BCP: That last paragraph doesn't really explain it!  Why do we need
```
dsainati1, line 181: I tried to resolve this with a change to the previous paragraph, and just shifting to a `.`-only convention for the rest of the file.

LF/Basics.lean
```diff
@@ -577,50 +462,8 @@ GRADE_THEOREM 1: andb3_test4
 
 ::::full
 Now that we've seen how to define our own booleans, let's switch
-back to Lean's built-in `Bool` type, which has the same structure
```
dsainati1, line 580: I just cut this, since it seems most people think it's not helpful

LF/Basics.lean
```diff
@@ -1236,24 +1107,14 @@ example : odd 4 = false := by rfl
 A multi-argument recursive function.
 :::
 
-:::dev
-TODO (Claude): The sealing strategy needs to be acknowledged in the
```
dsainati1, line 1240: Did this later.

LF/Basics.lean
```diff
@@ -1421,6 +1274,104 @@ Again, we recommend stepping through these proofs in VS Code --
    changes the proof state.
 ::::
 
+## Irreducibility, Rewriting, and Proof Engineering
```
dsainati1, line 1208: This is the biggest change I think. I think we really need to nail this explanation early in order for our approach to make any sense. Let me know what you all think of this.

LF/Basics.lean
```diff
@@ -138,13 +139,13 @@ A function on days:
 ```lean
 def nextWorkingDay (d : Day) : Day :=
```
chenson2018, line 140: The way I'd usually write this is
```lean
def Day.nextWorkingDay (d : Day) : Day :=
  match d with
  | monday    => tuesday
  | tuesday   => wednesday
  | wednesday => thursday
  | thursday  => friday
  | friday    => monday
  | saturday  => monday
  | sunday    => monday
```
because inside the appropriate namespace you need neither the `.` or fully qualified name. Maybe worth putting below along with `nextWorkingDay'` since they'll see this in other places.

Edit: I see this is mentioned below, I've left a comment there.

LF/Basics.lean
```diff
@@ -245,6 +220,10 @@ handy, this would be an excellent moment to fire up VS Code with
 the Lean extension and try it for yourself.  Load this file,
 `Basics.lean`, from the book's Lean sources, find the above
 example, and observe the result in the Lean InfoView panel.)
+
+:::dev
+DHS: Where are we showing resposnes in comments? I don't see thems
```
chenson2018: They should show up in the Verso build? If not something is misconfigured.

LF/Basics.lean
```diff
@@ -725,8 +571,8 @@ inductive Color : Type where
 
 An `inductive` definition does two things:
 
-- It introduces a set of new _constructors_. E.g., `RGB.red`,
-  `Color.primary`, `true`, `false`, `Day.monday`, etc. are constructors.
+- It introduces a set of new _constructors_. E.g., `.red`,
+  `.primary`, `true`, `false`, `.monday`, etc. are constructors.
```
chenson2018: Is omitting the inductive name an improvement here? I think I'd be confused reading this the first time. I'd put

```suggestion
- It introduces a set of new _constructors_. E.g., {name}`RGB.red`,
  {name}`Color.primary`, {name}`true`, {name}`false`, {name}`Day.monday`, etc. are constructors.
```
where `{name}` ensures that when they are reading the HTML version they get hover info still.

LF/Basics.lean
```diff
@@ -736,16 +582,15 @@ to zero or more other constructors or constructor expressions,
 obeying the declared number and types of the constructor arguments.
 E.g., these are valid constructor expressions...
 
-- `RGB.red`
-- `MyBool.true`
-- `true` (Lean's builtin Boolean type -- no `.` needed)
-- `Color.primary RGB.red`
+- `.red`
+- `true`
+- `.primary .red`
```
chenson2018: Especially here, I find this confusing with multiple names in a row.

LF/Basics.lean
```diff
@@ -888,6 +734,10 @@ def RGB.myOtherBlue : RGB := myBlue
 :::dev
 BCP: Huh? We just used `myBlue` as a top-level id, didn't we?  So
 why is it unknown?
+DHS: This is super confusing! It seems like defining
+`def RGB.myOtherBlue` implicitly opens the namespace for the
```
chenson2018: I don't find it implicit (or surprising, but I'm a bad reference point for a student...), because you explicitly placed the definition inside the namespace just the same as if you'd written
```lean
namespace RGB
def myOtherBlue := ...
end RGB
```

This is a common enough pattern that I'd just explain it when you explain namespaces.

LF/Basics.lean
```diff
@@ -1421,6 +1274,104 @@ Again, we recommend stepping through these proofs in VS Code --
...
+for writing good software. You may be familiar with object oriented programming,
+for example, in which it is considered good practice not to access the
+fields of objects directly, and instead to use getter and setter methods.
+This helps to encapsulate the object's definition, so if its fields or implementation
+change, the interface it exposes to the outside world remains the same.
+
+The same principle applies to definitions (and proofs about them) in Lean.
+In idiomatic Lean, it is considered poor style to "peek" through
+definitions and rely on `rfl` to implicitly simplify expressions
+that aren't actually equal. If you take a look at the proofs of
```
chenson2018: Maybe 
```suggestion
that aren't syntactically identical. If you take a look at the proofs of
```
is less confusing?

LF/Basics.lean
```diff
@@ -1421,6 +1274,104 @@ Again, we recommend stepping through these proofs in VS Code --
...
+
+end AddPlayground
+```
+
+Each of `add_zero` and `add_succ` correspond to one branch of the `match`
+statement defining `add`, and describe how the evaluation of `add` proceeds
+in that case. The `add_zero` theorem describes how `n + 0` evaluates,
+while `add_succ` describes (symbolically) how `n + succ m` evaluates.
+Because these theorems describe how to simplify more complex expressions
+involving `add`, we call them `add`'s _simplification lemmas_.
```
chenson2018: I really like the lead-in to later `simp` use here!

LF/Basics.lean
```diff
@@ -1421,6 +1274,104 @@ Again, we recommend stepping through these proofs in VS Code --
...
+
+The same principle applies to definitions (and proofs about them) in Lean.
+In idiomatic Lean, it is considered poor style to "peek" through
+definitions and rely on `rfl` to implicitly simplify expressions
+that aren't actually equal. If you take a look at the proofs of
+`add_zero` and `add_succ` above, you will notice this is exactly what we did
+when we used the `rfl` tactic.
+
+In this text, to enforce idiomatic style, we mark
+definitions with `@[irreducible]` to prevent this peeking.
```
chenson2018: Probably helpful to give this idea its commonly used name so it's easier to talk about?
```suggestion
definitions with `@[irreducible]` to prevent this peeking, which is refereed to as *definitional equality abuse* (often shortened to *defeq abuse*).
```

chenson2018 (commented):
This looks nice to me, especially the additions about proof interface.

My only concerns were about some of the namespacing readability, but this is very nitpicky.

LF/Basics.lean
```diff
@@ -1421,6 +1274,104 @@ Again, we recommend stepping through these proofs in VS Code --
...
+The `unseal` we wrote before the proof of `add_zero` temporarily
+allows this, but only in that proof. We have to
+allow it there, since there is no other way to prove the goal.
+However, except for `add_zero` and `add_succ`, we never need
+to unseal `add` anywhere else. Instead, we can rewrite
+with those theorems anywhere we want to describe how `add` evaluates.
```
chenson2018: I think we need a little more motivation for when and why we unseal. Here's a first attempt that you are welcome to iterate on:

```suggestion
allows this, but only in that proof. We allow unsealing the definition for
these foundational lemmas, but then expect that from this point on `add_zero` and 
`add_succ` should provide a characterization of the behavior of `add` that makes 
further unsealing of the definition unnecessary.  Instead, we can rewrite
with these theorems anywhere we want to describe how `add` evaluates.
The motivation for this strict discipline is a dual concern of readability and
performance, as unfolding definitions can have negative effects as a library scales.
```

LF/Basics.lean
```diff
@@ -1869,6 +1820,14 @@ negation is involutive (that is, that negation is its own inverse).
 Another example, using booleans:
 :::
 
+:::dev
+DHS: All these `Bool` examples are in the old style, where we simplify
+through `rfl` instead of rewriting by simplification lemmas. Do we want
+to go ahead and change all of this? If we do, we need to also add in
+an explanation of how to find the appropriate rewrite lemmas in the standard library,
+which is probably a new lesson in and of itself.
+:::
```
chenson2018: So the first example would be
```lean
theorem Bool.notb_involutive : ∀ b : Bool, (!!b) = b := by
  intro b
  cases b
  · rewrite [Bool.not_false, Bool.not_true]
    rfl
  · rewrite [Bool.not_true, Bool.not_false]
    rfl
```
which seems reasonable. If you want to side step teaching them how to search for things right here, you could just tell them what theorems to use as first, like Roger did working on naturals.

LF/Basics.lean
bcpierce00: Ah.  Maybe the confusion is that we are mixing two things here: The "responses in comments" and the InfoView display.

LF/Basics.lean
```diff
@@ -1612,37 +1577,48 @@ instance : BEq Nat where
...
+one for each of the four cases of control flow through the function.
+
+```lean
+unseal beq
+theorem zero_zero_beq_true : (0 == 0) = true := by rfl
+theorem zero_succ_beq_false (n : Nat) : (0 == (succ n)) = false := by rfl
+theorem succ_zero_beq_false (n : Nat) : ((succ n) == 0) = false := by rfl
+theorem succ_succ_beq (n m : Nat) : ((succ n) == (succ m)) = (n == m) := by rfl
+seal beq
+```
```
KihongHeo, line 1526: Since these definitions are available only in the full version, the terse version does not compile.

LF/Basics.lean
```diff
@@ -952,25 +816,28 @@ namespace Playground
 ::::full
 A single constructor with multiple parameters can be used to create
```
KihongHeo, line 800: This section uses the terms "parameters" and "arguments" interchangeably. Can we pick one and stick with it? I prefer "parameters" but would be ok if consistently used.

LF/Basics.lean
KihongHeo, line 1080: Readers may get confused whether "+" below is actually "add" above. At first glance, I was also a bit confused because I am not familiar with the typeclass instance syntax in Lean.

Can we just use the full name "add" or use "notation" rather than typeclass? I am wondering whether typeclass is really needed in this part.

dsainati1: I would tend to agree, actually. IMO defining `notation` here is simpler than using a typeclass.

LF/Basics.lean
dsainati1: Good catch. Parameters should be used for the variables in the declaration of a function and arguments for the values passed to a call. I will clean this up.

LF/Basics.lean
```diff
@@ -189,6 +189,14 @@ def nextWorkingDay' (d : Day) : Day :=
 
 When full qualification is not necessary to disambiguate, we will
 prefer the shorter syntax with just the `.`.
+
+:::dev
+BCP: I'm persistently confused by this explanation.  We start out saying that
```
dsainati1, line 196: I think I am also becoming confused by how the qualifying of names work. It seems like we have four ways to refer to constructors:
- Fully qualified name (e.g. `Day.monday`) 
- Just `.` when unambiguous (e.g. `.monday`) 
- Unqualified, when the type to which the constructor belongs has been opened via an `open` (e.g., `open Day`)
- Unqualified, in a definition that has itself been qualified with the type (e.g. `def Day.nextWorkingDay`). 

Thinking about how I would actually give this explanation is making me realize though that I don't actually have a good intuition for when each of these four styles would be preferred. @chenson2018 is there a part of the Lean/Mathlib style guide to lean on here?

LF/Basics.lean
dsainati1: Oh nvm I remember why we can't do `notation`. If we try this, everywhere we use `+` Lean will complain that it cannot synthesize an instance of the add typeclass.

LF/Basics.lean
mwhicks1: I like the text that we have now. Better to introduce the real definition first, and then the usability feature second. But maybe the usability feature introduction is also lacking, per BCP-inline comment below.

LF/Basics.lean
```diff
@@ -192,51 +206,20 @@ without the abbreviation and then introduce the shorter form?
...
-:::
-
 ::::full
 Having defined a function, we should check that it works on some
-examples.  There are actually three different ways to do this in
-Lean.  First, we can use the `#eval` command to evaluate a compound
+examples.  There are a few different ways to do this in
+Lean.  One is to use the `#eval` command to evaluate a compound
 expression involving `nextWorkingDay`.  (Lean's responses are shown
 in comments.)
```
mwhicks1, line 230: Are Lean's responses shown in comments? I don't see them, below. Per BCP's question about auto-checking the comments, I wonder if there's a meta-programming feature that would also compiling this source to add in the actual responses?

LF/Basics.lean
```diff
@@ -909,6 +767,12 @@ notations.
 
 :::dev
 BCP: That goes by *waaay* too fast!
+DHS: IMO we should cut this whole section down to the same
+length as the `modules` section from the Rocq book. All students
+*really* need to know here is that if you use a name defined
+in a namespace outside of it, you have to qualify it,
+but you can open the namespace to "import" the names defined inside it.
+BCP: Yes.
```
mwhicks1: Yes,

LF/Basics.lean
```diff
@@ -952,25 +816,28 @@ namespace Playground
...
 a tuple type. As an example, consider representing the four bits in
-a nibble (half a byte). We first define a datatype `Bit` that
+a nybble (half a byte). We first define a datatype `Bit` that
 resembles `Bool` (using the constructors `b1` and `b0` for the two
-possible bit values) and then define the datatype `Nibble`, which is
+possible bit values) and then define the datatype `Nybble`, which is
 essentially a tuple of four bits.
+BCP: Nibble seems to be the preferred spelling.  (I also liike nybble, but
+a bit of looking around convinced me that it will seem wrong to enough
+people that it isn't worth it.) But I don't feel very strongly about it.
```
mwhicks1: https://en.wikipedia.org/wiki/Nibble_(magazine)

LF/Basics.lean
```diff
@@ -1109,6 +979,8 @@ This is just syntactic sugar, and the two forms are interchangeable.
 
 :::dev
 BCP: Can we be more explicit about the machinery?
+DHS: I propose we just cut this entirely? If we are working
+with our own handrolled definition, it's okay not to have notation IMO.
```
mwhicks1: I tend to agree. As a student I'd prefer to have fewer balls in the air.

LF/Basics.lean
```diff
@@ -1204,15 +1084,17 @@ Recursive functions:
...
   | succ (zero) => false
   | succ (succ n') => even n'
 
-theorem even_zero : even zero = true := rfl
-theorem even_one : even (succ zero) = false := rfl
-theorem even_succ_succ n : even (succ (succ n)) = even n := rfl
+unseal even
+example : even 1 = false  := by rfl
+example : even 4 = true := by rfl
+seal even
```
mwhicks1, line 1022: Is the use of `1` and `4` here a consequence of the notation introduction above? DHS you propose we cut this so that we'd write (even (succ zero)) and (even (succ (... zero)) instead? You might make some abbreviations to make this easier to read, in that case.

LF/Basics.lean
```diff
@@ -1236,24 +1121,14 @@ example : odd 4 = false := by rfl
...
-:::
-
 ```lean
 @[irreducible]
 def add (n : Nat) (m : Nat) : Nat :=
   match m with
   | zero => n
   | succ m' => succ (add n m')
 
+--- Don't worry about this for now
```
mwhicks1: Why would I be worried? (Who is this comment for?)

LF/Basics.lean
```diff
@@ -1421,6 +1288,104 @@ Again, we recommend stepping through these proofs in VS Code --
...
+namespace AddPlayground
+
+@[irreducible]
+def add' (n : Nat) (m : Nat) : Nat :=
+  match m with
+  | zero => n
+  | succ m' => succ (add n m')
+
+unseal add in
+theorem add_zero' : forall (n : Nat), n + 0 = n := by
```
mwhicks1: I'm a little confused about our stance on notation for our own nats. Here we are using `+` and `0` etc. Is there a line of thought that says we should be writing `add (n,zero)` instead? To me, this is actually preferred. As a beginner, I'd prefer not to have to take the one extra step of mapping notation to definitions as I'm trying to understand things.

LF/Basics.lean
```diff
@@ -1421,6 +1288,104 @@ Again, we recommend stepping through these proofs in VS Code --
...
+with those theorems anywhere we want to describe how `add` evaluates.
+
+These two theorems also follow a particular pattern. Let's look again at the
+definition of `add`:
+
+```lean
+namespace AddPlayground
+
+@[irreducible]
+def add' (n : Nat) (m : Nat) : Nat :=
```
mwhicks1: Did you include `add'` just to make the definition of `add` closer at hand, for the reader? I didn't catch this right away. I found this a little confusing; put it in a comment instead?

LF/Basics.lean
mwhicks1: Also, if you are in the playground and have repeated `add` as `add'` is it weird that you are not using `add'` ?

LF/Basics.lean
```diff
@@ -1421,6 +1288,104 @@ Again, we recommend stepping through these proofs in VS Code --
...
+unseal pred in
+theorem pred_succ n : pred (succ n) = n := by rfl
+```
+
+Similarly, for each of the three branches of the definition of `even`,
+we need one simplification lemma:
+
+```lean
+unseal even
+theorem even_zero : even zero = true := rfl
```
mwhicks1, line 1301: Here you are using `zero`, `succ` etc. instead of `0` and `1 + _` ...

LF/Basics.lean
```diff
@@ -1808,25 +1769,25 @@ Sometimes simple calculation and rewriting are not enough...
 
 :::instructors
 We use `#guard_msgs` in a number of places in the SFL
+source files to help deter bitrot, and you are encouraged to add
```
mwhicks1: Is "bitrot" a word? It looks like a misspelling of "bistrot" to me. Maybe "bit-rot" ?

LF/Basics.lean
chenson2018: Again, I would explain your last two bullets identically: it is because you have access to the `Day` namespace.

I don't recall a style guide entry; I think there's some personal preference allowed. If I think about what I usually do the checklist is something like:
- anytime I define theorems or definitions about a type, it goes under that namespace and use unqualified names
- use the leading dot notation mostly in proofs, where the expected type is used to disambiguate
- otherwise use the full name

LF/Basics.lean
```diff
@@ -2045,6 +2010,7 @@ functions that are not required to terminate.
 
 :::dev
 BCP: That last paragraph will be opaque to many readers.
+DHS: Let's just drop it, IMO
```
mwhicks1: 👍

mwhicks1 (commented):
I like the changes, especially the explanation around irreducible and seal/unseal. The main cruftiness in my mind is still around the use of notation, or not. I'm against introducing notation for our own nats here, esp to keep them distinct in readers' minds from the real nats, and to focus on understanding what's under the covers. Left a few comments on places where we are inconsistent about notation use (should we choose to keep it).

LF/Basics.lean
dsainati1: > it is because you have access to the Day namespace

I see. We should move this explanation to the namespaces section later, I think maybe that is what is missing here.

LF/Basics.lean
```diff
@@ -387,31 +317,14 @@ Another familiar enumerated type:
...
-not namespace declarations -- `notb`, `andb`, etc. land at root scope
-and avoid clashes only because their names differ.  The correct
-section/namespace distinction is taught later in this very file, so a
-careful reader will notice the contradiction.  Fix the prose (or
-actually use `namespace`).
-:::
-
 ```lean
-section MyBool
+namespace MyBool
```
chenson2018: This needs to be moved down after the definition `MyBool`, because currently you're defining `MyBool.MyBool`.

chenson2018: @dsainati1 I see the way you resolved this was to remove the namespace completely, I'm not sure I agree. I thought it was fine except for this one definition.

dsainati1: Adding it back

chenson2018 (approved):
I think all my comments have been addressed, so approving from my end.

LF/Basics.lean
```diff
@@ -1520,14 +1470,16 @@ but not for `Nat.ble`...
 BCP: Readers may wonder what "instance" is...
 :::
 
+:::dev
+DHS: I removed the other notation in this chapter, but IMO the
+lemmas and proofs below become hideously ugly to the point of unreadability
+without the `==` shorthand. I think in this instance introducing the notation
+is worth it.
+:::
```
mwhicks1, line 1502: Sounds good.

LF/Basics.lean
mwhicks1: Still missing this part (responses in comments) ?

LF/Basics.lean
```diff
@@ -196,6 +198,7 @@ being explicit is very important, but then we say that the shorter syntax
 is fine.  I feel like this is an instance where trying not to overwhelm readers
 with too much detail is actually leading to more confusion.  Maybe we should
 just explain what's going on.
+DHS: How's this?
```
mwhicks1, line 201: it seems that the full explanation comes later? I'm still not sure I get what's being said here. I think it's maybe worth being systematic, right here, if it seems to slow you down.

dsainati1: I worry that explaining namespaces right here is too soon. This is essentially the very first thing we explain in the whole textbook, and going directly in to the details of how namespaces work immediately seems like a lot. Does it seem appropriate to say "names have to be qualified" here and then explain namespaces later?

mwhicks1: This is a key concept as the current comment explains. I'm a little inclined to put it all here. But I'm good with whatever for now. We can always change it later.

LF/Basics.lean
bcpierce00: Think of what specifically?

bcpierce00: Oh, this whole section I guess...

PR #35, opened 2026-06-25
Put basics into the prelude

dsainati1:
By adding the `prelude` to the top of `Basics`, we can avoid importing anything from Lean's standard library. This lets us define notations for `+` and `*` that don't conflict with the stdlib ones.

### DIFF
--- a/LF/Basics.lean
+++ b/LF/Basics.lean
@@ -1,3 +1,4 @@
+prelude
 import VersoManual
 import VersoManual.InlineLean
 import Illuminate
@@ -1058,6 +1059,15 @@ def add (n : Nat) (m : Nat) : Nat :=
   | succ m' => succ (add n m')
 ```
 
+::::full
+We can also define infix notation for our `add` functions.
+Don't worry too much about how this is defined, we will return to it
+in more detail later.
+::::
+```lean
+scoped infixl:65 " + " => add
+```
+
 # Proof by Rewriting
 
 ## Proving properties about functions in Lean
@@ -1071,7 +1081,7 @@ behavior.
 
 Here is a simple rule about `add`:
 
-- `add n zero = n`
+- `n + zero = n`
 
 In Lean, this rule looks like this:
 ::::
@@ -1083,7 +1093,7 @@ beginning of the proof?  Can we comment on this?
 
 ```lean
 unseal add in
-theorem add_zero : ∀ n : Nat, add n zero = n := by
+theorem add_zero : ∀ n : Nat, n + zero = n := by
   intro n
   rfl
 ```
@@ -1113,7 +1123,7 @@ We can then use the `add_zero` rule to carry out a simple proof
 about natural numbers!
 
 ```lean
-theorem add_zero_zero (n : Nat) : add (add n zero) zero = n := by
+theorem add_zero_zero (n : Nat) : n + zero + zero = n := by
   rewrite [add_zero]
   rewrite [add_zero]
   rfl
@@ -1200,7 +1210,7 @@ Here it is in Lean:
 
 ```lean
 unseal add in
-theorem add_succ : ∀ n m : Nat, add n (succ m) = succ (add n m) := by
+theorem add_succ : ∀ n m : Nat, n + (succ m) = succ (n + m) := by
   intro n m
   rfl
 ```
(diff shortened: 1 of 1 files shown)


chenson2018:
Is there any danger of this behaving oddly when imported into other chapters?

dsainati1:
I think once we reach `UsingLean.lean` we are going to not rely on anything from `Basics` or `Induction`, since we are going to switch over to Lean's Nats after that point. So we may not even need to import them into anything later, tbh.

chenson2018 (approved):
If this seems to be behaving as expected and won't be imported elsewhere, no objection from me.

xhalo32:
This is looking really good now! I have reviewed it locally and tried a couple of the exercises and they seem to work as expected. I reviewed did this by checking out to the branch, running `make serve`, and opening `_out/lf/student/lean/LF/Basics.lean` in VSCodium.

Is there a reason that `add_zero_zero_explained` and `add_zero_zero_zero` use `add` literally? The infoview in those proofs now shows the `+` notation making ``The goal is `add (add n zero) zero = n`.`` on line 702 (of the generated lean file) incorrect.

Another thing we should probably explain if how to hover over the terms in the infoview to see their structure. This is especially relevant now that we use infix notation.

`mul_succ`, `add_mul_zero`, `binToNat_b0` (and other nearby declarations) still use `add` and/or `mul` rather than the infix syntax. The prose around `General Proofs about Natural Numbers` still refers to `add` in the infoview rather than using the syntax.

Sorry for taking too long to comment as the PR is already merged!

dsainati1:
Any remaining uses of `add` and `mul` directly are oversights on my part. I will fix them in https://github.com/plclub/sf-in-lean/pull/36 since I still have the lock on basics. 

As for hovering in the info view to see term structure, I agree we should add this here. Once https://github.com/plclub/sf-in-lean/pull/36 is merged, do you want to open a PR in basics to add this?

xhalo32:
Sounds good, I can do that -- I anyways have a couple of `:::dev` comments to add to Basics

dsainati1:
Actually, feel free to just pull the branch that PR is on and add your comments to basics as part of that PR, I am done editing for the moment.

xhalo32:
I need a bit more time to organize my notes and am anyways finishing for today, so no need to wait for me. Thanks for the suggestion though!

rogerburtonpatel:
> I think once we reach `UsingLean.lean` we are going to not rely on anything from `Basics` or `Induction`, since we are going to switch over to Lean's Nats after that point. So we may not even need to import them into anything later, tbh.

That's correct!

xhalo32:
Somehow I totally missed the fact that the generated lean files under `_out/lf/*/lean/LF/Basics.lean` still import the prelude, i.e. don't start with `prelude`. Was it on purpose that only the verso document omits the prelude, or should the handout omit it too?

I also totally overlooked the fact that adding a `#check HAdd` or `#check 1 + 1` in any of the lean blocks still works (even though my understanding was that it should not work after this PR). `HAdd` and the `+` notation (as well as `HMul` and `*`) are transitively imported by the verso imports even though Basics.lean has `prelude` at top. Can someone clarify what was the problem that we were trying to solve by adding `prelude` at the top? (Removing it has no effect based on my testing)

As the `prelude` command doesn't have any effect on Basics.lean, maybe we can just remove it?

P.S. Adding `prelude` to the generated assignment leads to ~100-200 errors, some of which are resolved by importing Init.Prelude. However, it seems like the `example` keyword (among other things) is not registered before importing `Init.Notation` which also imports the `+` notation.

dsainati1:
Huh. That's very odd. I was under the impression that adding the `prelude` keyword to the top of the file would result in the prelude not being imported, which would mean that things like `HAdd` and its associated notation would not be imported by the file. The goal was this would allow us to define our own `+` temporarily for this file without conflicting with the `HAdd` typeclass, since we aren't ready to get into that material yet.

chenson2018:
I'm too surprised here, using `prelude` outside of core Lean is always a bit hacky.

xhalo32:
Your impression is correct, but I think it also requires that every other transitive import in the file must use `prelude` at the top, but the Verso modules don't use the `prelude` command, i.e. the prelude ends up imported.

I don't think there is a way to tell Lean explicitly to not import `prelude` even if some imports do import it transitively.

PR #36, opened 2026-06-25
DHS pass over induction

dsainati1:
Did a pass over induction. Cleaned up some of the definitions, switched back to using notation for `+` and `*` now that we found a way to do it that isn't confusing, and more rigorously enforced the irreducible rewriting discipline for this chapter. `UsingLean.lean` is next, and we can begin to relax it there.


LF/Induction.lean
```diff
@@ -1160,28 +1171,37 @@ example : ∀ b, natToBin (binToNat b) = b := by sorry
...
 
+@[irreducible]
 def doubleBin (b : Bin) : Bin :=
   -- ADMITDEF
   match b with
   | .z => .z
   | _ => .b0 b
   -- /ADMITDEF
 
+-- TODO (DHS): How to hide these theorem statements so that students can get practice writing them?
```
dsainati1, line 1192: Highlighting this TODO comment for reviewer attention.

Writing the simp lemmas out exactly specifies the behavior of the function, which is the point, but also gives the game away if we show them to the students. Can we hide them somehow as solutions?

LF/Induction.lean
```diff
@@ -1074,20 +1077,20 @@ def binToNat (m : Bin) : Nat
   to make the property easier to prove, feel free to do so!
 -/
 
-/- SOONER (DHS): This is a very-category theoretic way to present
-   this idea. Is this the most useful way to convey this to
-   an audience who is presumably unfamiliar with commutative diagrams? -/
-
+/- TODO (DHS): The printing in the lean info view here is really odd:
```
dsainati1: Highlighting this comment for reviewer attention.

I think it's basically mandatory that we figure out how to fix this, it's extremely confusing otherwise.

KihongHeo:
I'd be happy to review also. 

BTW, could you also make the induction chapter available via the Verso-generated webpage?
It will be nice if we can review both the source code and webpage together.

LF/Induction.lean
chenson2018: If you do `attribute [pp_nodot] Bin.b0 Bin.b1` this goal displays as
```lean
binToNat (incr (Bin.b1 b')) = binToNat (Bin.b1 b') + one
```

Does that help?

LF/Induction.lean
```diff
@@ -176,42 +168,55 @@ theorem review4 : ∀ n : Nat, add n zero = n := by
 -- /HIDE
 -- /QUIZ
 
+/- TODO (DHS): We use this theorem later,
+   so let's make it into a review exercise here -/
+/- review5 -/
+/- Prove the following theorem, using theorems from Basics: -/
+theorem succ_eq_add_one : ∀ n : Nat, succ n = add n one := by
```
chenson2018: At what point do we switch over to the more idiomatic

```suggestion
theorem succ_eq_add_one (n : Nat) : succ n = add n one := by
```
? I am thinking it should happen in this chapter. (Can be a follow-up PR to not muddy the diff)

LF/Induction.lean
```diff
@@ -867,43 +875,36 @@ theorem mul_two : ∀ p : Nat,
...
   case succ m ih =>
-    calc mul (succ n) (succ m)
-    _ = add (mul (succ n) m) (succ n)       := by rw [mul_succ]
-    _ = add (add (mul n m) m) (succ n)      := by rw [ih]
-    _ = succ (add (add (mul n m) m) n)      := by rw [add_succ]
-    _ = succ (add n (add (mul n m) m))      := by rw [add_comm _ n]
-    _ = succ (add (add n (mul n m)) m)      := by rw [add_assoc n _ m]
-    _ = succ (add (add (mul n m) n) m)      := by rw [add_comm n]
-    _ = add (mul n (succ m)) (succ m)       := by rw [mul_succ, add_succ]
+    rw [mul_succ, ih, add_succ, add_comm _ n,
```
chenson2018, line 909: We are still preferring `rewrite`?
```suggestion
    rewrite [mul_succ, ih, add_succ, add_comm _ n,
```

LF/Induction.lean
```diff
@@ -1040,20 +1025,38 @@ def incr (m : Bin) : Bin
...
+seal binToNat
+
 /-
   In Basics, we did some unit testing of `binToNat`, but we
   didn't prove its correctness. Now we'll do so.
 -/
 
 -- EX3! (binary_commute)
+
+/- SOONER (DHS): This is a very-category theoretic way to present
```
chenson2018, line 1055: My anecdotal experience makes me think this is fine. I first did this exercise before I knew any category theory and found the diagram intuitive.

LF/Induction.lean
```diff
@@ -1097,15 +1100,23 @@ theorem bin_to_nat_pres_incr : ∀ b : Bin,
...
+-/
+@[irreducible]
 def natToBin (n : Nat) : Bin :=
   -- ADMITDEF
   match n with
   | zero => .z
   | succ n' => incr (natToBin n')
   -- /ADMITDEF
 
+-- TODO (DHS): How to hide these theorem statements so that students can get practice writing them?
```
chenson2018, line 1107: David set it up so that if you put:
```
-- SOLUTION
-- END SOLUTION
```
in an exercise that it will turn into `-- FILL IN HERE` in both student version of the Lean files and the generated HTML.

LF/Induction.lean
```diff
@@ -1160,28 +1171,37 @@ example : ∀ b, natToBin (binToNat b) = b := by sorry
...
 theorem double_incr : ∀ n : Nat,
-    double (succ n) = add (double n) two := by
+    double (succ n) = (double n) + two := by
   -- ADMITTED
   intro n
   rw [double_succ]
   rw [two_eq_succ_one, one_eq_succ_zero, add_succ, add_succ, add_zero]
 -- /ADMITTED
--- GRADE_THEOREM 0.5: double_incr
+-- GRADE_THEOREM zero.5: double_incr
```
chenson2018: Stray find and replace?

LF/Induction.lean
KihongHeo: `+` instead of `add`?

LF/Induction.lean
KihongHeo: (same here) `+` instead of `add`

LF/Induction.lean
```diff
@@ -364,29 +367,29 @@ theorem eqb_self : ∀ n : Nat,
...
--- GRADE_THEOREM 0.5: mul_0_l
+-- GRADE_THEOREM zero.5: mul_zero_l
 
 theorem succ_add : ∀ n m : Nat,
-    add (succ n) m = succ (add n m) := by
+    (succ n) + m = succ (n + m) := by
   -- ADMITTED
   intro n m
   induction m
   case zero => rewrite [add_zero, add_zero]
```
KihongHeo: formatting? Either one line with `;` or multiple lines with a newline first.

LF/Induction.lean
```diff
@@ -410,20 +413,20 @@ theorem add_comm : ∀ n m : Nat,
...
   induction p
   case zero => rewrite [add_zero, add_zero]
                rfl
   case succ p' ih =>
-    rewrite [add_succ m p', add_succ n (add m p'), add_succ (add n m) p', ih]
+    rewrite [add_succ, add_succ, add_succ, ih]
     rfl
 -- /ADMITTED
--- GRADE_THEOREM 0.5: add_assoc
+-- GRADE_THEOREM zero.5: add_assoc
```
KihongHeo: Maybe an unintended effect of find-and-replace?

LF/Induction.lean
ionathanch: Maybe we should turn `pp.fieldNotation` off? It's kind of distracting in general unless it's actually a structure projection

chenson2018: I think I'd still prefer turning it off on a case by case basis. It's pretty nice for readability some later chapters when they chain together.

bcpierce00:
> I'd be happy to review also.
> 
> BTW, could you also make the induction chapter available via the Verso-generated webpage? It will be nice if we can review both the source code and webpage together.

This requires converting the file over to Verso format, which I will do as soon as it compiles cleanly.

LF/Induction.lean
rogerburtonpatel: There's a paragraph of introduction and explanation for `rw` when we switch to it. So before/after that.

LF/Induction.lean
bcpierce00: Could they be moved later so that at least the reader has the chance to do the exercise before encountering them?

bcpierce00 (approved):
It all looks good, modulo a few nits that I'll add to the PR commits and a couple of ongoing discussions.

bcpierce00:
I'd like to get this PR merged as soon as feasible so that I can move ahead with translating the file to Verso.  If there are ongoing discussions, let's move them from this issue into comments in the code / Zulip...

PR #37, opened 2026-06-25
Use the existing machinery for generalization from Lean.Meta.Tactic.Cases instead of rolling our own

ionathanch:
This fixes failures to generalize indices of targets depended upon by the goal by also generalizing the target with a heterogeneous equality. By design, `inversion` may leave behind this HEq if it cannot be solved away, such as in the (fake but illustrative) proof for `Wec`.

### DIFF
--- a/LF/CustomTactics.lean
+++ b/LF/CustomTactics.lean
@@ -94,7 +94,7 @@ partial def unifyEqs (eqs : List FVarId) (mvarId : MVarId) (subst : FVarSubst) (
   | eq :: eqs =>
     let some { mvarId, subst, numNewEqs } ← Option.join <$> (observing? $ unifyEq? mvarId eq subst MVarId.acyclic caseName?)
       | return (mvarId, subst)
-    let (newEqs, mvarId) ← mvarId.introNP numNewEqs
+    let (newEqs, mvarId) ← mvarId.introN numNewEqs
     let eqs := eqs.map (Expr.fvarId! ∘ subst.get)
     unifyEqs (newEqs.toList ++ eqs) mvarId subst caseName?
 
@@ -160,7 +160,7 @@ def inversionCore (h : FVarId) (config : InversionConfig) : TacticM (List MVarId
     let some ctx ← mkCasesContext? h
     | throwTacticEx `cases goal "not applicable to the given hypothesis"
     let gis@⟨newGoal, _, newTarget, numEqs⟩ ← generalizeIndices goal h
-    let (newEqs, newGoal) ← newGoal.introNP numEqs
+    let (newEqs, newGoal) ← newGoal.introN numEqs
     let subgoals ← inductionCasesOn newGoal newTarget default ctx
     let subgoals ← elimAuxIndices gis subgoals
     let subgoals ← unifyCasesEqs newEqs.toList subgoals
@@ -217,12 +217,14 @@ end -- meta section
 
 namespace Tests
 
-example (f : Nat → Nat) (n : Nat) (le : f n ≤ 0) : 0 = f n := by
-  -- cases le /- Dependent elimination failed: Failed to solve equation 0 = f n -/
-  inversion le; exact h
+set_option pp.proofs true
 
 example (f : Nat → Nat) (n : Nat) (le : f n ≤ 0) : 0 = f n := by
-  inversion +clear le; exact h
+  -- cases le /- Dependent elimination failed: Failed to solve equation 0 = f n -/
+  inversion le; assumption
+
+example (f : Nat → Nat) (n : Nat) (le : f n ≤ 0) : 0 = f n := by
+  inversion +clear le; assumption
 
 /-- warning: declaration uses `sorry` -/
 #guard_msgs(warning) in
--- a/LF/CustomTactics.lean
+++ b/LF/CustomTactics.lean
@@ -30,8 +30,10 @@ syntax (" | " caseArg)+ " => " tacticSeq : invAlts
 /--
   `inversion t` generalizes nonvariable indices of the type of `t` before invoking `cases t`,
   then solves away contradictory generated goals.
-  * If `inversion +clear t` is set, `t` is `clear`ed from the context.
-    By default, `t` will remain in the context.
+  `t` itself will also be generalized with a heterogeneous equality `heq`
+  if trying to generalize the indices produces an ill-typed result.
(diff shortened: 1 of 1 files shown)


berberman (commented):
Thanks for this nice improvement. It seems that the behavior is very much aligned to `cases` now, and I no longer see obvious example of `cases` succeeds while `inversion` fails.

One small UX issue I ran into is alts name handling. For example, in

```lean
inductive Foo : Nat → Prop where
  | mk : Foo 0

example (h : Foo 0) : True := by
  inversion h with
  | mk => sorry
```

`inversion` fails with `Invalid alternative name mk: Expected [anonymous]`. The issue is that we resolve alts using mvar tags instead of constructor names (we discard `ctorName` in `CasesSubgoal`.) It works in some cases by coincidence (when the goal tag happens to be the same as ctor name, e.g. `cons`).

I think we have a design choice here: we're using meta cases API and not reusing elaborator's alt handling (`evalAlts`). In order to get this to work, either we need to duplicate a small part of that name normalization logic (e.g. stripping prefixes) locally after saving all `ctorName`s, or somehow reuse `Alt` construction in `ElimApp`. IMO the latter approach would be harder.

xhalo32:
While testing the PR locally noticed the following (possibly a regression?) that is not present in the parent of this commit (a06ac5b6).

In the first examples/tests that were in a06ac5b6, the infoView now shows that a heterogenous equality is introduced after the inversion tactic. E.g. after the second example's `inversion +clear le` it shows
```
f : Nat → Nat
n : Nat
le : f n ≤ 0
h✝ : le ≍ ⋯
h : 0 = f n
⊢ 0 = f n
```

Also. I think we should be using `guard_hyp` more in the tests. For example, if the new heterogenous equality is desired, we should test for it:

```
example (f : Nat → Nat) (n : Nat) (le : f n ≤ 0) : 0 = f n := by
  inversion +clear le
  rename_i hle
  guard_hyp hle : le ≍ Nat.le.refl
  guard_hyp h : 0 = f n
  exact h
```

Update: to learn more about this (and metaprogramming in general) I managed to erase all heterogenous equalities from the context of each of the `subgoals`. As a beginner in metaprogramming, my solution is pretty ugly, but tell me if you want me to share it.

ionathanch:
A few things:
* In hindsight, I think the default behaviour of `inversion` being to retain the thing being inverted and `+clear` not retaining it is counterintuitive to existing tactics like `cases` and `induction`
* I'm not sure the default behaviour should be to clear away any new equations, including the unresolved heterogeneous ones, since the purpose of inversion is to generate those equations that `cases` might have otherwise failed on
* However, such heterogeneous equations are a little useless since you can't rewrite by them (such as in the `Wec` example), and they really *are* useless if nothing actually depends on the target (such as in the `le` example)

Therefore I think it would could expand the `+clear` mechanism to instead clear away the target and the heterogeneous equation on it if nothing else depends on the target, and clear away only the heterogeneous equation if the target is depended upon. Does that sound reasonable?

(Tangentially, I also noticed that in the `le` example, `h` is introduced as an accessible name, which is wrong. I'll push a fix for that shortly.)

ionathanch:
> One small UX issue I ran into is alts name handling. For example, in
> 
> ```lean
> inductive Foo : Nat → Prop where
>   | mk : Foo 0
> 
> example (h : Foo 0) : True := by
>   inversion h with
>   | mk => sorry
> ```
> 
> `inversion` fails with `Invalid alternative name mk: Expected [anonymous]`. The issue is that we resolve alts using mvar tags instead of constructor names (we discard `ctorName` in `CasesSubgoal`.) It works in some cases by coincidence (when the goal tag happens to be the same as ctor name, e.g. `cons`).
> 
> I think we have a design choice here: we're using meta cases API and not reusing elaborator's alt handling (`evalAlts`). In order to get this to work, either we need to duplicate a small part of that name normalization logic (e.g. stripping prefixes) locally after saving all `ctorName`s, or somehow reuse `Alt` construction in `ElimApp`. IMO the latter approach would be harder.

I've opened a new issue for this at #40 so that it can be handled in a separate PR

xhalo32:
The new changes address the problem I pointed out in a satisfactory way!

I wonder if there's a way to also test that `+clear` actually works as expected. For example the following very inelegant guard technically catches if `inversion +clear` doesn't clear `leq`:

```
/--
error: Unknown identifier `leq`
---
error: (kernel) declaration has metavariables '_private.LF.CustomTactics.0.Tests._example'
-/
#guard_msgs(error) in
example (f : Nat → Nat) (n : Nat) (leq : f n ≤ 0) : 0 = f n := by
  inversion +clear leq with
  | refl eq =>
    have := leq
    assumption
```

ionathanch:
Okay this is _still_ generating an accessible name for the HEq:

```lean4
example (f : Nat → Nat) (n : Nat) (leq : f n ≤ 0) : 0 = f n := by
  inversion leq -- h✝¹ : leq ≍ le.refl
  assumption
```

Something weird is going on with introduction, since introducing something else makes the previous HEq accessible all of a sudden:

```lean4
-- before intro
[debug] case refl
f : Nat → Nat
n : Nat
leq : f n ≤ 0
h✝ : leq ≍ le.refl   -- this is inaccessible
⊢ 0 = f n → 0 = f n

-- after intro
[debug] case refl
f : Nat → Nat
n : Nat
leq : f n ≤ 0
h✝¹ : leq ≍ le.refl  -- this is accessible
h✝ : 0 = f n         -- this is inaccessible
⊢ 0 = f n
```

Not sure what's going on here; maybe there's a way to simply force fvars to be inaccessible after the fact?

ionathanch:
> I wonder if there's a way to also test that `+clear` actually works as expected. For example the following very inelegant guard technically catches if `inversion +clear` doesn't clear `leq`:
> 
> ```
> /--
> error: Unknown identifier `leq`
> ---
> error: (kernel) declaration has metavariables '_private.LF.CustomTactics.0.Tests._example'
> -/
> #guard_msgs(error) in
> example (f : Nat → Nat) (n : Nat) (leq : f n ≤ 0) : 0 = f n := by
>   inversion +clear leq with
>   | refl eq =>
>     have := leq
>     assumption
> ```

I mean to post this before pushing lol. We can just reuse `guard_hyp`:

```lean4
/-- error: Unknown identifier `leq` -/
#guard_msgs(error) in
example (f : Nat → Nat) (n : Nat) (leq : f n ≤ 0) : 0 = f n := by
  inversion +clear leq
  guard_hyp leq
```

PR #41, opened 2026-06-26
DHS pass over UsingLean.lean

dsainati1:
Mostly wrote a new section about searching for theorems with `exact?` and `rw?`

### DIFF
--- a/LF/UsingLean.lean
+++ b/LF/UsingLean.lean
@@ -109,7 +109,7 @@ theorem complicated_computation : (2 * 3 + 4 * 5 : Nat) * 6 = 156 := by
   As part of using Lean's standard `Nat` type, we will also begin
   using theorems about `Nat`s from the standard library. Because we
   did not write or prove these theorems ourselves, however, we may not
-  know with all the available theorems off the top of our heads.
+  know all the available theorems off the top of our heads.
 
   Lean provides a few ways to search through the standard library to find theorems
   that may be useful during a particular proof. The first such way is the `exact?`
--- a/LF/UsingLean.lean
+++ b/LF/UsingLean.lean
@@ -136,7 +136,7 @@ example (a b : Nat) : a + b = b + a := by
   close the goal. You can click on this button to replace the
   occurence of `exact?` with this tactic; idiomatic Lean does not
   leave `exact?` tactics (or any other `?` tactics, as we will see
-  shortly) in the finished versions of proofs and to instead replaces
+  shortly) in the finished versions of proofs and instead replaces
   them with the suggested tactics.
 -/
 -- /FULL
@@ -320,7 +320,7 @@ example (n m : Nat) (h : 2 * n = m * 2) : n + n = m + m := by
 example (n : Nat) : square n + 0 = n * n := by
   dsimp [square]
 
-/- Like `rw` and `exact`, `dsimp` also has a `?` mark version that searches for
+/- Like `rw` and `exact`, `dsimp` also has a `?` version that searches for
    functions to simplify by. Many Lean tactics have `?` versions; try it out if you are unsure. -/
 
 /-
--- a/LF/UsingLean.lean
+++ b/LF/UsingLean.lean
@@ -41,13 +41,11 @@ import LF.Induction
 section long_example
 open NatPlayground.Nat
 /- Previously, we did computation like this... -/
-theorem test_mult1' : (3 * 3 : NatPlayground.Nat) = 9 := by
-  rewrite [three_eq_succ_two, two_eq_succ_one, one_eq_succ_zero]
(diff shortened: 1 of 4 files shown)


LF/UsingLean.lean
```diff
@@ -103,6 +99,114 @@ theorem complicated_computation : (2 * 3 + 4 * 5 : Nat) * 6 = 156 := by
...
+  # Using the Standard Library
+-/
+
+-- FULL
+/-
+  As part of using Lean's standard `Nat` type, we will also begin
+  using theorems about `Nat`s from the standard library. Because we
+  did not write or prove these theorems ourselves, however, we may not
+  know with all the available theorems off the top of our heads.
+
```
bcpierce00, line 113: Who is "we" there?

bcpierce00 (commented):
Just a quick comment for now -- I'll look at this chapter carefully when Induction is under control.

LF/UsingLean.lean
dsainati1: Readers, I suppose? This seemed like the same kind of grammatical pattern as saying "we prove this theorem" that we are using throughout the book, but maybe I'm wrong.

LF/UsingLean.lean
```diff
@@ -103,6 +99,114 @@ theorem complicated_computation : (2 * 3 + 4 * 5 : Nat) * 6 = 156 := by
...
+example (a b : Nat) : a + b = b + a := by
+  /- this will suggest that we use `exact Nat.add_comm a b` to close this goal -/
+  exact?
+
+-- FULL
+/-
+  If you are using the Lean 4 extension in VSCode, the InfoView will
+  have a blue `[apply]` button that shows the suggested theorem to
+  close the goal. You can click on this button to replace the
+  occurence of `exact?` with this tactic; idiomatic Lean does not
```
KihongHeo: This is a bit confusing. I am using VScode extension version 0.0.237 but I do not see the "blue apply" button. Instead, I see a light bulb (suggestion) button above `exact?`. Since the UI may change in future versions, can we describe this a bit more generally rather than referring to a specific button?

LF/UsingLean.lean
KihongHeo: It would be better to explain what `_` means.

KihongHeo, line 240: This part is probably unclear to readers.
I suggest adding a higher-level description using a math-book style notation as follows:
```
a + (b + c)
= (a + b) + c        ...   [by add assoc]
= (b + a) + c        ...   [by add comm]
= b + (a + c)        ...   [by add assoc]
```
And then, we can explain that `calc` is used to mimic such a process.

KihongHeo: Do we need this line?
Or replace with comments? Readers may wonder what it means.

LF/UsingLean.lean
chenson2018: This UI is stable, I don't think we should avoid mentioning the specific.

I believe the confusion here is that there are two UI components to apply suggestions. The lightbulb you mention is to apply code actions in VSCode, which appears inline. Then in the infoview there is another button that appears for suggestions.

KihongHeo: I see. I missed the button because the messages were folded. It might be helpful to add a brief note about this for users like me.

LF/UsingLean.lean
```diff
@@ -241,20 +382,21 @@ def minustwo (n : Nat) : Nat :=
...
-  case zero => rfl
+  case zero =>
+    rfl
   case succ n' ih =>
-    rewrite [even, ih, NatPlayground.Nat.notb_involutive]; rfl
+    rw [even, ih, Bool.not_not]
+  -- /ADMITTED
+-- GRADE_THEOREM 1: even_succ
 
   -- TODO: talk about using `Nat.add_zero` and friends from now on.
```
rogerburtonpatel, line 401: Let's address this after we merge. It's a 2-5 liner.

rogerburtonpatel (approved):
Looks quite solid. Left a comment on a lingering TODO that's easy to fix after merge.

mwhicks1 (approved):
Seems fine to me, but I wonder about a bigger change to both UsingLean.lean and Basics.lean, as follows:
1. Proofs in Basics are as in the Rocq original, using computation and unsealed definitions. This gives a way to explain various tactics and the basics of proofs for abstractions without hand-waves.
2. When we switch to Lean Nats in UsingLean we explain that the proofs are done differently: You do not have access to the underlying Nat definitions. We show how proofs work by rewriting using the exposed rules. Finally, we tie the knot, showing how you'd implement this rule-based approach using our own Nats, basically what's in Basics now.

I think this is not something to do now. We can do it later, if we agree, since the goal is that the "API" to the rest of the books will be the same.

PR #44, opened 2026-06-29
highlight tenets; other clarifications, comments, question

mwhicks1:
I added some clarification on the use of tenets, under "Philosophy." I think this process is important for us to move fast without too much coordination slowdown. I also hopefully clarified a few other points.

Finally: I left a comment about the cadence for PRs, that is worth discussing.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,17 +70,12 @@ ways of working if you see them!
 
 We use a standard branch-and-PR workflow.  See below for details.
 
-Make PRs frequently so that your local changes get folded back into
-the `main` branch as quickly as possible.  There is no need to
-completely finish all the work on an issue before merging what you've
-done back into main. If your branch compiles and won't interfere with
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -64,32 +73,41 @@ We use a standard branch-and-PR workflow.  See below for details.
 Make PRs frequently so that your local changes get folded back into
 the `main` branch as quickly as possible.  There is no need to
 completely finish all the work on an issue before merging what you've
-done back into main. If yoour branch compiles and won't interfere with
+done back into main. If your branch compiles and won't interfere with
 what someone else is doing, please make a checkpoint PR every day or
 so.
-
-We use the github issue tracker for recording large tasks that need to
+MWH: I don't agree with this last sentence. The point of a PR, and
```
xhalo32: I agree with this point, and second the proposed alternative.

Issue #39, opened 2026-06-26
Measure how long Basics.lean takes


rogerburtonpatel:
On **Tuesday, June 30th**, Basics.lean will be ready for a pass by some of our intrepid volunteers. This will give us a new time estimate. 

We will assign the people who are involved!

bcpierce00:
What matters is how long it takes to lecture.  Closing this issue because there's no way to find out until September.

PR #45, opened 2026-06-29
Basics beefup

rogerburtonpatel:
Approaching the fix point on Basics. This is one of the- hopefully final- steps there.

### DIFF
--- a/LF/Basics.lean
+++ b/LF/Basics.lean
@@ -1549,6 +1549,12 @@ example : leb four two = false := by rfl
 seal leb
 ```
 
+We'll be using `leb` a lot, so let's give it an infix notation too.
+
+```lean
+scoped infixl:55 " <? " => leb
+```
+
 :::slidebreak
 :::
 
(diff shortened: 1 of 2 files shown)


LF/Basics.lean
```diff
@@ -28,8 +28,8 @@ file := "Basics"
 :::instructors
 This file and Induction.lean each take about an hour to
 get through in a not-too-rushed fashion (with questions, etc.).
-(BCP: Actually, in 2025 this file alone took me two full hours.)
 (BCP: This estimate may need to be revised now that we are in Lean!)
+(RAB: Issue opened to measure this.)
```
rogerburtonpatel: #39

LF/Basics.lean
```diff
@@ -195,26 +197,17 @@ def nextWorkingDay' (d : Day) : Day :=
...
-MWH - it seems that the full explanation comes later? I'm still not sure I get
-what's being said here. I think it's maybe worth being systematic, right here,
-if it seems to slow you down.
-DHS - I worry that explaining namespaces right here is too soon. This is essentially
-the very first thing we explain in the whole textbook, and going directly in to the
-details of how namespaces work immediately seems like a lot. Does it seem appropriate
-to say "names have to be qualified" here and then explain namespaces later?
-MWH - This is a key concept as the current comment explains. I'm a little inclined to
-put it all here. But I'm good with whatever for now. We can always change it later.
+RAB: I'm resolving the discussion for now. The `.pattern` syntax shows
```
rogerburtonpatel, line 209: I will remove this entire thread after is there is agreement on this.

LF/Basics.lean
```diff
@@ -242,27 +235,28 @@ in comments.)
 ```
 
 ::::full
-(We show Lean's responses in comments; if you have a computer
-handy, this would be an excellent moment to fire up VS Code with
-the Lean extension and try it for yourself.  Load this file,
-`Basics.lean`, from the book's Lean sources, find the above
-example, and observe the result in the Lean InfoView panel.)
+If you have a computer handy, this would be an excellent moment
```
rogerburtonpatel, line 256: We need to determine the medium of programming: what do we call what I dub the "Lean web interface?"

LF/Basics.lean
```diff
@@ -242,27 +235,28 @@ in comments.)
...
+to fire up VS Code with the Lean extension or the Lean web interface
+and try it for yourself.  Load this file, `Basics.lean`,
+from the book's Lean sources, find the above example, and observe
+the result in the Lean InfoView panel.
 
 :::dev
 DHS: Where are we showing responses in comments? I don't see them.
+RAB: Why did we remove the comments?
```
rogerburtonpatel, line 264: These should be inserted back in or entirely removed. Let us decide here and now.

LF/Basics.lean
```diff
@@ -323,6 +310,12 @@ efficient executables. We'll come back to this topic in later
 chapters.
 ::::
 
+:::dev
+RAB: Is Lean compiling to C its "killer app," or is it the fact that it is an
```
rogerburtonpatel, line 333: @chenson2018

LF/Basics.lean
```diff
@@ -340,12 +333,6 @@ scratch; later we'll switch to Lean's built-in `Bool`.
 We use a different name to make explicit that this is not the same
 type as Lean's built-in, but their definitions are equivalent.
 
-:::dev
-BCP: Why call it MyBool instead of just Bool?  (Or, conversely, why call the constructors
-true and false instead of mytrue, myfalse, mynotb, etc.?
-Maybe this is a place to talk about how inductive types create namespaces?)
-:::
-
```
rogerburtonpatel, line 348: I think this is something students can grok if they could grok it in the Rocq book.

LF/Basics.lean
```diff
@@ -397,10 +384,6 @@ complete specification -- a truth table -- for the `orb` function:
 Note the syntax for defining multi-argument functions (`andb` and `orb`).
 :::
 
-:::dev
-HG: This feels like too little discussion of FP function calling for a non-FP
-person, and too much for anyone who knows FP. I propose we skip this.
-:::
```
rogerburtonpatel, line 403: Personally, I think this is fine.

LF/Basics.lean
```diff
@@ -411,25 +394,21 @@ example : orb .true  .true  = .true  := by rfl
 
 We can define new symbolic notations for existing definitions.
 Because Lean already defines the same notations for the built-in `Bool`,
-we restrict ours locally to a _section_. Don't worry for now about
+we restrict ours locally to a namespace. Don't worry for now about
 how the notation is defined.
```
rogerburtonpatel, line 419: I have removed `section` from this chapter.

LF/Basics.lean
```diff
@@ -500,24 +479,6 @@ but comes with a lot of useful functions and lemmas.
...
-:::
-
-:::dev
-TODO (Claude): Concretely: even trivial drop-in exercises would keep
-hands on keyboards through this stretch -- e.g., "define `isWeekend`",
-"write a `Color` inverter", "add your own `#check` and predict the
-output".  The namespaces/sections material in particular is reference
-content that could become an exercise-light aside.
-:::
-
```
rogerburtonpatel, line 520: I have added exercises.

LF/Basics.lean
```diff
@@ -553,13 +514,6 @@ each of type `Bool`, this function produces an output of type
 `Bool`."
 ::::
 
-:::dev
-HG: Again, maybe too basic for our audience.
-BCP: I think here I might diusagree...?
-DHS: This is almost exactly the text from the Rocq book. Are we aiming
-these at different audiences?
-:::
-
```
rogerburtonpatel, line 562: Agreed and resolved.

LF/Basics.lean
```diff
@@ -700,7 +647,83 @@ This function produces the same result as the old
...
+:::grade
+```
+GRADE_THEOREM 1: is_inversion
+```
+:::
+::::
+
+
+
+## Namespaces
```
rogerburtonpatel, line 748: Two new exercises. The GRADE_THEOREM components of course need to be changed; I see now the first one is a typo. I think this is all getting deprecated by our new system.

LF/Basics.lean
```diff
@@ -809,13 +851,6 @@ possible bit values) and then define the datatype `Nibble`, which is
...
 ::::
 
-:::dev
-DHS: Some discussion remains around Nibble vs. Nybble.
-From the GitHub thread, we had MWH linking
-`https://en.wikipedia.org/wiki/Nibble_(magazine)`,
-but BCP indicated that he saw the Nibble spelling more commonly.
-:::
-
 :::terse
```
rogerburtonpatel, line 875: Closed the discussion. Calling it Nibble.

LF/Basics.lean
```diff
@@ -974,21 +995,10 @@ or the `seal` and `unseal` that we use in the examples below.
...
-:::dev
-TODO:
-Lean user question: how to get (succ (succ zero)) rather than
-NatPlayground.Nat.succ (NatPlayground.Nat.succ (NatPlayground.Nat.zero))
-:::
-
-:::dev
-BCP: Yes!
-:::
+::::full
```
rogerburtonpatel, line 1020: We really want this feature. Can we get it?

LF/Basics.lean
```diff
@@ -1086,11 +1113,6 @@ Here is a simple rule about `add`:
 In Lean, this rule looks like this:
 ::::
 
-:::dev
-BCP: Why does the Info View say "Goals accomplished!" right at the
-beginning of the proof?  Can we comment on this?
-:::
-
```
rogerburtonpatel, line 1093: @bcpierce00 Mine doesn't... can you share an example?

LF/Basics.lean
```diff
@@ -1107,6 +1129,10 @@ BCP: The flow here is rough
   - We should also comment, as soon as it's relevant, on the binder
     form vs. the ∀ form for introducing variables.  Why do we use
     the ∀ form here?  Could we switch to the binder form?
+RAB: We desperately need a way for this not to be the introduction to proof.
+Explaining it away with "don't worry" feels painful. Can we hide the definition,
+or do _something_ to make sure we're not having to explain ∀ , intro, by, etc.
+here?
```
rogerburtonpatel, line 1156: This is the focal point of the file. We need to discuss exactly how we will present the laws. 

In the Lean Natural Number game, they are secretly given as axioms, which works but is very hard to explain. The best would be to hide their proofs somehow and only show the types.

LF/Basics.lean
```diff
@@ -1270,6 +1313,9 @@ and performance; unfolding definitions can have negative effects as libraries sc
 BCP: We start by saying that what we're going to here is not what real lean developments do, but then
 explain why what we're doing in a way that makes it sound like it is (or should be) standard. And we
 never say what is the style that we *don't* do (but that standard Lean practice does).
+RAB: This will (hopefully) be addressed by our decision on hiding these
+definitions. Even if not, it seems odd to discuss how to write this section
+before we make that choice.
 :::
```
rogerburtonpatel, line 1347: @bcpierce00

LF/Basics.lean
```diff
@@ -1626,44 +1670,48 @@ GRADE_THEOREM 1: add_id_exercise
...
-TODO: how to get these to show `∀ (n : Nat)` instead of `mul_zero (n : Nat)`
-:::
-
-:::dev
-BCP: Or maybe we need to explain that they mean the same thing?  I
-think we are a bit inconsistent, ourselves, in the way we write
-things.
-:::
-
 ```lean
```
rogerburtonpatel, line 1708: I wrote a scratch version of this just below.

LF/Basics.lean
```diff
@@ -2087,19 +2119,16 @@ end Nat
...
+As a warmup for the last section of the chapter, use the tactics you have
+learned so far to prove the following theorem about boolean functions.
 
-:::dev
-TODO (Claude): This exercise quietly requires rewriting with a
-*universally quantified* hypothesis -- a real conceptual jump from
-rewriting with `h : n = m` that deserves a sentence of preparation.
-:::
+Hint: You can use `rewrite` with _any_ hypothesis that has an `=` in it
+as long as the types line up.
```
rogerburtonpatel: I think they can figure this out from here.

LF/Basics.lean
```diff
@@ -2175,3 +2196,382 @@ GRADE_THEOREM 3: andb_eq_orb
...
+```lean
+end LateDays
+```
+:::dev
+RAB: If we are to have this exercise, we must either
+make the functions irreducible or teach about
+`rw` of a reducible definition.
+We also have to figure out how to make lowerGrade\_lowers
+go through without `dsimp` or `contradiction`. To discuss.
+:::
```
rogerburtonpatel, line 2629: This is probably the only important part of this addition. tl;dr one of our example proofs uses `dsimp`, and I'm not yet sure how to get it to work without using a non-introduced tactic at this point (i.e. using only `rewrite`, `intro`, `rfl`, `exact`. Also, writing this, I realize they need `exact` for this exercise, which they do not yet have.

LF/Basics.lean
```diff
@@ -93,11 +93,13 @@ providing definitions of booleans, numbers, and many common data
 structures like lists and hash tables.  But there is nothing magic
 or primitive about these library definitions.  To illustrate this
 fact, we will explicitly recapitulate most of the definitions we
-need in this course, rather than just referring to the standard
-library. However, we will take care to harmonize those definitions
-with the ones in the standard library, so that, by the time you are
-finished the course, you will have a good grasp of how the standard
-library is organized.
+need in this course to introduce them rather than just referring
```
```suggestion
need in this course to introduce them, rather than just referring
```

LF/Basics.lean
```diff
@@ -93,11 +93,13 @@ providing definitions of booleans, numbers, and many common data
...
+to the standard library. However, we take care to harmonize
+those definitions with the ones in the standard library, as well
+as to gradually introduce the actual library definitions once appropriate.
```
```suggestion
as to gradually introduce the actual library definitions where appropriate.
```

LF/Basics.lean
```diff
@@ -93,11 +93,13 @@ providing definitions of booleans, numbers, and many common data
...
+By the time you are finished with the course, you will have a good grasp
+of how the Lean standard library is organized and how to efficiently
+find resources within it.
```
```suggestion
navigate it.
```

LF/Basics.lean
dsainati1, line 209: I don't think it makes sense to leave this comment in the file if you are removing the previous text; on its own "I am resolving the discussion" doesn't appear to refer to anything anymore.

LF/Basics.lean
```diff
@@ -1471,6 +1522,7 @@ def beq (n m : Nat) : Bool :=
 :::dev
 TODO: We need to make a decision about `==`.
 Either we give decidable equality early, or we use this `==` thing.
+RAB: This is the other thing we need to decide on together.
```
dsainati1: I think we agreed to use this `beq` function for now until we introduce typeclasses. Decidable equality this early is too much too fast.

LF/Basics.lean
```diff
@@ -1497,6 +1549,12 @@ example : leb four two = false := by rfl
 seal leb
 ```
 
+We'll be using `leb` a lot, so let's give it an infix notation too.
+
+```lean
+scoped infixl:55 " <? " => leb
+```
+
```
dsainati1: This was removed on purpose. Lean does not have notation for leb, so neither should we, per our goal of being as idiomatic as possible.

LF/Basics.lean
```diff
@@ -1626,44 +1670,48 @@ GRADE_THEOREM 1: add_id_exercise
...
+Qualification, like `mul_zero` to `NatPlayground.Nat.mul_zero`, can happen
+automatically when printing a type in Lean.
+
+Another simple but important-to-note automatic display feature is _indexing_:
+`mul_zero : ∀ (n : Nat), n * zero = zero` may display as
+`mul_zero  (n : Nat) : n * zero = zero`.
+
+Note how the (n : Nat) has moved _before_ the colon and has lost the ∀.
+The two definitions are equivalent for our purposes right now, and you will
+find both in the Lean ecosystem.
```
dsainati1: I am wary of saying this. The impression I got from @chenson2018 was that the `(n : Nat)` version is the idomatic way to do things; to be honest I am wondering whether we should bother showing the `∀` version at all, and maybe delay even discussing the universal quantifier until we get to Logic.

LF/Basics.lean
```diff
@@ -323,6 +310,12 @@ efficient executables. We'll come back to this topic in later
 chapters.
 ::::
 
+:::dev
+RAB: Is Lean compiling to C its "killer app," or is it the fact that it is an
+executable programming language (unlike Gallina)? We should get a Lean pro's
+take on what to say here.
```
berberman, line 335: I would describe Lean as a general-purpose programming language, where emitting C code is the final stage of the compiler pipeline. Since Lean is bootstrapping, unlike Gallina which can be extracted to OCaml optionally, Lean needs to produce performant executables so that it can compile itself. Also, there are FFIs, `@[implemented_by]`, and `native_decide` in kernel which require some interops with native code for optimizations, though I think these implementation details might be too specific to include here.

I also have been wondering that whether (or where) we should give a big picture of the pipeline, like the elaborated expression flows to the kernel typechecker (a tursted small core) just for checking, and simultaneously to comiler passes (LCNF -> IR -> C/Bytecode) for execution.

LF/Basics.lean
```diff
@@ -2175,3 +2196,382 @@ GRADE_THEOREM 3: andb_eq_orb
...
+namespace LateDays
+open scoped NatPlayground.Nat
+
+-- Numeric literals (`9`, `17`, `21`) for our unary `Nat`.
+@[reducible] def ofNat : _root_.Nat → Nat
+  | .zero => .zero
+  | .succ n => .succ (ofNat n)
+
+instance (n : _root_.Nat) : OfNat Nat n := ⟨ofNat n⟩
+```
```
dsainati1, line 2268: IMO we should avoid this if possible. We really don't want to confuse the issue by letting students treat our nats as equivalent to lean nats. We can probably change the exercise later to refer to `lateWeeks` or something being less than `one` or `two`. Alternative we can just define `nine` and `seventeen` and `twentyone` here.

LF/Basics.lean
```diff
@@ -2175,3 +2196,382 @@ GRADE_THEOREM 3: andb_eq_orb
...
+::::
+
+```lean
+inductive Letter : Type where
+  | A | B | C | D | F
+```
+
+::::full
+Then we define the modifiers -- a `natural` `a` is just a "plain"
+grade of `a`.
```
```suggestion
Then we define the modifiers -- a `natural` `A` is just a "plain"
grade of `A`.
```

LF/Basics.lean
```diff
@@ -2175,3 +2196,382 @@ GRADE_THEOREM 3: andb_eq_orb
...
+::::
+
+```lean
+inductive Modifier : Type where
+  | plus | natural | minus
+```
+
+::::full
+A full `Grade`, then, is just a `letter` and a `modifier`.
```
```suggestion
A full `Grade`, then, is just a `Letter` and a `Modifier`.
```

LF/Basics.lean
```diff
@@ -2175,3 +2196,382 @@ GRADE_THEOREM 3: andb_eq_orb
...
+In Lean, a combination of several values is called a _structure_.  The `structure`
+keyword is used to define a new structure type.
```
dsainati1, line 2293: I don't think students are going to have any intuition for what a structure type is, and this explanation is far too short to teach them IMO. We have material already explaining this in `Lists` when we talk about tuples, so IMO here we should just stick to the inductives they've seen up to this point for defining `Grade`.

berberman (commented):
Just curious, it seems that we're mixing `rw` and `rewrite` in this file. Is this somewhat intented?

LF/Basics.lean
```diff
@@ -2175,3 +2196,382 @@ GRADE_THEOREM 3: andb_eq_orb
...
+example : letterComparison D D = eq := by rfl
+example : letterComparison B F = gt := by rfl
+```
+
+::::exercise (rating := 1) (name := "letter_comparison")
+```lean
+theorem letterComparison_Eq : ∀ l : Letter,
+    letterComparison l l = eq := by
+  solution!
+    intro l; cases l <;> rfl
```
dsainati1, line 2367: We shouldn't use `<;>` in theorems yet

LF/Basics.lean
```diff
@@ -2175,3 +2196,382 @@ GRADE_THEOREM 3: andb_eq_orb
...
+GRADE_THEOREM 2: grade_lowered_once
+```
+:::
+::::
+
+```lean
+end LateDays
+```
+:::dev
+RAB: If we are to have this exercise, we must either
```
dsainati1, line 2624: We teach about `rw` of definitions in `UsingLean`, here is too early IMO. If we make these definitions irreducible though, there are going to be a painful number of rewrite lemmas that need to be used. Perhaps this is not the right exercise for this chapter anymore? We definitely something to take its place if not though.

LF/Basics.lean
```diff
@@ -2175,3 +2196,382 @@ GRADE_THEOREM 3: andb_eq_orb
...
+
+```lean
+theorem lowerGrade_lowers : ∀ g : Grade,
+    gradeComparison ⟨F, minus⟩ g = lt →
+    gradeComparison (lowerGrade g) g = lt := by
+  solution!
+    intro g h
+    match g with
+    | ⟨l, plus⟩ =>
+      rw [lowerGrade, gradeComparison]
```
dsainati1: We should not be using `rw` yet I think

LF/Basics.lean
berberman: Yes, IIRC there's a convention that we should prefer declaration header style (`(n : Nat) : ...`) over universal quantification (`... : ∀ n, ... `) style in general. But I think it's still useful to make the equivalence explicit here, and we'll elaborate this more in later chapters. FYI when printing out the signature of some theorem using `#check`, unless `pp.foralls` is set to `false`, the pretter printer always prints out Pi type `(x : A) -> B x` as `∀ (x : A), B x`, so it's kind of hard to avoid ∀ in general. And more importantly, `∀ (n : Nat), n * zero = zero` makes sense mathematically, which I think gives better intuition.

rogerburtonpatel:
> Just curious, it seems that we're mixing `rw` and `rewrite` in this file. Is this somewhat intented?

Ah, no. This is part of that exercise that needs to be fixed.

LF/Basics.lean
rogerburtonpatel: I'm inclined to believe this is a good thing to show early while keeping the details to a high level and again later while getting deeper. A diagram (or even embedded picture, if we can get that to work with the Verso version) here, and then later in the Typeclasses chapter, would be a nice bit of parallel structure. I support this!

LF/Basics.lean
rogerburtonpatel: Yeah, I'm getting rid of this in the next pass since it seems no one has more to say about it.

rogerburtonpatel: I should have left a github comment instead, but was inconsistent...

LF/Basics.lean
rogerburtonpatel: Yes, I'm going to rewrite all the solutions in this part.

LF/Basics.lean
bcpierce00: Replaced the issue with an inline request for later insutrctors to add their experiences.  (No point in having an issue now -- we can't get any information about it till September!)

LF/Basics.lean
rogerburtonpatel: I kind of agree. We should at least talk about making a new one.

LF/Basics.lean
rogerburtonpatel: Fair enough :)

LF/Basics.lean
rogerburtonpatel: Ah, gotcha. Sorry that I missed that. This is +0.5 for changing the grading exercise at the end of the chapter, as it relies on this notion. We could do it without notation, of course, but I'm hesitant to find reasons to keep the exercise where it is rather than ones to move it.

LF/Basics.lean
rogerburtonpatel: I think this needs to be presented very well- likely better than now- in order to work. I do agree that showing it at some point, probably early, is good, because the Infoview can show either and \forall makes mroe math sense. Also, we have a theorem in this chapter with a universally quantified hypothesis (`identify_fn_applied_twice`). We could move it, but we have to introduce \foralls early-ish, so either here or in Induction is probably our target, and Induction is already a lot of more-sophisticated material.

LF/Basics.lean
rogerburtonpatel: Changing the last exercise will eliminate this problem. Writing out `twentyone` means you need to invoke 
rewrite rules twenty-one times to get to a reducible form in a proof.

LF/Basics.lean
rogerburtonpatel: +1 to changing the exercise

LF/Basics.lean
rogerburtonpatel: Yes, solutions too...

rogerburtonpatel (commented):
Addressed comments. 
Main points:
1. We should probably change the last exercise 
2. Having a diagram of Lean compilation here and in the Typeclasses chapter (at different levels of detail) might be neat.

rogerburtonpatel:
We should remove the last exercise for now to get the PR to merge.

LF/Basics.lean
bcpierce00: Probably useful information, but not in this chapter -- let's look for a good place later.  (Maybe the UsingLean chapter?  Or is that still too early?)

chenson2018: There is a chapter [Elaboration and Compilation](https://lean-lang.org/doc/reference/latest/Elaboration-and-Compilation/) in the reference manual that includes a diagram. This early is definitely too overwhelming, maybe even for volume one in general, but maybe we can link to something external like the reference manual as optional reading?

LF/Basics.lean
chenson2018: Is the difference in naming intentional? The core naming is [Nat.ble](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#Nat.ble)

rogerburtonpatel:
Not intentional, and this should change. I hacked this in to get the grading exercise to work; we should definitely come up with a more intelligent solution.

LF/Basics.lean
chenson2018: In short: yes, the style guide specifies that a hypothesis to the left of the colon is preferred. I think it is useful to use `∀` a bit at first so the student can see the meaning, but somewhere there needs to be a paragraph that explicitly explains that these are the same. (I'm surprised this doesn't exist already in the Rocq version?)

LF/Basics.lean
bcpierce00: I guess we are assuming that anybody reading this file will be either looking at the .lean version in an IDE or else looking at the HTML version in a browser.  Either way, hovering is enough to see the type.  (There's an edge case for lecturing, but even there I think it is probably fine / even beneficial to hover over the text and see the thing pop up in real time.)

LF/Basics.lean
bcpierce00: Leaving this comment pending for now, as it's part of a larger discussion.

LF/Basics.lean
bcpierce00: Yes

LF/Basics.lean
bcpierce00: E.g., just make APlus, A, AMinus, etc. separate constructors?

PR #46, opened 2026-06-29
Typeclasses

chenson2018:
This is a first pass at a chapter on typeclasses. This is still in a rough draft state, but I'm trying to follow Benjamin's request to PR things earlier if possible. There are a couple of organizational questions in my mind:

- In `LF/Maps.lean` I have already ported a version of that chapter, but using typeclasses. I left a placeholder here where it seems natural to fit in, but I wasn't sure if this was too long.
- The reflection material collects the examples from `Logic` and `IndProp`. I left things unfinished towards the end where I wasn't sure what was important to translate.


LF/Typeclasses.lean
```diff
@@ -0,0 +1,593 @@
...
+  one : α
+  two : α
+  one_neq_two : one ≠ two
+```
+
+which enforces that these are two distinct entries. Declaring instances works in much the same
+way, except that now {name}`HasTwo.one_neq_two` requires a proof:
+
+:::dev
+I'm not actually sure of the preferred way to prove `1 ≠ 2` at this point in the book -- CGH
```
dsainati1: I think it would be `intro contra; contradiction`

LF/Typeclasses.lean
```diff
@@ -0,0 +1,593 @@
...
+    rewrite [List.elem_poly_cons, List.elem_nat_cons, ih]
+    rfl)
+```
+::::
+
+# Maps
+
+:::dev
+Maps could go here as an example to reinforce this? Or maybe that's too long of an aside? The code
+exists (with limited prose) in `Maps.lean` already. --CGH
```
dsainati1: I think the idea is to fold `Maps` into this chapter entirely, using it as an extended example to motivate typeclasses.

LF/Typeclasses.lean
```diff
@@ -0,0 +1,593 @@
...
+particular, the type `Nat → Prop` does contain functions representing properties like
+"the nth Turing machine halts." The second row in the table follows directly from this essential
+difference. To evaluate a pattern match (or conditional) on a boolean, we need to know whether the
+scrutinee evaluates to `true` or `false`; this only works for `Bool`, not `Prop`.
+
+The third row highlights an important practical difference: equality functions like {name}`Nat.beq`
+that return a boolean cannot be used directly to justify rewriting with the rewrite tactic;
+propositional equality is required for this. Since `Prop` includes both decidable and undecidable
+properties, we have two options when we want to formalize a property that happens to be decidable:
+we can express it either as a boolean computation or as a function into Prop
```
dsainati1:
```suggestion
we can express it either as a boolean computation or as a function into Prop.
```

LF/Typeclasses.lean
```diff
@@ -0,0 +1,593 @@
...
+
+So what should we do in situations where some claim could be formalized as either a proposition or a boolean computation? Which should we choose?
+
+In general, both can be useful. Which we choose has to do with the _computational_ nature of Lean's
+core language, which is designed so that every function it expresses is total, and by default
+computable unless we explicit indicate otherwise. As an example, consider
+trying to write a function `α → α → Bool` checking for equality on an arbitrary type:
+
+```lean -keep +error
+def eq {α : Type} (a₁ a₂ : α) : Bool := if a₁ = a₂ then true else false
```
dsainati1, line 742: We use regular `if` for the first time here. It is probably necessary to explain at this point what `if` is and how it differs from `bif`.

LF/Typeclasses.lean
```diff
@@ -0,0 +1,593 @@
...
+
+```lean
+section
+unseal isEven
+
+example : Even 2 := by decide
+example : Even 4 := by decide
+example : Even 6 := by decide
+example : Even 100 := by decide
+example : Even 1000 := by decide +kernel
```
dsainati1: What is the `+kernel`

LF/Typeclasses.lean
chenson2018: I would be in support of this, I'll wait a bit to see that others agree before doing so.

LF/Typeclasses.lean
chenson2018: Here is the docstring (NB: for tactic options like this you can hover to see this info as well):

> If true (default: false), then use only kernel reduction when reducing the Decidable instance. This is more efficient, since the default mode reduces twice (once in the elaborator and again in the kernel), however kernel reduction ignores transparency settings.

It is required for to not hit the recursion depth for larger examples. If it's not important to mention here we could omit this.

dsainati1: IMO rather than explain this we should just not use an example this large tbh

LF/Typeclasses.lean
berberman: Should we clarify the difference between `=` and `==`? Observably `if` and `bif` can possibly accept both as the condition because of `Coe` or `DecidableEq` instances, which IMO could be confusing.

Probably we can talk a bit about the coercion system in this file as well, since `Coe` could be an example of typeclasses, so long as if we ignore the `outParam` thing...

chenson2018: I definitely intended for this to cover `=` versus `==`. Maps uses `LawfulBEq` (which says `=` and `==` coincide). If that will now appear here it's a good place to give some more detail?

LF/Typeclasses.lean
```diff
@@ -0,0 +1,593 @@
...
+exists (with limited prose) in `Maps.lean` already. --CGH
+:::
+
+# Reflection
+
+:::dev
+I think this will still exist in previous chapters, just not have the reflection explanations until
+here? Since I can't import these yet, just placing here at the top of this section -- CGH
+:::
+
```
rogerburtonpatel, line 537: These definitions of even as boolean computation and Prop should go below, after the table where we explain the difference.

LF/Typeclasses.lean
rogerburtonpatel: I think right after this part on decidability is good. It's a hefty chunk of information already, so keeping distinct ideas distinct is more likely than not a good call.

LF/Typeclasses.lean
rogerburtonpatel: Agreed

LF/Typeclasses.lean
```diff
@@ -0,0 +1,593 @@
...
+```lean
+set_option pp.all true in
+#print nat_eq
+```
+
+which proves that this equality is decidable.
+
+This is only half the story however, as while Lean's core theory enables this computation, Lean is
+also often used in applications where we don't care about computability. In particular, it is
+possible to write a function a function for arbitrary equality
```
rogerburtonpatel:
```suggestion
This is only half the story, however. While Lean's core theory enables this computation, Lean is
also often used in applications where we don't care about computability- 
for example, pure mathematics. In particular, it is possible to write a function a function 
for arbitrary equality:
```

LF/Typeclasses.lean
```diff
@@ -0,0 +1,593 @@
...
+```lean -keep
+open scoped Classical in
+noncomputable def eq {α : Type} (a₁ a₂ : α) := if a₁ = a₂ then true else false
+
+set_option pp.all true in
+#print eq
+```
+
+but we have indicated to Lean using the `noncomputable` keyword and `Classical` namespace
+that we are not interested in computation. What is happening in the background is that this allows
```
rogerburtonpatel:
```suggestion
But we have indicated to Lean using the `noncomputable` keyword and `Classical` namespace
that we are _not_ interested in computation. What is happening in the background is that this allows
```

LF/Typeclasses.lean
```diff
@@ -0,0 +1,593 @@
...
+
+:::dev
+I'm not sure what part of the signature here is important to translate. Is the point the
+`Bool`/`Prop` mismatch? -- CGH
+:::
+
+```lean
+example (a : α) [BEq α] [LawfulBEq α] (xs : List α) (neq : xs.filter (a == ·) ≠ []) : a ∈ xs := by
+  sorry
+```
```
rogerburtonpatel, line 859: Some more examples would be good. It might be good to start with `Nat` and then move to the `Indprop` ones. This is a short chapter, so 5-6 well-chosen, informative exercises could easily fit.

rogerburtonpatel (commented):
Looks like a great start. I recommend a few more exercises, and some shifting of text, but it's already in a nice spot if we assume a post-indprop audience.

chenson2018:
Note to self: confirmed in another PR that we will have structures by this point, so should use that in the initial explanation.

LF/Typeclasses.lean
bcpierce00: That was my impression also.

bcpierce00:
I've had a quick look at the PR -- will review the chapter more thoroughly later, but I suggest we close the discussions on the PR, merge it, and continue thinking about all the remaining stuff at leisure.

Doesn't seem to compile right now, though...

chenson2018:
> Doesn't seem to compile right now, though...

I think there's something odd happening with the generated Lean files, I'll have a look.

bcpierce00:
Ah.  Sigh.  I can also help look, if that's useful.

LF/Typeclasses.lean
```diff
@@ -0,0 +1,639 @@
...
+
+Suppose we wanted to specify that a type has at least one inhabitant -- i.e.,
+that it is not empty. In Lean, we could express this with the typeclass
+
+```lean
+class HasOne (α : Type) where
+  one : α
+```
+
+which explicitly carries an element `one : α` as a witness of non-emptiness. After defining this
```
KihongHeo: I think it would be better to explain the meaning of this class declaration more explicitly. Many readers may be new to typeclass. 

Here I feel that the terms `carries` and `witness` are somewhat unclear at first. Also it would be nice if the syntax is also explained together. Can we explain it something like this?
- We want to declare a class of types that satisfy certain conditions. The name of such a class is `HasOne`
- The condition that all classes in the class should meet is to have at least one element. The field of the class `one` describes the constraint.
- When you want to declare any instance of this class, you should provide a witness of this non-emptiness.
- (Examples follow)

LF/Typeclasses.lean
```diff
@@ -0,0 +1,639 @@
...
+typeclass, we can provide *instances* that show it is satisfied for a particular type. As an
+example, we could declare an instance
+
+```lean
+instance instHasOneNat : HasOne Nat where
+  one := 1
+```
```
KihongHeo, line 95: Similarly, can we explain what this syntax means in detail?
For example, the name `instHasOneNat` is never used in this file. So readers may wonder when this is used.

chenson2018: I don't think it's correct to say this instance is not used again in this file. The point of the immediately proceeding examples using `#synth` and `pp.all` is that they show the usage of this instance.

This can definitely be better explained, I'll try making this more explicitly explained when working on the other review feedback.

KihongHeo: Oh I see. Thanks.
I just did `grep instHasOneNat` and did not see any results. Maybe readers like me also do that.

chenson2018: Part of the point here is that we don't usually refer to instances explicitly, even me providing an explicit name is for pedagogy reasons. I'll make it more explicit in the text that what we're looking for in the infoview is the presence of this instance.

dsainati1:
This wasn't building because the `-keep` and `+error` flags were being ignored during Verso extraction. I left a comment to this effect in the file and commented out the erroring code.

dsainati1:
Build on this remains broken for two reasons, as best I can tell:
* The generated `LF.IndPropVerso` file does not seem to typecheck
* The import of Typeclasses into HL.Imp doesn't strip the verso module imports

dsainati1 (approved):

PR #47, opened 2026-06-30
First proof in Basics should be with `MyBool`s

dsainati1:
Currently, `Basics` has an issue where we show proofs before we actually explain how they work. This attempts to resolve that problem by presenting proofs on `Bool` types first, teaching intro and rfl, and then later on introduces proof by rewriting with `Nat`s.


LF/Basics.lean
```diff
@@ -490,10 +486,126 @@ GRADE_THEOREM 1: andb3_test4
...
+- `.true && b = b`
+
+This is an example of a _proposition_, a logical _claim_ which we can try to prove.
+This proposition says that for every `MyBool` `b`, `.true && b` is equal to `b`.
+
+How might we write this proposition in Lean?
+
+- `theorem true_and : ∀ b : MyBool, (.true && b) = b`
+
+The keyword `theorem` indicates that we stating (and eventually proving)
```
xhalo32: typo: we are

LF/Basics.lean
```diff
@@ -490,10 +486,126 @@ GRADE_THEOREM 1: andb3_test4
...
+a proposition; the text after the `:` is the proposition we want to prove.
```
xhalo32: Should clarify which `:` (first or second)

LF/Basics.lean
```diff
@@ -490,10 +486,126 @@ GRADE_THEOREM 1: andb3_test4
...
+about `&&`:
+
+- `.true && b = b`
+
+This is an example of a _proposition_, a logical _claim_ which we can try to prove.
+This proposition says that for every `MyBool` `b`, `.true && b` is equal to `b`.
+
+How might we write this proposition in Lean?
+
+- `theorem true_and : ∀ b : MyBool, (.true && b) = b`
```
xhalo32: Should we consider putting parenthese around `b : MyBool` to avoid syntactic confusion about the role of each colon?

LF/Basics.lean
```diff
@@ -490,10 +486,126 @@ GRADE_THEOREM 1: andb3_test4
...
+```lean
+theorem true_andb : ∀ b : MyBool, (.true && b) = b := by
+  intro b
+  rfl
+```
+
+::::full
+What does this mean?
+
+The first new piece of syntax is the `by` keyword, which signals
```
xhalo32: by is introduced on line 304 so it's not a new piece of syntax strictly speaking

LF/Basics.lean
```diff
@@ -490,10 +486,126 @@ GRADE_THEOREM 1: andb3_test4
...
+to Lean that we are beginning a sequence of _tactics_. Tactics are Lean's other programming language, alongside the functional language you've been working with so far, and
```
xhalo32: "other programming language" makes it sound like there are two programming languages. A better way of phrasing could be "by starts proof mode which is a sequence of tactics"

LF/Basics.lean
```diff
@@ -490,10 +486,126 @@ GRADE_THEOREM 1: andb3_test4
...
+    "We want to prove <some property> for all `MyBool`s `b`. So suppose
+    `b` is some arbitrary `MyBool`... <and then go on to prove the
+    property for this particular `b`>..." Since `b` was chosen
+    arbitrarily, we've now proved the property for all `b`.
+
+    A proof of a theorem beginning with a ∀ will typically start with
+    an `intro`.
+
+    Like the `example`s we saw above, we can use the `rfl` tactic,
+    which closes goals about equality where both sides are reflexively equal to one another.
```
mwhicks1: I don't know what "reflexively equal" means. Did it come up before? I think rfl refers to "reflexivity" which means something is equal to itself. Maybe reword "both sides are equal to one another according to the principle of _reflexivity_ which says that something is equal to itself."

LF/Basics.lean
```diff
@@ -490,10 +486,126 @@ GRADE_THEOREM 1: andb3_test4
...
+    Now, inspecting our goal will show that it is `(true && b) = b`, which
+    may not appear to be reflexively equal at first glance. However, if we look
```
mwhicks1: "may not appear to be equal to itself. However, the tactic _evaluates_ both sides of the equality and then compares them. In this case, if we look"

mwhicks1 (approved):
Looks great! I made two small suggestions.

LF/Basics.lean
```diff
@@ -490,10 +486,126 @@ GRADE_THEOREM 1: andb3_test4
...
+```
+GRADE_THEOREM 1: false_orb_exercise
+```
+:::
+::::
+
+::::full
+Obviously these facts about booleans are quite simple, so the tactics we need to
+prove them are also quite simple. Over the course of this book we are going to
+introduce new tactics and proof techniques gradually, expanding the set of propositions
```
xhalo32: Don't take this comment too seriously, but I think a "set of propositions" leads to logical paradoxes, i.e. it can't be a set. I don't think this is a real issue. This just reminds me of one lecturer I had who was careful about saying things like "set of all ..."

xhalo32 (commented):
Looking good! Left a couple of notes.

LF/Basics.lean
dsainati1: Are tactics not considered a separate language from Lean's normal language. I imagine the difference isn't as clear cut as Gallina vs. Ltac, but does one exist?

xhalo32: AFAIK the difference between tactic and term mode (i.e. the language of expressions/terms) is simply that they are different [syntax categories](https://lean-lang.org/doc/reference/latest/Notations-and-Macros/Defining-New-Syntax) i.e. "nonterminals in a context-free grammar."

I just noticed that the ref calls tactics a "programming language" in https://lean-lang.org/doc/reference/latest/Tactic-Proofs/#tactics so I guess it's fine that we call it as such.

dsainati1: Yeah, calling tactics a DSL makes sense to me.

bcpierce00: IMO calling tactics a programming language at this point is a recipe for confusion.  I reworded one instance -- maybe there are others.

bcpierce00:
It looks good overall.  Fine to merge.

Issue #25, opened 2026-06-24
Work in progress (tracking meta-issue)


bcpierce00:
# In-progress tasks

| Author | Locked files | Branch | Comment |
| :--- | :--- | :--- |  :--- |
| @bcpierce | all not-yet-versified LF chapters except`Typeclasses`, but tell me if you want locks | `bcp-versificationXX` | Versifying remaining chapters of LF |
| @mwhicks | `TS/Stlc.lean`, `LF/Typeclasses.lean` | `mwh/ts-stlc` (https://github.com/plclub/sf-in-lean/pull/156), https://github.com/plclub/sf-in-lean/pull/164 | Just getting started ... |
| @chenson2018 |  |  |  |
| @xhalo32 | `LF/Lists.lean` | https://github.com/plclub/sf-in-lean/pull/136 |  |
| @dsainati1 and @rogerburtonpatel | `Automation.lean` | dhs-automation | PR in progress |
| @rogerburtonpatel | `Prefix.lean` and `Postscript.lean` (new files) | | Issue #34: Convert prefix and postscript to lean |
| @berberman | `Poly.lean` | #138 | Clarify inductive parameters and clean up Rocq discussions |

-----------------------------------

# Translation status

_Please help keep this up to date!_

Substantially translated:
- Basics.lean — Mostly done -- just needs final polish (WIP polishing at #93)  
- Induction.lean — Also in pretty good shape.
- UsingLean.lean — New chapter covering Lean's built-in Nat, calc, dsimp, and
  stdlib search tactics. Needs more careful reading.
- Lists.lean — Converted to the new rewrite-forward style, but needs instructor eyes
- Poly.lean — Converted to the new rewrite-forward style, but needs instructor eyes
- Tactics.lean — Substantially translated; however, many exercises in this chapter are 
  based on defining and proving properties about `Nat.ble` and `BEq.eq`, which are not
  idiomatic in Lean. We should consider replacing these with a different set of exercises. 
- Logic.lean — Substantially translated, but with notes that the Classical
  Axioms section needs rewriting for idiomatic Lean. Also has similar issues to Tactics. 
- IndProp.lean — Compiling, but translation is very much a first draft.
- TS/Slang.lean (symlinks HL/Slang.lean) — Mostly done. All proofs need human re-do 
- HL/Imp.lean - Mostly done. All proofs need human re-do 
- TS/Smallstep.lean - Initial pass done. All proofs need human re-do 
[...]

PR #55, opened 2026-07-05
Introduce match code actions from Batteries in UsingLean

berberman:
This PR intorduces batteries to the project and updates UsingLean to include a demonstration of how to use the code action.


xhalo32 (dismissed):
Looks good, although I only reviewed by looking.

LF/Basics.lean
```diff
@@ -376,14 +381,24 @@ detail below.
 namespace MyBool
 ```
 
+Now {name}`true` resolves to {name}`MyBool.true` rather than the built-in {name}`Bool.true`,
+because we are inside the `MyBool` namespace.
+Hover over {name}`true` in the `#check` command to see this.
```
dsainati1: I think this is confusing because we haven't actually introduced the `Bool` type. IMO this explanation makes more sense later once we are done with `MyBool` and switch over to `Bool`.

LF/Basics.lean
```diff
@@ -1906,18 +1919,65 @@ negation is involutive (that is, that negation is its own inverse).
 :::slidebreak
 :::
 
+## Aside: Using the Code Action to Generate Match Skeleton
+
+:::full
+
+Lean's language server provides code action that can generate the missing branches for pattern matching.
```
```suggestion
Lean's language server provides code actions that can generate the missing branches for pattern matching.
```

LF/Basics.lean
```diff
@@ -1906,18 +1919,65 @@ negation is involutive (that is, that negation is its own inverse).
 :::slidebreak
 :::
 
+## Aside: Using the Code Action to Generate Match Skeleton
```
```suggestion
## Aside: Using the Code Action to Generate Match Skeletons
```

LF/Basics.lean
```diff
@@ -1906,18 +1919,65 @@ negation is involutive (that is, that negation is its own inverse).
...
+
+:::full
+
+Lean's language server provides code action that can generate the missing branches for pattern matching.
+This is very useful when using `match`, or tactics such as `cases` and `induction` (which we'll see in later chapters).
+
+For example, suppose we start with the following proof:
+
+```lean +error
```
dsainati1: Oh does this allow you to write an erroring Lean block without needing to `#guard_msgs`?

berberman: Yes. Code blocks with +error are temporarily stripped off in project extraction to prevent the build failure (implemented in this PR.) I'm not sure if we should keep them in the extracted student project by turning them into docstrings or inventing some command like guard_msgs. Or we might just skip building the extracted project?

dsainati1: Oh, I was mostly wondering how to actually put erroring code blocks into the extracted student project without causing the build to fail, so that we can use them to demonstrating failing proofs and the like. But it sounds like you still need #guard_msgs for that.

berberman: Yeah, the student project is in plain Lean with no verso magics, so we can't prevent wrong code failing the compilation without extra commands/setups.

I'm inclined to add a special command, something like `#invalid_lean`, which runs the following command in a sandbox, i.e. saving the the environment, elaborating the top-level command, capturing error messages (reporting optionally), and finally restoring the saved environment. The difference from `#guard_msgs` would be that `#invalid_lean` wouldn't require diagnosis as docstrings, so we can turn all `lean +error` code blocks into `#invalid_lean` guarded code in student project automatically. But the downside would be introducing a new syntax that is non-standard in Lean. Maybe @chenson2018 and @mwhicks1 have some thoughts?

LF/Basics.lean
berberman: IMO it's fine, given that we just said "The next command opens a new namespace so that our definitions don't clash with ones from the standard library. We'll discuss it in more detail below." above.

LF/Basics.lean
berberman: Oh I wasn't aware that `#guard_msgs` supports `drop all` option. Then it won't make sense to re-invent the entire `#guard_msgs`. But having an alias command might still be useful, IMO.

berberman:
I'm going to split the SFLMeta changes in to a separate PR for future discussions, and avoid using `+error` in this PR.

LF/Basics.lean
mwhicks1: I agree with Daniel. This is going to be part of a larger change around prefixing/dot-notation after his other PR drops. I wonder if we can separate out aspects of this PR that will conflict with Daniel's? (The indentation change below will also conflict.)

LF/Basics.lean
```diff
@@ -614,11 +629,15 @@ properties about them, let's switch to Lean's built-in `Bool` type, which has th
 but comes with a lot of useful functions and lemmas.
 
 One thing to note:
-because `Bool` is such an important type in Lean, its `true` and `false` constructors can be written
-without a `.` to qualify them.
+Because {name}`Bool` is such an important type in Lean,
+its constructors {name}`Bool.true` and {name}`Bool.false` can be referred to simply as `true` and `false`,
+since they are exported from the `Bool` namespace.
+However, inside another namespace that defines its own `true` and `false` constructors, such as {name}`MyBool` we've seen above,
+these unqualified names resolve to the local constructors instead.
```
mwhicks1: This is useful explanation. @dsainati1 seems to be planning a big change about naming in this section, so your explanation here should be accommodated by his change.

LF/Basics.lean
```diff
@@ -814,12 +825,12 @@ def is_weekend (d : Day) : Bool
   := solution!
     (match d with
     | .saturday => true
-    | .sunday => true
-    | _ => false
+    | .sunday   => true
+    | _         => false
```
mwhicks1: Just happens to occur to me here: The first time we introduce this notational convention (same-line match clauses matching up vertically), we should explicitly say it in the text. Maybe go back and find the first case and point it out?

LF/Basics.lean
```diff
@@ -1615,12 +1626,14 @@ equality, yielding a boolean.
...
-               | zero => false
-               | succ m' => beq n' m'
+  | zero =>
+    match m with
+    | zero   => true
+    | succ _ => false
+  | succ n' =>
+    match m with
+    | zero    => false
+    | succ m' => beq n' m'
```
mwhicks1: This may be the first place this notation convention is used (for nested matches); worth pointing out in text that you are doing it.

LF/Basics.lean
```diff
@@ -1906,18 +1919,65 @@ negation is involutive (that is, that negation is its own inverse).
 :::slidebreak
 :::
 
+## Aside: Using the Code Action to Generate Match Skeletons
```
mwhicks1: This change is very different from the indentation and namespacing changes. It should be in its own PR (and those in their own).

LF/Basics.lean
```diff
@@ -1906,18 +1919,65 @@ negation is involutive (that is, that negation is its own inverse).
...
+
+If you put your cursor after after `cases n`, Lean will offer a code action "Generate an explicit pattern match for 'cases'."
+In VSCode, you can usually open the code action menu with `Ctrl + .` on Windows/Linux and `Command + .` on macOS.
+
+After running the code action, Lean fills in the cases for us:
+
+```lean
+example (n : Nat) : True := by
+  cases n with
+  | zero => sorry
```
mwhicks1: We have not introduced `sorry` as a concept, I believe. 

I wonder: Is Basics the best place to introduce this concept? We are already cramming a lot into this file. Could this go into `UsingLean.lean` instead?

LF/Basics.lean
```diff
@@ -1906,18 +1919,65 @@ negation is involutive (that is, that negation is its own inverse).
...
+  | succ n => sorry
+```
+
+So instead of writing out all the brances by hand, we get the basic shape of the proof tree.
```
mwhicks1: typo: brances --> branches

LF/Basics.lean
```diff
@@ -1997,6 +2057,11 @@ As you can see, proofs by cases can become very verbose.
   We will introduce some tactics for writing shorter proofs
   by case analysis in `Tactics.lean`.
 
+:::dev
+YL: What tactics specifically? I imagine the shortest one would be `decide`. For case analysis,
+`cases b <;> cases c <;> cases d <;> rfl` should close the goal, but does `Tactics.lean` cover `<;>`?
+:::
```
mwhicks1: We mention `<;>` first in Induction, but don't spend a lot of time on its benefits.

scripts/check_verso_prose.py
```diff
@@ -52,7 +52,7 @@
 _MARKER_LINE_RE = re.compile(
     r"^\s*--\s*(/?FULL|/?TERSE|/?HIDE\w*|/?QUIZ|/?SOLUTION|/?QUIETSOLUTION"
     r"|/?WORKINCLASS|/?ADMIT\w*|EX\d*\w*|GRADE_\w+|INSTRUCTORS|RAB|RADDITION"
-    r"|BCP|JC|MWH|CGH|CH|HG|NB|MMG|APT|DHS|TODO|TOFIX|LATER|SOONER|\[\])\b",
+    r"|BCP|JC|MWH|CGH|CH|HG|NB|MMG|APT|DHS|YL|TODO|TOFIX|LATER|SOONER|\[\])\b",
```
mwhicks1: Just observing: I thought we had established the convention that we were using GitHub IDs, not initials? Several other places in this script, too, see below. @bcpierce00

mwhicks1 (commented):
My main concern about this PR is that it does three things, and they should all be separate:
1. It discusses name qualification
2. It fixes indentation for code using `match`, `cases`, etc.
3. It introduces Batteries with auto-complete for `match`, `cases`, etc.

The name qualification stuff will get subsumed in a big change @dsainati1 is making. The  indentation PR should drop on top of the most recent Basics PR. The Batteries change we should discuss. It doesn't seem to add that much value to me, and I think that Basics is already quite large. I might suggest including it in UsingLean instead.

LF/Basics.lean
chenson2018: Is this spacing around `=>` something existing or is being added in this PR? The usual style is single spaces.

berberman: > Is this spacing around `=>` something existing or is being added in this PR? The usual style is single spaces.

The guideline for this style was already there before I was in, so I followed it. My opinion: https://github.com/plclub/sf-in-lean/pull/56#discussion_r3531408467

chenson2018: Hmm. I personally do like aligning things, but maybe it's not too strictly enforced either way. In anticipation of the FRO autoformatter I'm inclined to not feel too strongly.

berberman:
Thank you for the review, @mwhicks1. I agree that operationally this PR should probably be split, and it is good to see that #64 will improve the namespace issue and subsume part of this PR.

Just to clarify, I put all three things into one PR because I was hoping to get this small bookkeeping change merged quickly so that it would not block @dsainati1's PR on Monday (#59). I claimed the lock in #25 on Saturday and had this PR ready for review by the end of the weekend. It turned out that this PR needed more refinement, so #59 was merged first. As I said in Zulip, I am happy to rebase on #59 since it is more valuable and has higher priority.

That said, this does make me a bit concerned about the lock mechanism we are trying to follow. I also noticed that #63 will likely to conflict with #57, even though #57 is holding the lock on Lists. Personally, I am not opposed to different people working on overlapping files in different PRs, since each PR has different progress and may take different amounts of time to land. It seems natural for later PRs to rebase and resolve conflicts. But is that something our locking convention is supposed to prevent?

For batteries, my impression was that we had agreed to introduce the code actions, and I chose to do so in Basics because: 1) later files import Basics, so we don't need to import the module everywhere; and 2) we had already discussed VS Code extension-related things in this file, so IMO this seemed like a reasonable place to add a bit more about it. I am fine with moving them to UsingLean, but then the code actions will not be available in Induction.

dsainati1:
I think the locking system is only functional if we actually use the locking issue properly, and in these two instances it seems like we did not. Looking back at #25, I see that you did have the lock on Basics already. I missed it when I was looking over the issue because it was not in the format I was expecting, but that's my fault. I should have waited for this to merge first. Similarly, #63 probably should not have been submitted, or should have at least been based on top of #57 since you have the lock. 

I am unsure what the solution is though, beyond moving to a heavier-weight conflict resolution system.

scripts/check_verso_prose.py
bcpierce00: Yes, we have, and at some point we need to go through all the files and change them accordingly.  Was planning to do that after versification, so this script will not matter.

lakefile.toml
```diff
@@ -22,7 +22,13 @@ weak.verso.code.warnLineLength = 0 # 0 disables warnings; consider setting to 70
 [[require]]
 name = "verso"
 git = "https://github.com/leanprover/verso"
-rev = "nightly-testing"
+rev = "nightly-2026-06-25"
+
+# To be switched back to main. See the note above verso.
+[[require]]
+name = "batteries"
```
chenson2018, line 29: Something we've not mentioned on this PR: when we first discussed adding an ecosystem dependency, I suggested Mathlib with the idea to only use a limited `Mathlib.Tactic.Common` import. (Mathlib includes batteries transitively, so is compatible with this PR)

I'd still personally prefer this and I believe Benjamin had initially agreed. If opinion on this has shifted and we do limit ourselves to batteries though, let's make sure we understand and document the differences. What comes to mind is:

- There are a few things that will remain vendored from Mathlib, I think mostly in `LF/CustomTactics.lean` currently. Let's give some thought to if this will increase much more in size.
- As an example, there are the Mathlib syntax linters to consider. The one I have in mind that would be nice to have is the flexible linter for the upcoming automation chapter @dsainati1 is writing, as I think discussion of non-terminal `simp` is essential material. This can still be vendored, but it's a bit larger. 
- The cache will work differently. I believe the new `lake cache` (as opposed to Mathlib `lake exe cache`) is very recently up and running? Either way, the basic usage should be documented.

berberman: I completely agree with you and I would prefer Mathlib as well. It seems that I wasn't in the initial discussion and I had a wrong impression that we wanted to avoid Mathlib for some reasons... Thanks for pointing this out.

chenson2018: I really should have spoken up sooner, I'm just stretched a touch thin with review work at the moment. If there aren't strong objections, I think in the interest of simplicity we should just update this PR to a Mathlib dependency (or split one off doing just this), communicate in the Zulip to use `lake exe cache get`, and follow up in future PRs with the simplifications this allows.

bcpierce00: I have no fundamental; objections to either Mathlib or Batteries -- just the usual software engineering concerns of limiting dependencies to what is really needed, plus overall simplicity (and the recognition that we will need to explain whatever we use).  

One significant point in favor of these dependencies, if I understand rightly, is that it means we can align our style better with standard usage.

mwhicks1: I don't know if we want to avoid Mathlib or not, but I would prefer we merge this PR as is, as it's been pending for awhile and in general I want to keep high velocity. Assuming we do want to use Mathlib, I think we need to understand how, and how to think about it, which warrants some (probably real-time) discussion.

mwhicks1 (approved):

LF/UsingLean.lean
```diff
@@ -447,3 +449,114 @@ theorem double_mul (n : Nat) : double n = 2 * n := by
...
+
+/-
+  You might notice that Lean generated `0` and `n + 1`, not `zero` and `succ n`.
+  That is because Lean has a feature called _custom pattern functions_.
+  Usually, patterns are written using constructors,
+  but Lean also lets certain specially marked functions appear in patterns.
+  These functions are unfolded until Lean sees the constructor pattern they stand for.
+
+  We will keep writing the constructors as `zero` and `succ` anyway, because it makes the structure
+  of `Nat` easier to see.
```
bcpierce00: The discussion about match expressions is a little worrying -- it seems like we are explaining custom pattern functions and implicitly inviting students to use them, but then remarking that they are potentially confusing so we are not going to use them ourselves.  

IMO we should make a choice: Either avoid them (and then not teach them, or at least not in this early chapter) or else decide that they are so useful in SLF that it is worth the overhead of really teaching them.

bcpierce00: Maybe this is worth a zulip discussion...

LF/UsingLean.lean
```diff
@@ -447,3 +449,114 @@ theorem double_mul (n : Nat) : double n = 2 * n := by
   rw [double_add, Nat.two_mul]
   -- /ADMITTED
 -- GRADE_THEOREM 1: double_mul
+
+/-
+  # Using the Code Action to Generate Match Skeletons
+-/
+
+/-
+  Lean's language server provides code actions that can generate the
```
bcpierce00: Some readers might not know what code actions are -- let's add a brief definition (and perhaps a pointer to documentation).

LF/UsingLean.lean
```diff
@@ -447,3 +449,114 @@ theorem double_mul (n : Nat) : double n = 2 * n := by
...
+  missing branches for pattern matching.
+
+  This is very useful when using `match`,
```
bcpierce00: see discussion below about match

LF/UsingLean.lean
```diff
@@ -447,3 +449,114 @@ theorem double_mul (n : Nat) : double n = 2 * n := by
...
+n✝ : Nat
+a✝ : eqb n✝ n✝ = true
+⊢ eqb (n✝ + 1) (n✝ + 1) = true
+-/
+#guard_msgs(error) in
+theorem foo (n : Nat) : eqb n n := by
+  induction n
+
+/-
+  If you put your cursor after after `induction n`, Lean will offer a code action
```
bcpierce00: again, for people new to code actions, "Lean will offer" might warrant a few words of explanation

bcpierce00:
I think this PR still needs some work to pull it apart into separate pieces.  I am going to go ahead and do the versification of as many chapters as possible.  This may involve some manual merge work for this PR.

lakefile.toml
berberman: Yes, let's keep batteries in this PR. I could switch to mathlib, but I felt I never figured out mathlib's linter options, like what kind of style we want to follow. Maybe @chenson2018 can help...

berberman:
Updated the text. We also need to make extracted projects depend on batteries (or mathlib later). I think we'll update `SFLMeta.Save`'s  `lakefile.toml` template to add an extra entry. TBH I don't want to touch `SFLMeta.Save` again before we get a chance to merge #62, as I've resolved the conflicts in that file several times...

bcpierce00:
... and indeed Save.lean may get changed again by the versification changes I'll commit soon.

bcpierce00:
Seems not to conflict with #62, so I am going to go ahead and merge

PR #42, opened 2026-06-26
Update cases/induction style in CONTRIBUTING.md

berberman:
This PR updates the style of alternatives as we've discussed earlier. It also clarifies a bit on alignments for inline and multiline respectively (which is open to change).

I plan to add `Batteries` to dependencies once the main branch temporarily switches to nightly verso.

Issue #52, opened 2026-07-01
Inversion doesn't work for empty relations


dsainati1:
```
inductive EmptyRelation : Nat → Nat → Prop where

theorem empty_relation_is_empty : ∀ n m, ¬ EmptyRelation n m := by
  intros n m contra
  inversion contra
```
Throws an error: `Tactic generalizeIndices failed: indexed inductive type expected`

From Zulip discussion:

DHS: Inversion should work for any inductively defined props; its behavior on inductively defined types is unspecified. We are trying to capture the pen-and-paper process of inverting a derivation rule, so any use of inversion that is not that is a conceptual type error. But this case should definitely work, we're inverting an inductively defined relation that has no rules.

JC: it should be straightforward to do both of these things, i.e. skip generalizing indices when there are no indices, and throw an error when the type's sort isn't a Prop

PR #56, opened 2026-07-05
Update code style of Induction and UsingLean

berberman:
This PR updates Induction and UsingLean to make them follow the style introduced in #42. No semantic changes.

### DIFF
--- a/LF/Induction.lean
+++ b/LF/Induction.lean
@@ -898,12 +898,10 @@ theorem mul_two (p : Nat) :
   `add_shuffle3`.  You don't need to use induction yet.
 -/
 
-
-/- ::::full
-Note: By default, `rewrite` and `rw` rewrites left-to-right. To rewrite from right
-to left, use `rw [← h]`, where `←` is typed as `\l` or `\<-`.
-::::
- -/
+/-
+  Note: By default, `rewrite` and `rw` rewrite left-to-right. To rewrite from
+  right to left, use `rw [← h]`, where `←` is typed as `\l` or `\<-`.
+-/
 
 theorem add_shuffle3 : ∀ n m p : Nat,
     add (add n m) p = add (add n p) m := by
@@ -939,6 +937,54 @@ theorem mul_comm (m n : Nat) :
 -- GRADE_THEOREM 2: mul_comm
 -- []
 
+-- TERSE
+/-
+  New tactic combinator: `t₁ <;> t₂` runs `t₁`, then runs `t₂` on every
+  subgoal produced by `t₁`.
+-/
+-- /TERSE
+
+-- FULL
+/-
+  Before moving on to the next batch of exercises, let's introduce one small _tactic combinator_: a way of combining tactics to form a larger tactic.
+
+  If `t₁` and `t₂` are tactics, then `t₁ <;> t₂` means: run `t₁`, then
+  run `t₂` on every subgoal produced by `t₁`.
+
+  This is useful when one tactic splits the goal into several subgoals
+  and all of them can be finished in the same way.
+-/
+-- /FULL
+
+example (b : Bool) : (b || true) = true := by
+  cases b <;> rfl
+
+-- FULL
+/-
+  This is short for:
+-/
+
+example (b : Bool) : (b || true) = true := by
+  cases b with
+  | false => rfl
+  | true  => rfl
+
+/-
+  We can also chain `<;>`s.  In the next example, `cases b` creates two
+  goals; in each of them, `cases c` splits the goal again; then `rfl`
+  solves all four remaining goals.
+-/
+
+example (b c : Bool) : (b && c) = (c && b) := by
+  cases b <;> cases c <;> rfl
+
+/-
+  Use `<;>` when the generated subgoals really do have the same proof.
+  If different branches need different arguments, it is usually clearer
+  to write the cases explicitly.
+-/
+-- /FULL
+
 -- EX3? (more_exercises)
 /-
   Take a piece of paper.  For each of the following theorems, first
@@ -948,6 +994,8 @@ theorem mul_comm (m n : Nat) :
   down your prediction.  Then fill in the proof.  (There is no need
   to turn in your piece of paper; this is just to encourage you to
   reflect before you hack!)
+  Some of these proofs can be shortened
+  with `<;>` when several generated subgoals have the same proof.
 -/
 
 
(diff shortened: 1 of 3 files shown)


LF/Induction.lean
```diff
@@ -111,7 +111,8 @@ theorem review1 : (true || false) = true := by rfl
 -- HIDE
 /- review2 -/
 theorem review2 : ∀ b : Bool, (true || b) = true := by
-  intro b; rfl
+  intro b
+  rfl
```
mwhicks1, line 115: I like this but: Where is the style guideline to put these on separate lines? Is it just that we haven't introduced semi-colon yet? Or, maybe this follows the Lean style guidelines of one tactic per line unless they combine into a single mathematical idea?

LF/Induction.lean
```diff
@@ -444,7 +452,7 @@ theorem add_assoc (n m p : Nat) :
 @[irreducible]
 def double (n : Nat) : Nat :=
   match n with
-  | zero => zero
+  | zero    => zero
```
mwhicks1, line 455: I see that you are doing this to line up the `zero` and `succ n'` righthand sides. But you are not doing this in the examples above. And I don't see this as a guideline in the Lean style guide, but perhaps I missed it. Let's either _not_ do this kind of indentation, or else define a guideline in our CONTRIBUTING.md file (with justification) and do it everywhere. (I'd vote not to do it at all.)

LF/Induction.lean
```diff
@@ -539,10 +545,11 @@ unseal even in
...
     even (succ n) = !even n := by
   -- ADMITTED
-  induction n
-  case zero =>
-    rw [even_zero, even_one]; rfl
-  case succ n' ih =>
+  induction n with
+  | zero =>
+    rw [even_zero, even_one]
+    rfl
```
mwhicks1, line 574: Normally `rw` does `rfl` by default at the end, doesn't it, so do we really need it here?

xhalo32 (dismissed):
Looks good

LF/Induction.lean
```diff
@@ -949,38 +954,39 @@ theorem mul_comm (m n : Nat) :
...
+  | true  => rw [Bool.true_and]
 -- /ADMITTED
 
 theorem all3_spec (b c : Bool) :
     (b && c) || ((!b) || (!c)) = true := by
   -- ADMITTED
-  cases b
-  case true => cases c <;> rfl
-  case false => rfl
+  cases b with
```
mwhicks1, line 1046: Just makes me wonder: Do we have a preferred time/place to introduce `<;>` ? If we haven't done it by this point, this might be a good example to demo it.

LF/UsingLean.lean
```diff
@@ -94,9 +94,10 @@ theorem complicated_computation : (2 * 3 + 4 * 5 : Nat) * 6 = 156 := by
 
   theorem rfl_not_enough (n m : Nat) (h : n = m) : n = m := by
     -- rfl will not work here!
-    /- We need to use `h` to rewrite the goal, and then `rfl` will work. -/
-    rewrite [h]
-    rfl
+    /- We need to use `h` to rewrite the goal, and then `rfl` will work.
+       Equivalently, we can use `rw`:
+    -/
+    rw [h]
```
mwhicks1: Let's show it both ways: Do it the way it's shown above, then show that you can use `rw [h]` as a second (duplicate) example. Then explain what `rw` is doing, compared to `rewrite`. From the Lean documentation: "[rw](https://lean-lang.org/doc/reference/latest/Tactic-Proofs/Tactic-Reference/#rw) is like [rewrite](https://lean-lang.org/doc/reference/latest/Tactic-Proofs/Tactic-Reference/#rewrite), but also tries to close the goal by "cheap" (reducible) [rfl](https://lean-lang.org/doc/reference/latest/Basic-Propositions/Propositional-Equality/#rfl-next) afterwards." Maybe copy this here?

Then update mentions of `rewrite` below to be `rw` (even in comments).

mwhicks1 (changes requested):
Looks good, thank you! A few minor changes requested.

LF/Induction.lean
berberman: I don't think there's a particular style guideline (If I didn't miss it), and I made this change because I thought this would improve the readability a little bit.

LF/Induction.lean
berberman: We do have this style guideline for `induction`/`cases` in our `CONTRIBUTING.md‎`, but not explicitly for `match`, though I think `match` should follow the similar style in general.

I agree this kind of alignment is annoying. I personally don't like it and I never use it in my project... However, it seems to be popular in Lean community, to some extent. I'm not sure about the Mathlib but the Lean 4 codebase has a lot, for example

https://github.com/leanprover/lean4/blob/a54f7e9c7ec212c629363121c0f977d07a8c4eff/src/Lean/AddDecl.lean#L47-L50

Moreover, we also have had this style guideline before I made it clear in #52: "...and **align the `=>`** as above." That's why I followed it. But if most of us agree not to have this alignment, I'm happy to revert all the changes for aligning the arrows and update CONTRIBUTING.md. (I'm on this side.)

LF/Induction.lean
berberman: Lean will complain if the goal was closed when there are more tactics, so I think `rfl` here is necessary.

Here is reason I can tell: strictly speaking, `rw [...]` not equivalent to `rewrite [...]; rfl`. `rw [...]` expands to something like `rewrite [...]; try (withReducible rfl)`, which runs `rfl` in `.reducible` transparent mode (only reducible definitions may be delta expanded). On the other hand, `rfl` by default runs in `.default` transparency mode, which expands more definitions. So it's possible to have a required `rfl` after `rw [...].`

LF/Induction.lean
berberman: Correct me if I'm wrong, I don't think we introduced `<;>` anywhere, so I removed it. We can do it here, or maybe in Tactics?

LF/Induction.lean
mwhicks1: I see that the earlier examples I referred to ("But you are not doing this in the examples above") are multi-line, so that it is consistent in the file. Since this space is what's in CONTRIBUTING.md then that's Ok---no need to re-litigate it now.

LF/Induction.lean
mwhicks1: I don't see it anywhere in the text. I am suggesting that we _can_ introduce it here, but we should add some text to explain what it's doing. That is: Start with the solution being what you have, then show the theorem again but using this tactic for the solution.

mwhicks1 (approved):
Looks great, thank you!

PR #57, opened 2026-07-06
minor changes and comments on Lists

KihongHeo:
I reviewed `Lists.lean`, made a few small changes, and left a comment.


berberman (commented):
I think all `NatList`-related operations should live in the `NatList` namespace. However, there's already a `namespace NatList` declaration at the top of the file, so everything up to `end NatList` is already inside that namespace. Defining functions with the `NatList.` prefix (for example, `NatList.app`) ends up nesting them under `NatList.NatList` instead of `NatList`. Could we fix the namespace usage in this file?

LF/Lists.lean
```diff
@@ -255,17 +256,17 @@ def mylist3 : NatList := [1, 2, 3]
    (We use `myRepeat` because `repeat` is a reserved keyword in Lean.) -/
 
 @[irreducible]
-def myRepeat (n count : Nat) : NatList :=
+def NatList.myRepeat (n count : Nat) : NatList :=
```
bcpierce00: Have we already discussed defining a name "into" a namespace from outside?  This seems like something we might want to postpone.

bcpierce00: Another issue is that `NatList` seems to be used as the name of both a type and a namespace.  Is this an idiomatic Lean thing?  I feel like it might confuse newcomers...

KihongHeo:
It seems that the way we introduce `namespace` in `Basics.lean` is as follows:

https://github.com/plclub/sf-in-lean/blob/cc7cd80161053768390f38eeb0eb1e4892b6bc73/LF/Basics.lean#L375-L387

https://github.com/plclub/sf-in-lean/blob/cc7cd80161053768390f38eeb0eb1e4892b6bc73/LF/Basics.lean#L1170-L1184

So one way to fix the above issues is to move down `namespace NatList` below the type definition. What do you think?

KihongHeo:
@dsainati1 is the above strategy for namespace consistent with your PR #64 ?

dsainati1:
Yeah this seems right to me. Re: @bcpierce00's questions, both of these are things we do mention in Basics, since having namespaces associated with each type does indeed appear to be an idiomatic Lean pattern.

berberman (approved):

bcpierce00:
> It seems that the way we introduce `namespace` in `Basics.lean` is as follows:
> ...
> So one way to fix the above issues is to move down `namespace NatList` below the type definition. What do you think?

The wording here 
```
" The next command opens a new namespace so that our definitions don't 
 clash with ones from the standard library. We'll discuss it in more 
 detail below. "
```
implies that the namespace is being created now, but in fact it was already created by the `inductive` command.  This may make later discussions more confusing.  Recommend being more explicit already here.

KihongHeo:
@bcpierce00 If I understand your recommendation correctly, you mean we need to change the sentences in `Basics.lean`? I am not sure we should add something about namespace in `Lists.lean` since the concept should be introduced in `Basics.lean`.

Now I can see @hgoldstein95 holds a lock (#76). @hgoldstein95 could you take a look at that part?

bcpierce00:
The sentence here is also a bit incorrect, strictly speaking: the namespace is not new at the point where it is described as new.

KihongHeo:
Cleaned up and fixed sentences in `Lists.lean` and `Basics.lean`.
Please let me know if I still missed something.

mwhicks1 (approved):

PR #59, opened 2026-07-06
Incorporate MRC's comments into Basics

dsainati1:


LF/Basics.lean
```diff
@@ -184,6 +195,11 @@ of `nextWorkingDay` is given below.
 
 :::dev
 BCP: I *still* find this quite tangled.  We say Lean requires all uses of constructors to be prefixed with their namespace, but then, right after, we suggest exactly not doing this!  (And I am still confused, myself, about identifiers like `true` and `false`, which seem not to need to be prefixed, even though they come from their own namespace.)
+DHS: I spent some time thinking about how to clean up this explanation, and what I realized
```
dsainati1, line 198: @chenson2018 any thoughts here?

chenson2018: Re `true`/`false`, see [Exporting Names](https://lean-lang.org/doc/reference/latest/Namespaces-and-Sections/#The-Lean-Language-Reference--Namespaces-and-Sections--Namespaces--Exporting-Names) for an explanation.

Past that, I'm not sure I fully understand the confusion. (Not that it's not real or unjustified, just that I'm too used to Lean...). Lean usually resolves names by looking at the current namespace, and this is the single explanation for why whether I'm in `namespace Day` or the body of a function `Day.nextWorkingDay`, I don't need to write the full `Day.monday`. When you use the [leading dot notation](https://lean-lang.org/doc/reference/latest/Terms/Identifiers/#The-Lean-Language-Reference--Terms--Identifiers--Leading--___) you are asking Lean to instead use the expected type's namespace, so that `.monday` used in a place expecting something of type `Day` will resolve to `Day.monday`.

dsainati1: I guess what I don't fully understand is why `.monday` is a thing in the first place, basically. Why shouldn't we just `open Day` and write `monday` instead? Or if we don't want to open the whole namespace, just write `Day.monday`?

LF/Basics.lean
```diff
@@ -1898,6 +2031,8 @@ The generated subgoals are tagged by the names of the constructors.
 `case zero =>` and `case succ n' =>` select which subgoal to work on next
 and introduce variable names.
 
+Note also that when we enter a `case`, we increase the level of indentation at which we are working.
```
chenson2018: If you want to be specific, the style guide specifies two spaces per indent.

LF/Basics.lean
chenson2018: It is just for convenience when the fully qualified name is long and opening the namespace would be awkward for whatever reason. Especially for a proof closing with something like `exact`, it can significantly shorten things because it is aware of the type of the goal.

dsainati1: Maybe the thing to do then is to defer the explanation of the `.monday` style until later, when there is actually a motivating example for why it's nice. So we only use the fully qualified `Day.monday` style until the section about namespaces, then we can explain opening namespaces to get just `monday`, and then even later when we have long names we can explain `.monday`.

chenson2018: This seems fine to me. Leading dot notation is not an essential feature to explain.

bcpierce00: That seems good to me too.

dsainati1: Okay. This is a pretty invasive change so I will make a follow-up PR to this one with that change once this merges.

rogerburtonpatel (approved):
This looks strong. I'll work on other issues while Daniel PRs the `.monday` -> `Day.monday` fix.

LF/Basics.lean
mwhicks1: I added a note to the file itself that captures this discussion, as placeholder for making this change.

mwhicks1 (approved):
Looks great! I added a couple of notes and tweaked the wording in a couple of places.

PR #60, opened 2026-07-06
tweaks to CONTRIBUTING.md to colocate style recommendations

mwhicks1:
Just some tidying of the file. One thing I noticed is that `UsingLean.lean` introduces `rw?` before `rw` is introduced, which is in the file `Induction.lean`, which comes after it. We should fix that. Or, we should reorder `UsingLean.lean` to come after `Induction.lean` -- would that make sense?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -127,68 +127,69 @@ We also have branch protection enabled, which requires the following before merg
 
 ## Cadence of code changes
 
-A general guideline is: Prefer making changes as a PR to talking about them first.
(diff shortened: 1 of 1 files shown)


dsainati1:
`UsingLean` does come after `Induction` I thought? Looking at the file changes it seems like the original file order did have `Induction` first but it was changed in this PR to put `UsingLean` first, which is not correct.

mwhicks1:
Just added a few more changes to the PR to clarify the PR cadence, and addressed the ordering issue (UsingLean comes after Induction).

bcpierce00:
Might want to quickly review my edits?

PR #61, opened 2026-07-06
Started port of HL materials

mwhicks1:
This PR beings to port the first chapter of the HL volume: **`HL/Imp`** covers the Imp language through "Reasoning About Imp Programs" (arith/boolean expressions, commands, evaluation relation). The chapter builds bare and through `to_verso`, has zero `sorry`s. (This PR used to contain other chapters but they have been backed out.)


HL/Imp.lean
```diff
@@ -0,0 +1,2504 @@
...
+   identifiers can now elaborate to strings in a custom grammar -- this may
+   help a lot in PLF!
+   The release notes now include a section on Notation changes. See
+   https://coq.github.io/doc/v8.15/refman/changes.html#id15 for 8.15 and
+   https://coq.github.io/doc/v8.14/refman/changes.html#id19 for 8.14
+   NDS'25: That does not seem right. I could not find anything to that effect
+   in the changelogs, and this hack (from 2022!) also seems to contradict
+   this statement: https://github.com/rocq-prover/rocq/issues/15643 -/
+
+-- MWH (port note): Constructors of `Aexp`/`Bexp` are always written with
```
chenson2018: I disagree with the porting note, see my comments below.

HL/Imp.lean
```diff
@@ -0,0 +1,2504 @@
...
+  boolean expressions.
+-/
+-- /FULL
+-- TERSE: /- Abstract syntax trees for arithmetic and boolean expressions: -/
+
+inductive Aexp where
+  | ANum (n : Nat)
+  | APlus (a1 a2 : Aexp)
+  | AMinus (a1 a2 : Aexp)
+  | AMult (a1 a2 : Aexp)
```
chenson2018: The idiomatic way to write this would be

```suggestion
inductive Aexp where
  | num (n : Nat)
  | plus (a1 a2 : Aexp)
  | minus (a1 a2 : Aexp)
  | mult (a1 a2 : Aexp)
```
It is a Rocq-ism to repeat the name of the type in each constructor; Lean having namespaces means this just adds more clutter. Note also that the constructors are lowercase, in accordance with the naming conventions.

HL/Imp.lean
```diff
@@ -0,0 +1,2504 @@
...
+-/
+
+/- _Evaluating_ an arithmetic expression produces a number. -/
+
+def aeval (a : Aexp) : Nat :=
+  match a with
+  | .ANum n => n
+  | .APlus  a1 a2 => aeval a1 + aeval a2
+  | .AMinus a1 a2 => aeval a1 - aeval a2
+  | .AMult  a1 a2 => aeval a1 * aeval a2
```
chenson2018: The better way to write this is not by using leading dot notation in the match, but making this `Aexpr.eval`.

```suggestion
def Aexp.aeval (a : Aexp) : Nat :=
  match a with
  | num n => n
  | plus  a1 a2 => aeval a1 + aeval a2
  | minus a1 a2 => aeval a1 - aeval a2
  | mult  a1 a2 => aeval a1 * aeval a2
```

This applies broadly to the entire file.

HL/Imp.lean
```diff
@@ -0,0 +1,2504 @@
...
+/- Similarly, evaluating a boolean expression yields a boolean. -/
+
+def beval (b : Bexp) : Bool :=
+  match b with
+  | .BTrue      => true
+  | .BFalse     => false
+  | .BEq a1 a2  => aeval a1 == aeval a2
+  | .BNeq a1 a2 => aeval a1 != aeval a2
+  | .BLe a1 a2  => decide (aeval a1 ≤ aeval a2)
+  | .BGt a1 a2  => decide (aeval a1 > aeval a2)
```
chenson2018: These `decide` are unnecessary.

```suggestion
  | .BLe a1 a2  => aeval a1 ≤ aeval a2
  | .BGt a1 a2  => aeval a1 > aeval a2
```

HL/Imp.lean
```diff
@@ -0,0 +1,2504 @@
...
+  The compound tactic `t <;> t'` first performs `t` and then performs `t'`
+  on _each subgoal_ generated by `t`.
+
+  For example, consider the following trivial lemma.  Splitting on `n`
+  leaves two subgoals that are discharged identically:
+-/
+
+theorem foo (n : Nat) : n = 0 ∨ n ≥ 1 := by
+  cases n with
+  | zero   => omega
```
chenson2018: It's more common to use the newer `lia` nowadays.

HL/Imp.lean
```diff
@@ -0,0 +1,2504 @@
...
+  We have presented `aeval` and `beval` as functions defined by
+  recursion.  Another way to think about evaluation -- one that is often
+  more flexible -- is as a _relation_ between expressions and their
+  values.  This perspective leads to inductive definitions like the
+  following.  We name the hypotheses in each case (`h1`, `h2`); this
+  gives us readable names to refer to during proofs.
+-/
+-- /FULL
+
+inductive AevalR : Aexp → Nat → Prop where
```
chenson2018: I would probably name this `Rel` under an appropriate namespace.

HL/Imp.lean
```diff
@@ -0,0 +1,2504 @@
...
+
+/-
+  We can add variables to the arithmetic expressions we had before simply
+  by including one more constructor.  (This is a fresh `Aexp`, replacing
+  the variable-free one from the `AExp` namespace above.)
+-/
+
+inductive Aexp where
+  | ANum (n : Nat)
+  | AId (x : String)                -- NEW
```
chenson2018: For consistency with with our definitions of maps and what is more common in Lean, I would make this a type variable and assume `DecidableEq` when using it.

HL/Imp.lean
```diff
@@ -0,0 +1,2504 @@
...
+  | .BGt a1 a2  => decide (aeval st a1 > aeval st a2)
+  | .BNot b1    => !beval st b1
+  | .BAnd b1 b2 => beval st b1 && beval st b2
+
+/- We write the empty state (every variable `0`) as `∅`, and reuse the
+   total-map update notation `x →ₜ v ; st` for states. -/
+-- Claude: we write single-variable states inline as `X →ₜ 5 ; empty_st`
+-- rather than introducing a dedicated "singleton state" shorthand.
+
+def empty_st : State := ∅
```
chenson2018: I would not make this definition, it is idiomatic to just use `∅`

HL/Imp.lean
```diff
@@ -0,0 +1,2504 @@
...
+-- /FULL
+
+/-
+  ######################################################################
+  ## Evaluation
+-/
+
+/- _Evaluating_ an arithmetic expression produces a number. -/
+
+def aeval (a : Aexp) : Nat :=
```
chenson2018: Even if in this later volume we'd like to loosen up on the use of `irreducible` (which seems fine to me), I'd still write characterizing lemmas and make these `simp`.

HL/Imp.lean
```diff
@@ -0,0 +1,2504 @@
...
+-- /FULL
+
+/-
+  It will be convenient to have an infix notation for `AevalR`.  We'll
+  write `e ==> n` to mean that arithmetic expression `e` evaluates to
+  value `n`.  (We scope the notation to this namespace so it doesn't
+  collide with other evaluation relations later.)  In Lean the notation is
+  declared right after the inductive.
+-/
+/- HIDE: OLD: (This notation is one place where the limitation to ASCII
```
chenson2018: Here's an example of a comment that seems redundant to retain, especially if the notation is updated to use the desired unicode.

HL/Imp.lean
```diff
@@ -0,0 +1,2504 @@
...
+   Notation "x > y"   := (BGt x y) (in custom com at level 70, no associativity).
+   Notation "x = y"   := (BEq x y) (in custom com at level 70, no associativity).
+   Notation "x <> y"  := (BNeq x y) (in custom com at level 70, no associativity).
+   Notation "x && y"  := (BAnd x y) (in custom com at level 80, left associativity).
+   Notation "'~' b"   := (BNot b) (in custom com at level 75, right associativity).
+
+   Open Scope com_scope.
+   ```
+
+   NOTATION dev notes:
```
chenson2018: Another example. I don't think preserving the details of very Rocq-specific notation issues is terribly helpful.

HL/Imp.lean
```diff
@@ -0,0 +1,2504 @@
...
+-/
+
+/-
+  The compound tactic `t <;> t'` first performs `t` and then performs `t'`
+  on _each subgoal_ generated by `t`.
+
+  For example, consider the following trivial lemma.  Splitting on `n`
+  leaves two subgoals that are discharged identically:
+-/
+
```
berberman: We have `<;>` introduced in Induction per your suggestion (#56) :)

HL/Imp.lean
```diff
@@ -0,0 +1,2504 @@
...
+    - numeric constants, addition, subtraction, and multiplication by
+      constants,
+    - equality (`=`, `≠`) and ordering (`≤`, `<`, `≥`, `>`), and
+    - the logical connectives `∧`, `∨`, `¬`, and `→`,
+
+  then `omega` will either solve it or report that it is false.
+-/
+-- /FULL
+
+/- Claude: `omega` is Lean's analogue of Rocq's `lia`; unlike `lia` (which in
```
chenson2018: Comments like this are simply out of date. I find it frustrating to have to comb through these for inaccuracies.

HL/Imp.lean
```diff
@@ -0,0 +1,2504 @@
...
+/- LATER: Maybe we want to move the discussion of "try solve" from later to
+   here?  It might be helpful for students, but it will make this
+   already-longish chapter a bit longer... -/
+
+theorem silly1 (P : Prop) (hp : P) : P := by
+  try rfl -- `rfl` would fail here, but `try` swallows the failure...
+  exact hp -- ...so we can still finish some other way.
+
+theorem silly2 (ae : Aexp) : aeval ae = aeval ae := by
+  try rfl -- here `try rfl` just does `rfl`
```
berberman: Not specific to this PR, but should we stop using `theorem silly* ...` for one-time examples and using `example ...` instead?  I know this convention comes from Rocq, but Lean's `example` command won't force us to invent a name.

HL/Imp.lean
```diff
@@ -0,0 +1,2504 @@
...
+
+instance (n : Nat) : OfNat Aexp n where
+  ofNat := .ANum n
+
+/- With these, we can write `.APlus 3 (.AMult X 2)` instead of
+   `.APlus (.ANum 3) (.AMult (.AId "X") (.ANum 2))`. -/
+
+def example_aexp : Aexp := .APlus 3 (.AMult X 2)
+def example_bexp : Bexp := .BAnd .BTrue (.BNot (.BLe X 4))
+
```
berberman: Coercion is useful, but I'd prefer something like quasi-quotation-style syntax, e.g. `⟦aexp| ... ⟧`. We can discuss about the notations later.

HL/Imp.lean
```diff
@@ -0,0 +1,2504 @@
...
+  match c with
+  | .CSkip => st
+  | .CAsgn x a => (x →ₜ aeval st a ; st)
+  | .CSeq c1 c2 =>
+      let st' := ceval_fun_no_while st c1
+      ceval_fun_no_while st' c2
+  | .CIf b c1 c2 =>
+      if beval st b then ceval_fun_no_while st c1
+      else ceval_fun_no_while st c2
+  | .CWhile _ _ => st               -- bogus
```
berberman: Since we explained the "correct" implementation would lead to `failed to show termination`, should we just delete this placeholder implementation?

HL/Imp.lean
berberman: I'm not sure if parameterizing the AST over a type variable would lead to any confusion for now... Maybe we just define `abbrev Id := String` and use `Id` everywhere? BTW this also reminds me we should introduce `section` and `variable` command somewhere.

HL/Imp.lean
mwhicks1: Fixed.

HL/Imp.lean
mwhicks1: Fixed.

HL/Imp.lean
mwhicks1: Maybe so, but the CONTRIBUTING.md file had stated we'd be using `omega`, esp. since `lia` seems to be based on `grind`. Nevertheless I made this change. Easy to back out later.

HL/Imp.lean
mwhicks1: I don't think `Rel` is the right name for this, but I did change it to `evalR` under the `Aexp` namespace as with normal `eval`.

HL/Imp.lean
mwhicks1: I made @berberman 's suggestion, and left a note in the file about @chenson2018 's idea.

HL/Imp.lean
mwhicks1: Nice! I've updated.

HL/Imp.lean
mwhicks1: Good call. Fixed.

HL/Imp.lean
mwhicks1: Left a note in the file about doing this; can discuss offline.

HL/Imp.lean
mwhicks1: I'm leaving it for now, simply because it was there in the SF version for a pedagogical purpose. I need to think more and maybe discuss before removing.

HL/Imp.lean
mwhicks1: Addressed.

HL/Imp.lean
mwhicks1: Removed.

HL/Imp.lean
mwhicks1: I would prefer to use `empty_st`, per the original SF materials, as it conveys intent. Changed it to `abbrev` though.

HL/Imp.lean
mwhicks1: I left a note in the file to do that. Would like to discuss the right way to do that, as well as to deal with the various tacticals and whatnot that were introduced in `Imp` before moving ahead.

HL/Imp.lean
mwhicks1: Totally. I removed this and many others like it.

HL/Imp.lean
mwhicks1: Removed.

HL/Imp.lean
chenson2018: Since `State` is an `abbrev` for `TotalMap Ident Nat`, it should consistently use the notation as the canonical way to write the empty map. Making an `abbrev` in the opposite direction from notation to a longer name is straightforwardly unidiomatic in Lean.

mwhicks1: That's fair. I guess the question for @bcpierce00 is why they did it this way in the Rocq version. Maybe `State` should be a `def` -- would that change your opinion on idiomicity?

HL/Imp.lean
```diff
@@ -0,0 +1,2195 @@
...
+  | minus a1 a2 ih1 ih2 =>
+    simp only [Aexp.optimize_0plus, Aexp.eval]
+    rw [ih1, ih2]
+  | mult a1 a2 ih1 ih2 =>
+    simp only [Aexp.optimize_0plus, Aexp.eval]
+    rw [ih1, ih2]
+
+/-
+  ######################################################################
+  # Tactic Combinators
```
dsainati1: I think a lot of this stuff should be moved to the new Automation chapter, probably?

mwhicks1: I completely agree. My plan is to delete it on my next go-around that adds notation, simp annotations, and proofs. It will be in the commit history if you want to refer back to it (in case it saves you any time) when you are adding stuff like it to Automation.

dsainati1: Sounds good. I just wrote up a thread on this in the zulip; I think what makes the most sense is to teach both combinators and `lia` in the automation chapter.

dsainati1 (approved):

PR #62, opened 2026-07-07
Handle Lean code block flags in project extraction

berberman:
This PR introduces two new commands:

- `sf_experiment ... end` -- command in this block won't affect the later code after the experiment ends, which is effectively the -keep in verso.
- `sf_expect_failure ... end` -- at least on command in this block is expected to fail, which is similar to +error. 

With these new commands, verso code blocks get converted from

````lean
```lean +error
example : 1 = 2 := rfl
```
````

to

```lean
sf_expect_failure

  example : 1 = 2 := rfl

end
```

and from

````lean
```lean -keep
def tmp := 1
```
````

to

```lean
sf_experiment

  def tmp := 1

end
```

in project extraction.


SFLMeta/Save.lean
```diff
@@ -237,29 +237,37 @@ private def decodeExercise? (data : Json) : Option (Nat × String) :=
...
+traversal keeps the one selected by the `showSolutions` flag, so the same
+compiled document serves both the student and solutions builds. HTML/TeX
+rendering passes through to the surviving child; the saver checks the stored
+metadata to decide whether to emit the saved source into extracted `.lean`
+files. -/
+
+structure LeanSavedData where
+  teacher : String
+  student : String
+  expectedError : Bool
```
chenson2018: Can we also similarly include `-keep` here?

berberman: Sure, but how would `-keep` affect extraction? A few points in my mind:
1. Remove the automatic `stripGuardMsgs` in favor of the manual flags on use sites.
2. Since `+error` means an intended failure, in project extraction either we emit a `#guard_msgs (drop all)` to suppress the diagnosis or use a command like I described in https://github.com/plclub/sf-in-lean/pull/55#discussion_r3533253669; or alternatively, convert them to docstrings. 
3. For `-keep`, the logic is similar to `+error` -- if we want to have them in the extracted project (which I think we should, for the sake of coherence), either we convert them to docstrings or execute them in sandbox.
4. Not sure about `-show`. They are not rendered in verso manual, but should we keep them in extracted projects anyway?

What's your thought?

chenson2018: 1. I'm not sure I follow, do you mean to extend `Verso.Genre.Manual.InlineLean.LeanBlockConfig` with a new flag for this?
2. The joint use of `+error` and `-keep` makes me think that docstrings might be easiest?
3. same as (2)
4. Do these `show` sometimes need to be kept for setup reasons? I'll trust your judgement here.

Maybe instead of `expectedError : Bool` it is better to make this a `LeanBlockConfig` for consistency?

berberman: Oh for 1 I meant we should use existing `LeanBlockConfig` flags to configure the code block, instead of adding more `#guard_msgs` in verso doc and stripping them off in extraction (which is the current pattern IIRC)

berberman:
Now this exmaple (in Basics)

````lean
```lean -keep
theorem really_bad : one = two := by sorry
```
````

becomes

```lean
-- (commented out as it is for discussion)

-- theorem really_bad : one = two := by sorry
```

in generated projects. Previously it would become normal Lean code, ignoring the `-keep` flag.

Personally I feel turning them into docstrings is not that great, becuase we lost the interactive features which we could have. I guess I still wanted to have a custom command or use `#guard_msgs`, but that would need extra explanations. Maybe @dsainati1 has some thoughts?

chenson2018:
> I guess I still wanted to have a custom command or use `#guard_msgs`

I'm not really opposed to this, I just had trouble imagining multiple flags interacting. I'd say try it out if you're motivated to do so!

rogerburtonpatel:
> > I guess I still wanted to have a custom command or use `#guard_msgs`
> 
> I'm not really opposed to this, I just had trouble imagining multiple flags interacting. I'd say try it out if you're motivated to do so!

How would folks feel about having a custom command like this in this book: 
`#fail`
that unfolds to 
`#guard_msgs(error)`
(perhaps with some extra pretty-printing), 
which also handles the cases mentioned above,

and saying explicitly what we've done? I think @berberman's idea of a custom command is reasonable. I know we want to minimize non-idioms in the book, though I think this is a case where having something that looks nicer and handles edge cases for grading goes a long way and isn't something students will miss in the wild where there is no autograder. 
@chenson2018 thoughts?

bcpierce00:
Let's discuss tomorrow!

xhalo32:
Now `SFLCompat.lean` is extracted into the `LF` directory. Would it make more sense to write `SFLCompat.lean` to another directory like `Meta` instead? Then we could add the following to lakefile.toml

```toml
[[lean_lib]]
name = "Meta"
globs = ["Meta.+"]
```

The `["Meta.+"]` tells lake to not look for `Meta.lean` at the root of the repository.

xhalo32 (approved):
Looks good! I have tested it locally and it works as expected.

One minor issue which I have no idea if it's possible to even fix is that in VSCode hovering over code inside `expect_failure` or `experiment` blocks highlights the entire block rather than the expected subexpression.

Another tiny detail is if I "accidentally" remove `end_experiment` after an `experiment`, and the rest of the file contains syntax declarations (e.g. `local prefix:40 (priority := high) "!" => notb`) then weird errors like this pop up:
```
Type mismatch
  b
has type
  MyBool
but is expected to have type
  Bool
```

berberman:
Thanks for the review, @xhalo32. I think the hover issue is addressed now, and I added `sf_` prefix to the commands to highlight that they are not real Lean commands. I also replaced `experiment_end` and `expect_failure_end` with just `end`.

bcpierce00:
This PR has developed quite a bit, and I'm not 100% sure I understand where it wound up... can someone summarize, or provide instructions on how to build and see an example?

xhalo32:
Good work @berberman ! The hovering works now. The syntax problem I mentioned is not a blocker, and shouldn't cause any confusion.

To test this I did the following steps:
1. Checkout the `verso-error` branch
2. Add the following lines in `LF/Basics.lean`:
    ````lean
    ```lean +error
    example : 1 = 2 := rfl
    ```
    
    ```lean -keep
    def tmp := 1
    ```
    ````
    (e.g. before `inductive Day`)
3. Ran `make lf-student`
4. Opened `_out/lf/student/lean` in the editor and opened `LF/Basics.lean`
5. Played around a bit with the code (adding `#check tmp` inside and outside `sf_experiment` to make sure it doesn't leak; made the `sf_expect_failure` not fail to see the error message it gives out)

I would like to hear someone else's opinion on whether `SFLCompat.lean` should be under `LF` or somewhere else (like `Meta`).

berberman:
> This PR has developed quite a bit, and I'm not 100% sure I understand where it wound up... can someone summarize, or provide instructions on how to build and see an example?

@xhalo32 gave a concrete example above. The idea is that verso supports annotating `lean` code blocks with `+error` flag for demonstrating expected failing examples without requiring us to write `#guard_msgs`. For example, in verso doc we can write something like

````lean
```lean +error
example : 1 = 2 := rfl
```
````

Verso renders this correctly in the book. But we also need to extract standalone Lean projects for student/solutions/terse, and those code blocks can't be emitted directly as they don't compile. So previously we couldn't leverage this verso feature. This PR adds new `sf_experiment ... end ` command block and wraps the content of `+error` verso code block in `sf_experiment ... end` in extracted the project.

Similarly there's another flag `-keep` which makes the verso block non-persistent, i.e. won't affect later code blocks and those blocks marked with `-keep` will be enclosed in `sf_experiment ... end` during extraction.

> I would like to hear someone else's opinion on whether `SFLCompat.lean` should be under `LF` or somewhere else (like `Meta`).

The current implementation copies `SFLMeta/SFLCompat.lean` to `LF/SFLCompat.lean` or `HL/SFLCompat.lean` in extraction. I think this is reasonable -- `SFLMeta/SFLCompat.lean` acts as a source template. We could also move it to `LF` volume like `LF/CustomTactics`, but `SFLMeta.Save` would still need to handle it specially.

**Comment by berberman** [comment]
I discovered a caveat when I was trying to show the code actions in #55:

```lean
sf_expect_failure
def foo (n : Nat) : Nat :=
  match n
end
```

The incomplete `match` won't parse and we can't easily capture the parse errors. One workaround could be that we stop calling the `command` parser in `sf_expect_failure` syntax, but running a string parser and later we run command parser on the collected strings manually. But this would make identifying the `end` complicated. Probably we would need to switch back to `sf_expect_failure ... end_sf_expect_failure` or use some fencing, e.g.
```lean
sf_expect_failure %%{
def foo (n : Nat) : Nat :=
  match n
}%%
```

But anyway, I won't try to solve the issue in this PR.

bcpierce00:
I'm seeing this on the `import` line at the top of the generated Basics:
```
stderr:
✖ [2/3] Running LF.SFLCompat
error: no such file or directory (error code: 4294967294)
  file: /Users/bcpierce/bcp24work/projects/sf-in-lean/LF/SFLCompat.lean
✖ [3/3] Running setup-file /Users/bcpierce/l/_out/lf/student/lean/LF/Basics.lean
error: /Users/bcpierce/l/_out/lf/student/lean/LF/Basics.lean: bad import 'LF.SFLCompat'
Some required targets logged failures:
- LF.SFLCompat
- setup-file /Users/bcpierce/l/_out/lf/student/lean/LF/Basics.lean
Failed to build module dependencies.
```

bcpierce00:
Still not 100% sure about the concrete syntax -- I'll bet we can find clearer keywords, and maybe we don't need the indentation?  But the fundamentals seem ok.

bcpierce00 (approved):

PR #64, opened 2026-07-08
Rework namespaces flow in Basics

dsainati1:
This reworks the presentation of namespaces in Basics. The new flow is:
- Start out with everything being fully qualified, including `Day`, `MyBool`, `Color` and `RGB`. This is consistent with what the Lean InfoView will show students and what VSCode will show them on hover, so it is verbose but minimally confusing.
- In the section about namespaces, explain that inside namespaces you don't have to qualify names in that namespace.
- Explain that you can open or export namespaces so you can use them unqualified outside of the namespace. Explain that this is what Lean does with `Bool`.
- Explain how partially qualified names work using type inference. 

After the namespaces section of the file everything is the same.

### DIFF
--- a/LF/Basics.lean
+++ b/LF/Basics.lean
@@ -990,8 +990,8 @@ Names in the `Bool` `namespace` are available without qualification.
 
 ::::full
 Finally, Lean can often automatically figure out which namespace a qualified name lives in,
-saving us the need to explicitly specify it every time we use the name. In addition the
-to fully qualified style (e.g., `Day.monday`), we can opt for a partially qualified style,
+saving us the need to explicitly specify it every time we use the name. Instead of
+the fully qualified style (e.g., `Day.monday`), we can opt for a partially qualified style,
 writing just `.monday`.
 
 When we do this, Lean tries to resolve the `.monday` name by seeing what its expected type is,
--- a/LF/Basics.lean
+++ b/LF/Basics.lean
@@ -1011,7 +1011,8 @@ and inferring which namespace it must be from based on that type. If there is on
 namespace (i.e., if it is unambiguous which constructor we're referring to), then it will
 automatically resolve to the expected value.
 
-So, for example, we could also write `nextWorkingDay` like so:
+So, for example, we could also write `nextWorkingDay` like so, using the shorter
+style for both the value being matched upon and the value being returned:
 ::::
 
 ::::terse
--- a/LF/Basics.lean
+++ b/LF/Basics.lean
@@ -634,19 +634,19 @@ matches the given type and signal an error if not.
 #check (Bool.not Bool.true : Bool)
 ```
 
-Functions like `Bool.not` are themselves data values, just like `Bool.true`
+Functions like {name}`Bool.not` are themselves ordinary values, just like {name}`Bool.true`
 and `Bool.false`.  Their types are called _function types_, and they are
 written with arrows.
 
 ```lean
-#check not
+#check Bool.not
 ```
 
 ::::full
 The type of `Bool.not`, written `Bool → Bool` and pronounced "`Bool`
 arrow `Bool`," can be read, "Given an input of type `Bool`, this
 function produces an output of type `Bool`." Similarly, the type of
-`and`, written `Bool → Bool → Bool`, can be read, "Given two inputs,
+{name}`Bool.and`, written `Bool → Bool → Bool`, can be read, "Given two inputs,
 each of type `Bool`, this function produces an output of type
 `Bool`."
 ::::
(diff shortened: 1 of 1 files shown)


LF/Basics.lean
```diff
@@ -1039,22 +955,127 @@ def Day.nextWorkingDay'' (d : Day) : Day :=
...
+::::
+
+::::terse
+Lean can often guess which qualified name we mean if we don't supply it explicitly:
+::::
+
+```lean
+def nextWorkingDay' (d : Day) : Day :=
+  match d with
+  | .monday    => .tuesday
```
chenson2018, line 1031: It is maybe worth explicitly pointing out what we're using the leading dot on both the match pattern and the value.

LF/Basics.lean
```diff
@@ -1039,22 +955,127 @@ def Day.nextWorkingDay'' (d : Day) : Day :=
...
+```
+:::
+::::
+
+::::full
+Here, though, because `not` is a function that takes a `Bool` argument, Lean knows that
+`.true` must here be a `Bool`:
+
+```lean
+#check (not .true)
```
chenson2018: Maybe

```suggestion
#check (Bool.not .true)
```
so the student isn't having to think about both `not` being exported at the same time?

chenson2018 (commented):
This is a vast improvement, just a couple of small questions.

LF/Basics.lean
```diff
@@ -662,47 +614,36 @@ introduce new tactics and proof techniques gradually, enriching the propositions
...
 ```lean
 #check not
 ```
 
 ::::full
-The type of `not`, written `Bool → Bool` and pronounced "`Bool`
+The type of `Bool.not`, written `Bool → Bool` and pronounced "`Bool`
 arrow `Bool`," can be read, "Given an input of type `Bool`, this
 function produces an output of type `Bool`." Similarly, the type of
 `and`, written `Bool → Bool → Bool`, can be read, "Given two inputs,
```
berberman:
```suggestion
{name}`Bool.and`, written `Bool → Bool → Bool`, can be read, "Given two inputs,
```

LF/Basics.lean
```diff
@@ -662,47 +614,36 @@ introduce new tactics and proof techniques gradually, enriching the propositions
...
 ```
 
-Functions like `not` are themselves data values, just like `true`
-and `false`.  Their types are called _function types_, and they are
+Functions like `Bool.not` are themselves data values, just like `Bool.true`
+and `Bool.false`.  Their types are called _function types_, and they are
 written with arrows.
 
 ```lean
 #check not
```
berberman:
```suggestion
#check Bool.not
```

LF/Basics.lean
berberman: I would add a bit on associativity, something like "arrows associate to the right, so `Bool → Bool → Bool` means `Bool → (Bool → Bool)`."

LF/Basics.lean
```diff
@@ -662,47 +614,36 @@ introduce new tactics and proof techniques gradually, enriching the propositions
...
 ```lean
-#check (true : Bool)
-#check (not true : Bool)
+#check (Bool.true : Bool)
+#check (Bool.not Bool.true : Bool)
 ```
 
-Functions like `not` are themselves data values, just like `true`
-and `false`.  Their types are called _function types_, and they are
+Functions like `Bool.not` are themselves data values, just like `Bool.true`
```
berberman:
```suggestion
Functions like {name}`Bool.not` are themselves ordinary values, just like {name}`Bool.true`
```
_Data value_ doesn't seem to be a standard term to me. Did we define it somewhere?

LF/Basics.lean
dsainati1: My only concern with this is this conflicts with our previous explanation of `Bool → Bool → Bool` as a function that takes two inputs and gives a `Bool`. It seems like we are electing not to explain currying/uncurrying at this point in the text (this explanation comes directly from the Rocq), so maybe we should delay explaining the arrow associativity for when we talk about currying and partial application.

berberman (approved):

LF/Basics.lean
```diff
@@ -1039,22 +955,128 @@ def Day.nextWorkingDay'' (d : Day) : Day :=
...
 
 ```lean
 namespace MyNamespace
-def myDef : Bool := true
+def myDef : Bool := Bool.true
 end MyNamespace
 
 open MyNamespace
 
 #check myDef -- Bool
```
bcpierce00, line 969: Are we (still) commenting the @check results?

bcpierce00:
This looks wonderful.  Adding a bunch of tweaks and comments in a new commit.

LF/Basics.lean
dsainati1: Unsure. I did this because it seemed like the convention in the file, but I guess I don't know if we want to keep that.

PR #70, opened 2026-07-09
BCP versification

bcpierce00:
This PR doesn't actually versify any of the current .lean files, but it updates those files and the related infrastructure so that everything up to (but not including) IndProp now builds in verso format.


bcpierce00:
Sorry -- this was not ready for PRing

PR #71, opened 2026-07-09
Bcp versification

bcpierce00:
Try again, after resolving some conflicts...


LF/Logic.lean
```diff
@@ -1732,12 +1633,16 @@ theorem nonzero_bool_prop (n : Nat) :
 
     Again, these two notions are equivalent: -/
 
+/- (For the reverse direction we need the simple fact that `==` is
+    reflexive.) -/
+theorem beq_refl (n : Nat) : (n == n) = true := decide_eq_true rfl
```
dsainati1, line 1640: Is this a tactic we have explained?

LF/Maps.lean
```diff
@@ -231,3 +235,11 @@ theorem update_subset (m₁ m₂ : PartialMap α β) (a : α) (b : β) (h : m₁
     assumption
 
 end PartialMap
+
+-- BCP: The following was added by Claude to repair a build error in
```
dsainati1, line 239: Surfacing this for review by @chenson2018 I guess?

LF/Tactics.lean
```diff
@@ -1399,11 +1401,20 @@ theorem double_injective_take2 : ∀ n m,
 -- ######################################################
 -- Rewriting with conditional statements
 
+/- We'll use a boolean "less or equal" test on numbers, written `n ≤? m`
+    (the library function `Nat.ble`), together with the fact that it
+    commutes with successor on both sides. -/
+
+-- BCP: Added, to make the file compile, on Claude's suggestion. But is this the right way?
+infix:52 " ≤? " => Nat.ble
```
dsainati1, line 1410: We elected not to add this notation way back in basics because it apparently does not exist in Lean. So I think the appropriate fix would be to replace the `≤?`s with `Nat.ble`s

LF/Logic.lean
chenson2018: I don't see a tactic here? But I agree that we should use a proof specialized to `Nat` here, as this currently uses decidability.

LF/Tactics.lean
berberman: I think we can figure this out together later and improve the chapter materials in follow-up PRs. IMO this PR is already getting  sufficiently large...

LF/Maps.lean
bcpierce00: Yes, but I'm hoping this PR gets merged in the next few minutes, so we should also ping him elsewhere... Not sure if this will get his attention after merging...

LF/Logic.lean
bcpierce00: Will leave a query in the text

berberman:
I think we should merge this asap to avoid blocking future changes.

(The will conflict with #62, as I'm holding the lock on SFLMeta.Save, but I can resolve them in #62 after this gets merged.)

dsainati1 (approved):
accepting to unblock then

LF/Maps.lean
chenson2018: Hmm, removing an instance is usually local so I'd not expected any issues. This seems fine for now, I'll make a mental note to check this as I move this into typeclasses.

PR #72, opened 2026-07-09
Adds notation to Imp

mwhicks1:
This PR adds Imp concrete syntax and uses it throughout HL/Imp
    
It adds `aexp { … }`, `bexp { … }`, and `imp { … }` notation for Imp
arithmetic/boolean expressions and commands, following the ssft24 Imp
recipe (per-category syntax with an embedding hook and a `~` Lean escape).
See https://github.com/david-christiansen/ssft24.
    
Key design points:
- `true`/`false` and `skip` are recognized via a bare-identifier form plus a
  macro name-check, avoiding reserving them as global keywords (which would
  break ordinary Lean use and existing `match`/`induction` on `Com.skip`).
- Variable references elaborate to `Aexp.id x` with the identifier spliced
  as a term (matching hand-written terms and Rocq's `AId X`), so variables
  must be declared `Ident` constants (W/X/Y/Z).
    
I followed the ssft24 layer split for the use of notation in pattern matches
and proofs: Expr uses constructor patterns, Stmt uses notation. Proofs
still do not use simp + characterizing lemmas. That will happen in a
subsequent PR.

### DIFF
--- a/HL/Imp.lean
+++ b/HL/Imp.lean
@@ -414,11 +414,6 @@ inductive Aexp.evalR : Aexp → Nat → Prop where
 -- /FULL
 
 /-
-  mwhicks1: We will very likely want to use different notation, both
-  here and for defining AExp and BExp terms themselves.
--/
-
-/-
   It will be convenient to have an infix notation for `Aexp.evalR`.  We'll
   write `e ⇓ n` to mean that arithmetic expression `e` evaluates to
   value `n`.  (We scope the notation to this namespace so it doesn't
@@ -545,7 +540,7 @@ scoped notation:55 e:56 " ⇓ " n:56 => Aexp.evalR e n
 -- /HIDE
 
 /-
-  mwhicks1: Not sure if we need ⇓b is needed, or whether we can define
+  mwhicks1: Not sure if we need ⇓b, or whether we can define
   ⇓ overloaded. Don't understand Lean notation yet!
 -/
 
@@ -937,12 +932,6 @@ inductive Bexp where
   | not (b : Bexp)
   | and (b1 b2 : Bexp)
 
-/-
-  mwhicks1: SHould we be defining variables as lowercase letters, rather
-  than uppercase ones? Maybe notational conventions in Lean should be
-  different.
--/
-
 /- Defining a few variable names as shorthands will make examples easier
    to read. -/
 /- INSTRUCTORS: We usually don't use x as a "bare identifier" in examples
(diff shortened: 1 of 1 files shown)


HL/Imp.lean
```diff
@@ -1053,13 +1153,13 @@ def Bexp.eval (st : State) (b : Bexp) : Bool :=
 /- We reuse the total-map notation (`x →ₜ v ; ∅` etc.) for states. -/
 
 /- test_aexp1 -/
-example : Aexp.eval (X →ₜ 5 ; ∅) (.plus 3 (.mult X 2)) = 13 := by rfl
+example : Aexp.eval (X →ₜ 5 ; ∅) (aexp { 3 + (X * 2) }) = 13 := by rfl
```
dsainati1: Bikeshedding: I'm not sure how I feel about this arrow subscript for maps. Easy to change later but just flagging to discuss.

dsainati1 (approved):
Nice. The notation looks good as static text, but after this merges I want to play with it and see how it feels during proofs.

HL/Imp.lean
mwhicks1: This comes from the Maps chapter, which @chenson2018 is working on. Thoughts on alternative notation? I wonder if there is a keyboard shortcut for ↦ we might use?

mwhicks1: Gonna resolve this thread and merge; can discuss on Zulip (left a note in the file itself).

chenson2018: I'm open to alternative suggestions. I chose this because having `→ₜ` and `→ₚ` seemed like a decent way to distinguish between partial and total maps that I liked better than what is done in the original.

`↦` does have (several) keyboard shortcuts, but this is problematic to use as this is the unicode for functions, e.g.
```lean
/-- info: fun x => x + 1 : Nat → Nat -/
#guard_msgs in
#check λ x : Nat ↦ x + 1
```

(If you are wondering why it is possible to write `fun`/`λ` and `=>`/`↦` interchangeably, this is a bit of a historical quirk IMO that doesn't have a strict rule in the style guide)

HL/Imp.lean
```diff
@@ -979,18 +971,124 @@ def Z : Ident := "Z"
   ## Notations
 -/
 
+-- FULL
 /-
```
chenson2018, line 980: Can you also include the corresponding delaborators that appear in David's repo? There are in the `Imp/*/Delab.lean` files. We can hide these and don't need to discuss in the text beyond a terse "these functions are helpful in making the printing of these types in the infoview match our custom syntax".

mwhicks1: Updated in https://github.com/plclub/sf-in-lean/pull/72/commits/002a91a8797e27becfe8acacccc968c4173c6cac. Let me know if it seems Ok and I'll merge.

chenson2018 (approved):
The addition of the delaborators looks good to me. I also pushed a few very minor formatting changes to make better use of dot notation.

There are two things I'd like to make sure we do in follow up PRs:

- As soon as possible move this into Verso format. I thought this was originally going to be done in the PR adding this file, but seems to have not happened. With Benjamin wrapping this up for LF, *please* do not add any more new non-Verso files into the repo. This is a completely avoidable time sink.
- There is still some naming  weirdness in the relation forms of evaluation that should be addressed later.

PR #76, opened 2026-07-10
Comments and polish in LF/Basics

hgoldstein95:
This isn't ready for review yet, but I'm putting it here in case folks want to look at it.

### DIFF
--- a/LF/Basics.lean
+++ b/LF/Basics.lean
@@ -340,7 +340,7 @@ Functions over booleans can be defined in the same way as above
 ::::
 
 ```lean
-def notb (b : MyBool) : MyBool :=
+def not (b : MyBool) : MyBool :=
   match b with
   | MyBool.true => MyBool.false
   | MyBool.false => MyBool.true
@@ -350,12 +350,12 @@ def notb (b : MyBool) : MyBool :=
 :::
 
 ```lean
-def andb (b1 : MyBool) (b2 : MyBool) : MyBool :=
+def and (b1 : MyBool) (b2 : MyBool) : MyBool :=
   match b1 with
   | MyBool.true => b2
   | MyBool.false => MyBool.false
 
-def orb (b1 : MyBool) (b2 : MyBool) : MyBool :=
+def or (b1 : MyBool) (b2 : MyBool) : MyBool :=
   match b1 with
   | MyBool.true => MyBool.true
   | MyBool.false => b2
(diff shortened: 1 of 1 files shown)


LF/Basics.lean
```diff
@@ -1737,69 +1764,75 @@ GRADE_THEOREM 2: test_mult1
...
 mean it: even testing equality is a user-defined operation!
 
-Here is a function `beq` that tests natural numbers for
+:::dev
+HG: I don't like this framing. It sounds like we're telling students they'll have to build
+everything themselves. I think what we really mean is "Lean is extremely flexible; even equality
+testing can be defined by the user." or something like that.
+:::
+
+Here is a function `eq` that tests natural numbers for
```
dsainati1: These ones I think should still have the `b` prefix (per https://lean-lang.org/doc/reference/latest/Basic-Types/Natural-Numbers/#Nat___beq)

LF/Basics.lean
```diff
@@ -363,28 +374,32 @@ def orb (b1 : MyBool) (b2 : MyBool) : MyBool :=
...
+complete specification -- a truth table -- for the `or` function:
 ::::
 
 :::terse
-Note the syntax for defining multi-argument functions (`andb` and `orb`).
+Note the syntax for defining multi-argument functions (`and` and `or`).
 :::
 
+:::dev
+HG: I think this is the right place (or just before this?) to introduce `.true` as sugar for
```
chenson2018: This is a recent change Daniel made to make things more explicit until the `## Namespaces` section that gives more explanation. If this changes back, I think we need think about moving that section up here and see if it fits well.

I personally think it's good to have a little bit of time with the full name. I agree it's clunky, but I think immediately throwing the student into the deep end of "this namespace is inferred by the expected type" is also rough.

LF/Basics.lean
```diff
@@ -1737,69 +1764,75 @@ GRADE_THEOREM 2: test_mult1
 When we say that Lean comes with almost nothing built-in, we really
 mean it: even testing equality is a user-defined operation!
 
-Here is a function `beq` that tests natural numbers for
+:::dev
+HG: I don't like this framing. It sounds like we're telling students they'll have to build
```
chenson2018: Agreed, though I do think the first mention of this "Naturally, Lean also comes with an extensive standard library..." does get the nuance across. Maybe we mirror this and say something like

> When we say that Lean comes with almost nothing built-in, we really
mean it: even testing equality is not a primitive operation, but an ordinary function that we could re-implement ourselves as users.

chenson2018 (commented):
Minus Daniel's comment, this looks good to me. Comments on the notes you left:

LF/Basics.lean
hgoldstein95: IIRC the Lean books just say "sometimes you can omit the thing before the dot, we'll explain why later" and that kind of does the job. But we can leave it as-is if we want.

mwhicks1: I'd rather stick with what we have. Too much thrashing.

dsainati1 (approved):

dsainati1:
This PR's changes to the names of the bool functions broke the build in `Induction`. Why does our CI not catch this?

Edit: https://github.com/plclub/sf-in-lean/pull/84 should fix this problem for the future.

PR #78, opened 2026-07-10
Permanent conversion of HL/Imp.lean to verso

mwhicks1:
I ran `to_verso.py` to convert `HL/Imp.lean` to verso format, and then
1. scrutinized the result, to make sure no content was lost
2. made sure everything builds properly, solutions are hidden/revealed, files render properly
3. dropped the non-verso Imp file and replaced it with the Imp one, and updated the Makefile accordingly

In this process I found that SFLMeta/Save.lean was missing some functionality: It needs to be able to import files from other volumes (i.e., LF, for LF/Maps.lean). So I updated that. I also found that to_verso.py was mis-translating some directives, so I fixed those too. Mostly around handling exercise solutions.

I used Claude to help with the script code changes.

### DIFF
--- a/HL/Imp.lean
+++ b/HL/Imp.lean
@@ -28,7 +28,7 @@ htmlSplit := .never
 file := some "Imp"
 %%%
 
-```instructors
+:::instructors
 This chapter plus `Maps` takes a little more than one
    80-minute lecture.  It could be streamlined a bit further without
    losing much, by removing (for example) the inference rules and BNF
@@ -37,9 +37,9 @@ This chapter plus `Maps` takes a little more than one
    (BCP 21: ... Actually, I tried removing inference rules from the
    TERSE version; eventually decided that it makes some of the
    definitions harder to talk about.)
-```
+:::
 
-```dev
+:::dev
 SOONER: Needs some WORKINCLASSes and some quizzes
 
 LATER: Another nice challenge exercise at some point would be to add
@@ -48,9 +48,9 @@ LATER: Another nice challenge exercise at some point would be to add
    aliasing / etc.).
 
 SOONER: BCP 25: Maybe we should write /\ instead of && in assertions,
-   to save a mismatch in the dec_minimum exercise in Hoare2?
+   to save a mismatch in the `dec_minimum` exercise in Hoare2?
 
-HIDE: At some point we could consider moving material from the old
+   At some point we could consider moving material from the old
    HoareLists to this chapter (and into later files, as
    appropriate).  We haven't done it yet because it's a shame to
    complicate the nice simple presentation here when it's used as the
@@ -66,7 +66,7 @@ below is a recap.  For linear arithmetic we use `lia`;
 NOTE that LF currently
 introduces `omega`, not `lia`, so this needs to be reconciled volume-wide
 (either introduce `lia` in LF, or keep `omega`).
-```
+:::
 
 ::::full
 In this chapter, we take a more serious look at how to use Lean as a
@@ -92,12 +92,12 @@ _Hoare Logic_, a popular logic for reasoning about imperative programs.
 
 # Arithmetic and Boolean Expressions
 
-```dev
+:::dev
 SOONER: At this point, I usually take some of the lecture time to
    give a high-level picture of the structure of an interpreter, the
    processes of lexing and parsing, the notion of ASTs, etc.  Might be
    nice to work some of those ideas into the notes. - BCP
-```
+:::
 
 ::::full
 We'll present Imp in three parts: first a core language of _arithmetic
@@ -109,7 +109,7 @@ sequencing, and loops.
 ## Syntax
 
 ```lean
-namespace AExp
+namespace Warmup
 ```
 
 ::::full
@@ -138,19 +138,15 @@ inductive Bexp where
   | and (b1 b2 : Bexp)
 ```
 
-```dev
-MWH: Will we develop `ImpParser`? Mentioned below as an optional chapter
-```
+:::dev
+  SOONER: mwhicks1: Will we develop `ImpParser`? Mentioned below as an optional chapter
+:::
 
 ::::full
 In this chapter, we'll mostly elide the translation from the concrete
 syntax that a programmer would actually write to these abstract syntax
 trees -- the process that, for example, would translate the string
-`"1 + 2 * 3"` to the AST
-
-```
-.plus (.num 1) (.mult (.num 2) (.num 3))
-```
+`"1 + 2 * 3"` to the AST `.plus (.num 1) (.mult (.num 2) (.num 3))`.
 
 The optional chapter `ImpParser` develops a simple lexical analyzer and
 parser that can perform this translation.  You do not need to understand
@@ -209,12 +205,12 @@ implementations and proofs.
 
 _Evaluating_ an arithmetic expression produces a number.
 
-```dev
+:::dev
 chenson2018: TODO: seal evaluators with `@[irreducible]` and prove
    *characterizing lemmas* (one
    `rfl` equation per constructor) that proofs rewrite with instead of
    unfolding the definition; tag lemmas `@[simp]`.
-```
+:::
 
 ```lean
 def Aexp.eval (a : Aexp) : Nat :=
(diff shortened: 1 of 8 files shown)


HL/Imp.lean
```diff
@@ -1,102 +1,127 @@
...
 
-/- INSTRUCTORS: This chapter plus `Maps` takes a little more than one
+#doc (Manual) "Imp: Simple Imperative Programs" =>
+%%%
+tag := "Imp"
+htmlSplit := .never
+file := some "Imp"
+%%%
+
+```instructors
```
chenson2018: The decision to make these available as both codeblocks and directives does not make sense to me. I would change these back to using `:::` and then remove the ability to write these as codeblocks.

(cc @bcpierce00 who made this change in #71 in case I'm missing something, but I'm having trouble thinking of of a good reason for this)

HL/Imp.lean
```diff
@@ -303,75 +323,83 @@ theorem optimize_0plus_sound (a : Aexp) :
...
-/-
-  Since the `Aexp.optimize_0plus` transformation doesn't change the value of an
-  `Aexp`, we should be able to apply it to all the `Aexp`s that appear in a
-  `Bexp` without changing the `Bexp`'s value.  Write a function that
-  performs this transformation on `Bexp`s and prove it sound.  Use the
-  combinators we've just seen to make the proof as short and elegant as
-  possible.
--/
+:::::exercise (rating := 3) (name := "optimize_0plus_b_sound")
+Since the `Aexp.optimize_0plus` transformation doesn't change the value of an
```
chenson2018: Not asking you to do this now, but it would be nice now that we'll soon have everything in Verso format to start using some of the features that make the HTML nicer. As an example, you can do

```suggestion
Since the {name}`Aexp.optimize_0plus` transformation doesn't change the value of an
```
so that you are able to click and see the type of existing theorems.

HL/Imp.lean
```diff
@@ -384,149 +412,148 @@ inductive Aexp.evalR : Aexp → Nat → Prop where
...
--- chenson2018: There is still some naming weirdness in the relation forms of evaluation that should be addressed later.
--- mwhicks1: I think this comment is about wanting to call `evalR` something like `Rel`, but it's not clear to me that I want to do that, e.g., since there are multiple evaluation relations (such as big-step and small-step) there won't be just one name
+```dev
+chenson2018: There is still some naming weirdness in the relation forms of evaluation that should be addressed later.
 
--- FULL
-/-
-  A small notational aside. We could instead have presented this relation
-  with *positional* hypotheses -- no names for the premises:
+mwhicks1: I think this comment is about wanting to call `evalR` something like `Rel`, but it's not clear to me that I want to do that, e.g., since there are multiple evaluation relations (such as big-step and small-step) there won't be just one name
```
chenson2018: What I meant is the naming convention for both the inductive and its constructors. The inductive should be in upper camel case (you can disambiguate which evaluation with its name) and the constructors should not have these `E_A` prefixes.

HL/Imp.lean
```diff
@@ -111,88 +136,87 @@ inductive Bexp where
...
+
+::::full
+In this chapter, we'll mostly elide the translation from the concrete
+syntax that a programmer would actually write to these abstract syntax
+trees -- the process that, for example, would translate the string
+`"1 + 2 * 3"` to the AST
+
+```
+.plus (.num 1) (.mult (.num 2) (.num 3))
+```
```
chenson2018: Seems meant to be a code block:
````suggestion
```lean
#check Aexp.plus (.num 1) (.mult (.num 2) (.num 3))
```
````

File: HL/Imp.lean [hunk 5]
```diff
@@ -1,102 +1,127 @@
...
+::::
 
-/-
-  ######################################################################
-  ## Syntax
--/
+## Syntax
 
+```lean
 namespace AExp
```
chenson2018: I'm not going to attempt to untangle it, but this namespace is definitely misplaced. You've defined things like `AExp.Bexp` below.

HL/Imp.lean
```diff
@@ -542,90 +569,89 @@ scoped notation:55 e:56 " ⇓ " n:56 => Aexp.evalR e n
...
-  definitions of evaluation agree.
--/
--- /HIDEFROMADVANCED
-
-/- SOONER: BCP 23: Why can't we do induction on H in the ← direction?? -/
+:::::exercise (rating := 1) (name := "beval_rules")
+Here, again, is the definition of the `Bexp.eval` function:
 
+```
+def Bexp.eval (b : Bexp) : Bool :=
```
chenson2018, line 581: We should discuss a way to recall definitions without having to write them out manually like this. I think a simple `#print` may work, but it's currently hard to tell because of the namespace issues.

chenson2018 (commented):
I've taken a relatively brisk look; this looks okay for the most part. Some of my comments that are out of scope for Versofication could be in a follow-up PR.

HL/Imp.lean
bcpierce00: This was not intentional! -- must have been a consequence of something else I (or Claude) did.  100% agree that these should be directives.

SFLMeta/Save.lean
```diff
@@ -753,6 +761,72 @@ private def buildProject (dest : System.FilePath) (kind : String) :
...
+        -- Not a bundleable source file (e.g. a spurious match); skip it.
+        bundleLoop needsBundle rest (m :: seen) files libs
+      else
+        let content ← IO.FS.readFile path
+        let top := modTop m
+        let libs := if libs.contains top then libs else libs.push top
+        let deps := (headerImports content).toList.filter needsBundle
+        bundleLoop needsBundle (rest ++ deps) (m :: seen)
+          (files.push (path, content)) libs
+
```
berberman, line 829: This feels a bit magical to me. For Imp, we know the generated HL project needs to import `LF.Maps` and add `LF` as an extra `lean_lib`. Can we declare this explicitly somewhere near HL.lean or Targets.lean instead of computing it by scanning the import commands from source file? Copying files and adding extra `lean_lib` are fine IMO, but I'd rather have a dependency list directly.

berberman (commented):
Left a comment on SFLMeta.Save. I'm okay with merging this part for now, though I don't think it's a sustainable design. We can improve it in the future anyway.

HL/Imp.lean
mwhicks1: I wondered about this myself! Ack about converting them to directives.

HL/Imp.lean
mwhicks1: I can do this now. I had wondered what `{name}` does when I saw it in others' PRs ...

SFLMeta/Save.lean
mwhicks1: I originally had this information in `Targets.lean` as you suggested, but it seemed more fragile to me to specify dependencies in two places. We have them in the original `HL/Imp.lean` file, so why not use those when deriving these other versions of the file? I agree this is a lot of code, but I imagine it could serve us well to avoid wasted work in the longer term.

HL/Imp.lean
mwhicks1: I think it is not meant to be a code block; see the text that precedes it. I've moved the code into the paragraph.

HL/Imp.lean
mwhicks1: This is just badly named. I'll rename it `Warmup`. It's there to distinguish `Aexp` and `Bexp` definitions that have no IDs in them (the start) from those that do (outside a namespace, later).

HL/Imp.lean
mwhicks1: Updated the scripts to hopefully fix this.

HL/Imp.lean
mwhicks1: Added some markup. Also updated `CONTRIBUTING.md` about what's available, and left notes at the end of this file for more ideas of what to do.

HL/Imp.lean
mwhicks1: Added note in the file to consider in the future.

dsainati1 (approved):

chenson2018 (approved):
Approving modulo me not not looking at much outside `Imp.lean`

SFLMeta/Save.lean
berberman: I see. My concern is about the dependency computation from textual source scanning, and there could be some edge cases. For example, I checked the branch locally, and it seems that when HL/Imp.lean imports another file in HL, e.g. HL/Foo.lean, the extracted project will fail because that `import HL.Foo` will generate a duplicate `lean_lib` entry in lakefile. Maybe there are more complications (`public meta import`?), I'm not sure yet. So I agree it's nice to have the dependencies automatically derived, but I think we might still need some explicitly.

berberman (approved):
Let's merge this first.

SFLMeta/Save.lean
mwhicks1: Sounds good. We can improve as we go.

PR #79, opened 2026-07-13
Tell the story of `irreducible`, `rewrite`, and automation from a pedagogical and practical perspective.

rogerburtonpatel:
I've written sections in `Basics` and `UsingLean` to address this building on @dsainati1's prior descriptions.


LF/Basics.lean
```diff
@@ -1525,16 +1522,41 @@ changes the proof state and hovering over each argument to `rewrite` to see its
 The definitions and proofs above use a few somewhat mysterious conventions:
 we write `@[irreducible]` above some of our definitions, and we
 write `unseal` before some of our proofs and `seal` after them.
-These are not things you will usually see in real Lean developments;
-however, we use them in this book to enforce a particular convention
-to help you build good Lean habits.
+We use `@[irreducible]` and `seal` to prevent a function like `add`
```
dsainati1: I believe we had wanted to experiment with changing this to setting the definition after defining the function and its simplification rules, so this way we don't ever need to `unseal`. IMO if we are going to be modifying the wording here we should also update the annotations to this lighter weight version?

LF/Basics.lean
```diff
@@ -1525,16 +1522,41 @@ changes the proof state and hovering over each argument to `rewrite` to see its
...
+  rewrite [add_zero] -- After each line, you can observe the proof state after this step.
+  rewrite [add_zero]
+  rfl
+```
+
+The purpose of this style is twofold:
+
+1. Pedagogically, executing computation one step at a time allows you as a student
+  to manipulate and examine these steps at a fine-grained level.
+  This will give you a better understanding of how functions compute in Lean,
```
```suggestion
  This will give you a better understanding of how functions compute in Lean
```

LF/Basics.lean
```diff
@@ -1525,16 +1522,41 @@ changes the proof state and hovering over each argument to `rewrite` to see its
...
+  and give you more control over this computation as you write your proofs.
+2. Practically, it prepares you to write more sophisticated Lean proofs on structures
+  that are _only_ accessible through simplification rules and have no accessible
+  computable properties at all.
+  In real-world Lean developments, this is the standard.
```
dsainati1: Is this true? I was under the impression that the issue was mostly that typeclasses block simplification, but if you unfold enough of them you can eventually `dsimp` the underlying definition.

LF/Basics.lean
```diff
@@ -1525,16 +1522,41 @@ changes the proof state and hovering over each argument to `rewrite` to see its
...
 
 Lean, like any other programming language, has conventions and best practices
-for writing good software. You are probably familiar with object oriented programming,
+for writing good software. You are probably familiar with Object-Oriented programming,
```
dsainati1: why capitalize this?

LF/Basics.lean
```diff
@@ -1626,6 +1646,22 @@ with their simplification lemmas. After proving these lemmas, instead of using `
 to peek through the definitions, we will prefer rewriting
 by the lemmas, using `@[irreducible]` to enforce this policy,
 and only `unseal`ing the definition in the proofs of those lemmas themselves.
+
+Eventually, we will introduce a way to _automatically_ derive
+simplification rules from a function, and then a way to automatically _apply_ them.
+Real Lean programmers do this using tactics like `dsimp`, `simp`, and `grind`.
```
dsainati1: Unsure if we are actually doing this. Per the discussion in last week's meeting I don't think we want to actually put `[@simp]` on function definitions?

LF/Basics.lean
```diff
@@ -1626,6 +1646,22 @@ with their simplification lemmas. After proving these lemmas, instead of using `
...
+
+For now, though, these tactics are forbidden by our autograder.
```
dsainati1: IMO we should just not mention the tactics at all if we don't want students using them.

LF/UsingLean.lean
```diff
@@ -370,6 +371,43 @@ example (n : Nat) : square n + 0 = n * n := by
...
+   small amount of its own automatic reasoning using other basic
+   tactics like `rfl`. If you're ever confused by what `dsimp` is
+   doing, don't be afraid to switch back to `rewrite` to examine
+   what's going on.
+
+   Later in this volume, we will introduce the more powerful automated
+   tactic `simp`, and in following volumes we introduce the even-more
+   powerful `grind`. These are tactics that can sometimes solve entire
+   complex proofs by themselves, and are correspondingly extremely
+   common in real-world Lean developments.
```
dsainati1: Same as above, I see no reason to tempt students with powerful shiny tools they can't use at this point.

LF/Basics.lean
rogerburtonpatel: Let's update once the lighter-weight version is integrated.

LF/Basics.lean
```diff
@@ -1525,16 +1537,41 @@ changes the proof state and hovering over each argument to `rewrite` to see its
...
+
+The purpose of this style is twofold:
+
+1. Pedagogically, executing computation one step at a time allows you as a student
+  to manipulate and examine these steps at a fine-grained level.
+  This will give you a better understanding of how functions compute in Lean
+  and give you more control over this computation as you write your proofs.
+2. Practically, it prepares you to write more sophisticated Lean proofs on structures
+  that are _only_ accessible through simplification rules and have no accessible
+  computable properties at all.
```
rogerburtonpatel:
```suggestion
2. Practically, it prepares you to write more sophisticated Lean proofs on structures
  that are _only_ meant to be accessed through simplification rules, and on which trying to force computation to run all the way can have disastrous results on performance and readability. 
```

LF/Basics.lean
rogerburtonpatel: Addressed!

chenson2018: This is true of `noncomptable` or unexposed definitions for instance.

LF/Basics.lean
rogerburtonpatel: my fault, used to something that was wrong

LF/Basics.lean
```diff
@@ -1525,16 +1537,41 @@ changes the proof state and hovering over each argument to `rewrite` to see its
...
+  In real-world Lean developments, this is the standard.
 
 Lean, like any other programming language, has conventions and best practices
-for writing good software. You are probably familiar with object oriented programming,
+for writing good software. You are probably familiar with Object-Oriented programming,
```
rogerburtonpatel:
```suggestion
for writing good software. You are probably familiar with object-oriented programming,
```

LF/Basics.lean
rogerburtonpatel: Ugh, I keep getting a different story on this. What are we doing later on with `simp`? Are we eventually going to autogenerate `simp` lemmas or no?

LF/Basics.lean
rogerburtonpatel: Let's discuss in our meeting.

LF/Basics.lean
chenson2018: Yeah, this is a bit tricky. Often the attribute goes on the `rfl` lemmas. Maybe let's leave a comment to revisit this after you've had your first pass at the automation chapter.

LF/Basics.lean
dsainati1: Still not sure of the wording on this one. Written this way IMO it begs the question of what we could possibly be talking about. Can we just cut the whole thing after the comma and leave it at that?

bcpierce00 (commented):
Did you intend to delete `settings.json`?  (Why?)

LF/Basics.lean
bcpierce00: Agree that we don't want to mention them by name.

bcpierce00:
Since Roger is going to be away next week, could one of @hgoldstein95 or @dsainati1 have a look at this, repair conflicts, and get it merged? If it gets stale for another week it is going to be painful...

dsainati1:
I tried to merge this into the recent changes to Basics, let me know if this looks good to you @bcpierce00. If so it will merge when you accept

dsainati1 (approved):

PR #80, opened 2026-07-13
Formatting and naming improvements to Imp

mwhicks1:
Several across-the-board transformations:
1. Align relation naming and style with Mathlib conventions (e.g., `E_AExpNum` becomes `num` in the `Aexp.EvalR` namespace)
2. Use dot notation for variable receivers consistently (`Aexp.eval st a -> a.eval st` etc.)
3. Wrap the remaining >100-column code, comment and prose lines.
4. Adjust command evaluation relation notation so that commands don't need to be wrapped in `imp { ... }`

PR #82, opened 2026-07-14
added characterizing lemmas; updated proofs accordingly

mwhicks1:
This PR adds characterizing lemmas for the `Bexp.eval` and `Aexp.eval` functions, and updates the proofs to use those lemmas and no longer use `Bexp.eval` and `Aexp.eval` directly. I also did a little tidying. There are still missing exercises from the end, and those proofs need work from an expert (not me!).

Note this PR is stacked on changes I made to address PR #80.

### DIFF
--- a/HL/Imp.lean
+++ b/HL/Imp.lean
@@ -1870,13 +1870,13 @@ inductive Com.EvalR : Com → State → State → Prop where
       (hloop : Com.EvalR (imp {while (~b) {~c}}) st' st'') :
       Com.EvalR (imp {while (~b) {~c}}) st st''
 
-notation:40 st0:41 " =[ " c " ]=> " st1:41 => Com.EvalR c st0 st1
+notation:40 st0 " =[ " c " ]=> " st1 => Com.EvalR c st0 st1
 -- Also accept a bare Imp command between the brackets, so concrete programs can
 -- be written without the `imp { … }` wrapper. Bare `Com` terms still work via the
 -- notation above; splice a Lean term into the command with `~`.
 syntax:40 term:41 " =[ " imp_com " ]=> " term:41 : term
 macro_rules
-  | `($st0 =[ $c:imp_com ]=> $st1) => `($st0 =[ imp { $c } ]=> $st1)
+  | `($st0 =[ $c:imp_com ]=> $st1) => `(Com.EvalR (imp { $c }) $st0 $st1)
 ```
 
 The cost of defining evaluation as a relation instead of a function is
--- a/HL/Imp.lean
+++ b/HL/Imp.lean
@@ -397,13 +397,17 @@ gives us readable names to refer to during proofs.
 
 ```lean
 inductive Aexp.EvalR : Aexp → Nat → Prop where
-  | num (n : Nat) : EvalR (.num n) n
-  | plus (a1 a2 : Aexp) (n1 n2 : Nat) (h1 : EvalR a1 n1) (h2 : EvalR a2 n2) :
-      EvalR (.plus a1 a2) (n1 + n2)
-  | minus (a1 a2 : Aexp) (n1 n2 : Nat) (h1 : EvalR a1 n1) (h2 : EvalR a2 n2) :
-      EvalR (.minus a1 a2) (n1 - n2)
-  | mult (a1 a2 : Aexp) (n1 n2 : Nat) (h1 : EvalR a1 n1) (h2 : EvalR a2 n2) :
-      EvalR (.mult a1 a2) (n1 * n2)
+  | num (n : Nat) :
+      Aexp.EvalR (.num n) n
+  | plus (a1 a2 : Aexp) (n1 n2 : Nat)
+      (h1 : Aexp.EvalR a1 n1) (h2 : Aexp.EvalR a2 n2) :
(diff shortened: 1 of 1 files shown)


HL/Imp.lean
```diff
@@ -205,25 +205,33 @@ implementations and proofs.
...
+::::
+
+```lean
+@[simp] theorem Aexp.eval_num (n : Nat) : (num n).eval = n := rfl
+@[simp] theorem Aexp.eval_plus (a1 a2 : Aexp) : (plus a1 a2).eval = a1.eval + a2.eval := rfl
+@[simp] theorem Aexp.eval_minus (a1 a2 : Aexp) : (minus a1 a2).eval = a1.eval - a2.eval := rfl
+@[simp] theorem Aexp.eval_mult (a1 a2 : Aexp) : (mult a1 a2).eval = a1.eval * a2.eval := rfl
 
-example : Aexp.eval (.plus (.num 2) (.num 2)) = 4 := by rfl
+example : Aexp.eval (.plus (.num 2) (.num 2)) = 4 := by simp
```
xhalo32, line 230: Would it make more sense to use `dsimp` here? This is something we should have a general principle for. I propose the following principle:
- `simp` calls should be ["terminal"](https://leanprover-community.github.io/contribute/style.html#squeezing-simp-calls) (i.e. not in the middle of a proof) or "squeezed" to a `simp only`.
- `dsimp` is okay to use in the middle of a proof
- ~~Use `dsimp` instead of `simp` if possible~~ (this might be too strong of an opinion, would like to hear ideas)

HL/Imp.lean
xhalo32, line 276: I think we should make this irreducible as well. If everyone agrees, I can also do it once I make my pass over the file.

HL/Imp.lean
```diff
@@ -299,35 +320,60 @@ theorem optimize_0plus_sound (a : Aexp) :
...
+lemmas, and uses the induction hypotheses -- which `simp_all` picks up from the
+local context automatically -- to close each goal. The whole proof collapses to
+two lines.
+::::
+
+```lean
+theorem optimize_0plus_sound' (a : Aexp) :
+    a.optimize_0plus.eval = a.eval := by
+  induction a using Aexp.optimize_0plus.induct <;>
+    simp_all [Aexp.optimize_0plus]
```
xhalo32, line 367: As there's no explanation of `simp_all` thus far, will it be explained in this chapter or another one like in the upcoming automation chapter?

HL/Imp.lean
```diff
@@ -2176,7 +2202,7 @@ theorem plus2_spec (st : State) (n : Nat) (st' : State)
   unfold plus2 at heval
   cases heval with
   | asgn _ _ m _ h =>
-      simp only [Aexp.eval] at h
+      simp at h
```
xhalo32: This goes against [mathlib's simp conventions](https://leanprover-community.github.io/contribute/style.html#squeezing-simp-calls), see also my earlier comment
https://github.com/plclub/sf-in-lean/pull/82/changes#r3586872288

HL/Imp.lean
```diff
@@ -2238,7 +2264,7 @@ theorem loop_never_stops (st st' : State) : ¬ (st =[ loop ]=> st') := by
       induction hce with
       | whileFalse b s0 c0 hb =>
           intro heq; unfold loop at heq; injection heq with e1 _
-          subst e1; simp [Bexp.eval] at hb
+          subst e1; simp at hb
```
xhalo32, line 2273: This is okay however, because `simp` is terminal.

xhalo32 (approved):
This is good improvement overall! Discussion about my questions can also happen on zulip if you want, and we can merge this as is.

HL/Imp.lean
dsainati1: `simp_all` will be explained in the Automation chapter, yes, which is going to precede this one at the end of the previous book.

HL/Imp.lean
dsainati1: > Use dsimp instead of simp if possible

I am curious why this would be the case. The understanding I had was that `simp` was going to replace `dsimp` for intents and purposes once we taught them the tactic in the Automation chapter, but it's certainly possible I am misunderstanding the idiomatic style.

xhalo32: Ahh, I seem to have missed that conversation. My thinking is that `dsimp` is more "deterministic" and fits well with e.g. `induction foo with dsimp ...` style proofs. I don't think mathlib has a guideline for when to use `dsimp` but it's certainly used (although much less than `simp`)

HL/Imp.lean
mwhicks1: Why do we want to make it irreducible? I thought that we were only doing that for the first few modules of LF, and from then on were favoring the default.

HL/Imp.lean
mwhicks1: Good catch. Fixed now.

HL/Imp.lean
mwhicks1: Ack. I fixed the other two that were not terminal.

HL/Imp.lean
mwhicks1: I'm going to start a convo in Zulip.

PR #83, opened 2026-07-14
LF/Basics: Pulling apart full and terse content

hgoldstein95:
Previously, LF/Basics was not well split into full vs. terse. This PR aims to fix most (but not yet all) of those issues.

I'll fix the rest of the issues tomorrow (I don't want to hold the lock all day).

### DIFF
--- a/LF/Basics.lean
+++ b/LF/Basics.lean
@@ -319,16 +319,16 @@ code is a whole process; Lean just compiles like any other programming language.
 Following the pattern of the days of the week above, we can
 define the standard type `Bool` of booleans, with members `true`
 and `false`.
-::::
-
-:::terse
-Another familiar enumerated type:
-:::
 
 We define our own `MyBool` to teach the concept of building booleans from
 scratch; later we'll switch to Lean's built-in `Bool`.
 We use a different name to make explicit that this is not the same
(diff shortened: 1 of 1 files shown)


LF/Basics.lean
```diff
@@ -1442,38 +1478,50 @@ theorem add_zero_zero_zero : ∀ n : Nat, n + zero + zero + zero = n := by
...
- that is, that they reduce to the same term. (So, in particular, a
- term is always definitionally equal to itself.)
+The `rfl` tactic closes a goal of the shape `a = a`, for any `a`. It
+checks that both sides of the equality are _definitionally equal_ --
+that is, that they reduce to the same term. (So, in particular, a
+term is always definitionally equal to itself.)
+::::
+
+::::terse
+The `rfl` closes a goal that looks lik `a = a`.
```
```suggestion
The `rfl` closes a goal that looks like `a = a`.
```

I do also wonder if this is _too_ terse; in particular, it misses the part about reduction to the same term.

dsainati1 (approved):

bcpierce00:
Warning: There seem to be some other PRs with changes to Basics!  Someone will need to sort out what is what (in the files and in our workflow processes)...

PR #84, opened 2026-07-14
Polish and build fixes for IndProp, and fix CI

dsainati1:
Fix the build for `IndProp` and add a new exercise relating the `In` predicate to Lean's `Mem`.

This also updates the build process to identify type errors in all files, not just `Basics`.

### DIFF
--- a/LF.lean
+++ b/LF.lean
@@ -3,6 +3,13 @@ import SFLMeta.Ignore
 import SFLMeta.Save
 
 import LF.Basics
+import LF.Induction
+import LF.UsingLean
+import LF.Lists
+import LF.Poly
+import LF.Logic
+import LF.Tactics
+import LF.IndProp
 -- Add `import LF.XXXVerso` here for each generated chapter included below.
 
 import VersoManual
--- a/LF/IndProp.lean
+++ b/LF/IndProp.lean
@@ -531,8 +531,8 @@ inductive Le : Nat → Nat → Prop where
 scoped infix:50 (priority := high) " ≤ " => Le
 
 /- FULL: This definition is a bit simpler and more elegant than the
-   Boolean function `leb` we defined in `Basics`.  As usual, `Le`
-   and `leb` are equivalent, and there is an exercise about that
+   Boolean function `ble` we defined in `Basics`.  As usual, `Le`
+   and `ble` are equivalent, and there is an exercise about that
    later. -/
 
 example : 3 ≤ 5 := by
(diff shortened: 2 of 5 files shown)


LF.lean
```diff
@@ -3,6 +3,13 @@ import SFLMeta.Ignore
 import SFLMeta.Save
 
 import LF.Basics
+import LF.Induction
+import LF.UsingLean
+import LF.Lists
+import LF.Poly
+import LF.Logic
+import LF.Tactics
+import LF.IndProp
```
dsainati1, line 12: We need to keep this up to date with all the chapters to make sure that they all build

LF/IndProp.lean
```diff
@@ -1621,6 +1621,56 @@ theorem ev_plus_plus : ∀ n m p,
...
+
+/-
+  We can also write this definition inductively like so:
+-/
+
+inductive In_Inductive {α : Type} (a : α) : List α → Prop
+  | head (as : List α) : In_Inductive a (a::as)
+  | tail (b : α) {as : List α} : In_Inductive a as → In_Inductive a (b::as)
+
+/- In fact, this is exactly how Lean defines this proposition, which it calls `Mem`, and which
```
dsainati1:
```suggestion
/- In fact, this is exactly how Lean defines this proposition, which it calls `Mem` and which
```

LF/IndProp.lean
```diff
@@ -1621,6 +1621,56 @@ theorem ev_plus_plus : ∀ n m p,
...
+/-
+  We can also write this definition inductively like so:
+-/
+
+inductive In_Inductive {α : Type} (a : α) : List α → Prop
+  | head (as : List α) : In_Inductive a (a::as)
+  | tail (b : α) {as : List α} : In_Inductive a as → In_Inductive a (b::as)
+
+/- In fact, this is exactly how Lean defines this proposition, which it calls `Mem` and which
+   is written `x ∈ l`. A good exercise to test your understanding of induction on
```
dsainati1:
```suggestion
   is written `x ∈ l` (the Unicode symbol is written \in). A good exercise to test your understanding of induction on
```

LF/IndProp.lean
```diff
@@ -2141,54 +2191,54 @@ theorem leb_complete : ∀ (n m : Nat),
     case zero =>
       contradiction
     case succ m' =>
-      rw [succ_leb_succ] at h
+      dsimp [Nat.ble] at h
```
chenson2018, line 2194: Could you explain these few changes to use `dsimp`? It seems different from anywhere else.

dsainati1: The `succ_ble_succ` lemma in Basics that this was originally referencing only works for the special version of `Nat`s that we defined there (i.e., this was a type error before the change), and there is no equivalent lemma that I could find for real `Nat`s. `dsimp`ing the definition (which we have taught them at this point in `UsingLean`) accomplishes the same thing, but I'd happily change this to a rewrite rule if I just missed the appropriate one here

chenson2018: No, this makes sense, thanks for the explanation.

chenson2018 (approved):

mwhicks1 (approved):

PR #85, opened 2026-07-14
Fixes to Induction and UsingLean

bcpierce00:
Split out from #79


bcpierce00:
Not compiling right now -- am investigating why...

dsainati1 (approved):

PR #86, opened 2026-07-15
Initial work porting TS/Types

mwhicks1:
Initial pass at `TS/Types.lean`. This is a draft because it's not close to merging yet. But wanted to get it up rather than keep it on my machine. No need for anyone to comment, but feel free to look.

Worth noting:
- In SF for Rocq, Types follows Smallstep, which follows Imp. There are dependencies to both. For now I've inlined the definitions in Smallstep that are needed (a small handful) but the bigger problem is the text/commentary, which assumes the reader is familiar with concepts not really covered in SFL's current LF. My current thinking is a pared down Smallstep that is at the start of both HL and TS, so that if you go from HL to TS or vice versa you can skip Smallstep the second time through.
- The notation is maybe risky. The basic approach (as used in Imp) doesn't allow for defining notation for a type and then using it in the type's (inductive) definition. Claude figured out a way and that's what's here, but it breaks hygiene selectively. Needs review (and suggestions for alternatives).
- Much of the text has not been touched, neither proofs etc. Just getting the definitions down for now and will come back to this.


TS/Types.lean
```diff
@@ -0,0 +1,1586 @@
...
+:::
+
+```lean
+-- Reserve the step notation, then give its meaning (`Tm.Step`) with
+-- `macro_rules`, so the relation can be *defined* using its own `⟶` notation --
+-- the Lean analogue of Rocq's `Reserved Notation "…" … where "…" := step`.  The
+-- `Tm.Step` in the macro resolves at use-time, i.e. during the inductive's own
+-- elaboration; the `app_unexpander` below prints `Tm.Step a b` back as `a ⟶ b`.
+syntax:40 term:41 " ⟶ " term:41 : term
+macro_rules | `($t ⟶ $t') => `($(Lean.mkIdent `Tm.Step) $t $t')
```
chenson2018: If it is decided that it is essential to have notation while making definitions, I would expect that we use `set_option hygiene false` which is less cryptic. See [here](https://github.com/david-christiansen/sf-port-demo/blob/09bdf09fd12d6849067a8f067c7ce85e97d26d28/port/PLF/STLC.lean#L484) from the port that David had done of this material for a corresponding example.

My first instinct was to avoid this, but maybe it's relatively harmless. Perhaps at worst it's one more case of telling students to avert their eyes while we setup some notation.

xhalo32: I tried the `set_option hygiene false` and found a drawback: the syntax disappears in branches of induction unless we also have the `@[app_unexpander Tm.Step]`. E.g. in the `progress` theorem's `tru` branch the goal would be `<{ true }>.IsValue ∨ ∃ t', <{ true }>.Step t'`.

Compare this with using the more "standard approach", where one uses the name of the inductive type in the constructors and adds the syntax afterwards. This way it doesn't need an unexpander and the goal in the `| tru =>` looks like `<{ true }>.IsValue ∨ ∃ t', <{ true }> ⟶ t'`
```lean
namespace Tm

inductive Step : Tm → Tm → Prop where
  | ifTrue (t1 t2 : Tm) : Step <{ if true then t1 else t2 }> t1
  | ifFalse (t1 t2 : Tm) : Step <{ if false then t1 else t2 }> t2
  | ifStep (c c' t2 t3 : Tm) :
      Step c c' → Step <{ if c then t2 else t3 }> <{ if c' then t2 else t3 }>
  | succStep (t1 t1' : Tm) : Step t1 t1' → Step <{ succ t1 }> <{ succ t1' }>
  | predZero : Step <{ pred 0 }> <{ 0 }>
  | predSucc (v : Tm) : Tm.IsNValue v → Step <{ pred (succ v) }> v
  | predStep (t1 t1' : Tm) : Step t1 t1' → Step <{ pred t1 }> <{ pred t1' }>
  | isZeroZero :Step <{ iszero 0 }> <{ true }>
  | isZeroSucc (v : Tm) : Tm.IsNValue v → Step <{ iszero (succ v) }> <{ false }>
  | isZeroStep (t1 t1' : Tm) : Step t1 t1' → Step <{ iszero t1 }> <{ iszero t1' }>

scoped infix:40 " ⟶ " => Tm.Step

end Tm
-- ...
theorem progress (t : Tm) (T : Ty) (hT : <{ ⊢ t ⦂ T }>) : Tm.IsValue t ∨ ∃ t', t ⟶ t' := by
  solution!
    induction hT with
    | tru => exact .inl (.inl .tru)
-- ...
```

With the unexpander, the goal in `tru` works as expected, and hovering over the `⟶` symbol in the infoview acts identically as with the "standard approach".

TS/Types.lean
```diff
@@ -0,0 +1,1586 @@
...
+
+inductive Tm.Step : Tm → Tm → Prop where
+  | ifTrue (t1 t2 : Tm) : <{ if true then t1 else t2 }> ⟶ t1
+  | ifFalse (t1 t2 : Tm) : <{ if false then t1 else t2 }> ⟶ t2
+  | ifStep (c c' t2 t3 : Tm) :
+      c ⟶ c' → <{ if c then t2 else t3 }> ⟶ <{ if c' then t2 else t3 }>
```
xhalo32: In Imp.lean we opted to use named hypotheses. Should we follow the same convention here?

TS/Types.lean
```diff
@@ -0,0 +1,1586 @@
...
+```
+SOONER: Not exactly sure why, but this seems to be very tricky for
+students.
+```
+:::
+
+```lean
+theorem nvalue_is_nf (t : Tm) (h : Tm.IsNValue t) : Tm.IsNormalForm t := by
+  induction h with
+  | zero => intro hc; obtain ⟨t', hstp⟩ := hc; cases hstp
```
xhalo32, line 394: Here, I think `nomatch hstp` is more idiomatic and signals that there are no cases. I don't think we have introduced `nomatch` in an earlier chapter already. Maybe it could be introduced in IndProp.lean?

TS/Types.lean
```diff
@@ -0,0 +1,1586 @@
...
+inductive Ty where
+  | bool
+  | nat
+
+-- As with `⟶`, reserve the typing notation so the relation can be defined
+-- using it.  The whole judgment is wrapped in `<{ … }>` (like the Rocq
+-- `<{ |-- t \in T }>`): the term is written in the object grammar (bare
+-- variables, no inner `<{ }>`) and the type as `Bool`/`Nat` (a type variable is
+-- spliced, `~T` escapes to a Lean `Ty`).  The `app_unexpander` prints it back.
+syntax:max "<{ " "⊢ " tm " ⦂ " ident " }>" : term
```
xhalo32, line 753: Is the reason to use `⦂` over `:` technical or pedagogical? Lean should handle a regular colon based on a quick experiment.

TS/Types.lean
chenson2018: I use `cases` like this all the time personally...

xhalo32: Okay interesting, I grepped mathlib for `nomatch` and `nofun` and found a very small number of results. So maybe we should skip those entirely and stick with `cases`...

TS/Types.lean
chenson2018: Any time I've tried to do this I've found it very confusing. I in general try to stay away from reusing core notation if at all possible.

TS/Types.lean
xhalo32: In Imp.lean we already have
> We'll write `e ⇓ n` to mean that arithmetic expression `e` evaluates to value `n`.  (We scope the notation to this namespace so it doesn't collide with other evaluation relations later.)  In Lean the notation is declared right after the inductive.

So I propose we stick with this style rather than trying macro trickery to get notation in the constructors.

mwhicks1: Ok, at least for now I'll drop the `macro_rules`/hygiene trickery. I'll use explicit `Tm.Step` … in the constructors, and the notation is attached afterward with a plain `scoped notation:40 t:41 " ⟶ ...` After further discussion we can push through another update.

Vtec234: Hi! You can get both by defining the notation twice - first as `local infix` with `hygiene false` to define the type, then using the standard method with hygiene back on. A `local` notation wrapped in a `section` is only available within that section, so should not conflict with the real notation.
```lean
section

set_option hygiene false in
local infix:40 " ~> " => Step

inductive Step : Nat → Nat → Prop where
  | mk (n m : Nat) : n ~> m

end

infix:40 " ~> " => Step

example : 0 ~> 1 := by
  sorry -- ⊢ 0 ~> 1
```

berberman: Thank you for the suggestion @Vtec234. I would also do this.

(And if we still prefer the macro rules, at least we should use `mkCIdent` instead of `mkIdent`.)

mwhicks1: I've just pushed an update that makes this change for `Tm.Step` and `Tm.AltStep`. Much nicer! @Vtec234 do you think we can do something similarly simple for the typing judgments `Tm.HasType` starting around line 734?

bcpierce00 (commented):
This is going to need a couple of line-by-line proofreading passes once it stabilizes, but from a quick pass it seems good so far.  I left some comments inline.

bcpierce00 (commented):
Oops -- I seem to have marked this as "Ready for review," but I don't know how to unmark it!

KihongHeo:
Silly question: how to type `⦂`? Hovering over the symbol does not give any message. `\typecolon` does not work.

chenson2018:
> Silly question: how to type `⦂`? Hovering over the symbol does not give any message. `\typecolon` does not work.

Not a silly question, as there is not an abbreviation. It should be added to settings `.vscode/settings.json`. 

(Note that `\:` is already an abbreviation for `∶`, another Unicode colon. I didn't suggest it because it is probably confusing in being very hard to tell apart from the usual colon.)

TS/Types.lean
mwhicks1: Followup from @Vtec234 on a separate channel:

> Over there you could similarly use unhygienic `local macro_rules` in the judgment definition and a hygienic `macro_rules` afterwards.
>
> But I don't think you can easily get rid of the unexpanders. This is because you are using `macro_rules` for another reason besides hygiene, namely to implement a somewhat complex syntax for the typing judgment.
>
> Lean has a hierarchy of syntax-defining operations, ranging from simple-to-use but less expressive ones, to more expressive but lower-lever ones. notation automatically generates unexpanders for you but (IIRC) does not support custom syntactic categories. You are using one - the `tm` category - so `macro_rules` which does not generate unexpanders must be used.

The most recent commit now uses this scoped hygiene trick but otherwise uses macro rules etc.

PR #90, opened 2026-07-16
Fix explanation for proof by rewrite

KihongHeo:
- Fixed incorrect explanation
- More discussion on "proof by rw" is needed. Let me add comments below.

### DIFF
--- a/LF/Lists.lean
+++ b/LF/Lists.lean
@@ -557,14 +557,12 @@ abbrev add (v : Nat) (s : Bag) : Bag :=
 -- test_add1
 example : count 1 (add 1 [1, 4, 1]) = 3 := by
   -- ADMITTED
-  dsimp [add]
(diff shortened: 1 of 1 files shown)


dsainati1 (approved):

LF/Lists.lean
KihongHeo, line 558: In this file, we have two different styles of proofs. 
Here we use `rw`-based ones, 
```lean
example : count 1 (add 1 [1, 4, 1]) = 3 := by
  dsimp [add]
  rw [count_cons_same, count_cons_same, count_cons_diff, count_cons_same, count_nil] <;> dsimp
```

but other parts use `unseal` + `rfl` styles like
```lean
unseal count
unseal remove_one
example : count 5 (remove_one 5 [2, 1, 5, 4, 1]) = 0 := by rfl
```
 I think it would be better to make it consistent?

KihongHeo: I think the former better aligns with our educational goals. However, I’m not sure whether writing it in such a verbose way is idiomatic Lean?

PR #99, opened 2026-07-18
Split out Slang from Imp

mwhicks1:
Split out the HL/Imp chapter's opening (~810 lines: variable-free Aexp/Bexp, evaluation,
optimization, and evaluation-as-a-relation) a new TS/Slang chapter.
    
- Delete the Arithmetic and Boolean Expressions / Optimizing Booleans / Evaluation as a Relation sections (Warmup, AevalRDivision, AevalRExtended).
- No import of TS.Slang: nothing here uses it in code, and importing would make HL depend on a TS-volume module, against the independent-paths goal. I like this better than (a) moving Slang into the end of LF, and (b) duplicating it at the start of both TS and HL.
- Reference Slang in prose (plain text; a live cross-book link isn't supported across separate books yet).
- Redirect the two Warmup back-references to Slang.


TS/Slang.lean
```diff
@@ -0,0 +1,839 @@
...
+      | a − a
+      | a × a
+
+  b := bool
+      | a = a
+      | a ≠ a
+      | a ≤ a
+      | a > a
+      | ¬ b
+      | b ∧ b
```
xhalo32, line 122: These don't match what we currently have in Imp.lean. Do we want to switch to this syntax or stick with what we have? I like the unicode except for `×`, which I prefer remains `*`.

xhalo32 (approved):
This looks good, although I didn't review every line. As text was moved from file to file and new modifications were made it made this pretty difficult to review. Having those in separate commits would have made it easier.

TS/Slang.lean
mwhicks1: I'll switch it to `*` in this BNF in this comment.

Can elaborate on what you mean by "These don't match what we currently have"? Maybe
- There is no informal syntax (BNF) at all in `Imp.lean`
- The notation on the AST in `Imp.lean` for `Aexp` and `Bexp` do not use Unicode. (`Slang.lean` does not use notation on the syntax.)

I assume you mean the latter. If so, I agree that it makes sense to change it.

Do you think we should introduce notation in `Slang.lean` too? (I'm going to say 'no' and say we leave it as is, so it's simpler.)

TS/Slang.lean
```diff
@@ -0,0 +1,839 @@
...
+lemmas, and uses the induction hypotheses -- which `simp_all` picks up from the
+local context automatically -- to close each goal. The whole proof collapses to
+two lines.
+::::
+
+```lean
+theorem optimize0plus_sound' (a : Aexp) :
+    a.optimize0plus.eval = a.eval := by
+  induction a using Aexp.optimize0plus.induct <;>
+    simp_all [Aexp.optimize0plus]
```
berberman: Ah, I wasn't aware we are using functional induction. IMO we should prefer `fun_induction` over the generated `.induct` definition, e.g.

```lean
fun_induction Aexp.optimize_0plus <;> simp_all
```

Generated names are not supposed to be used manually. I would introduce this tactic first.

mwhicks1: TBH, I don't know if we are doing this or not. This part will need to change once the Automation chapter is done. @dsainati1 what do you think?

dsainati1: Hmm the automaton chapter doesn’t talk about functional induction at all. If we want students to use this we will need to dedicate some space to teaching it elsewhere probably?

mwhicks1: I've made this change to the repo, and left a comment about where `fun_induction` should be introduced.

TS/Slang.lean
xhalo32: I agree that we should not introduce notation in Slang. The updated syntax in Imp.lean now looks good to me and resolves my comment!

PR #100, opened 2026-07-18
First draft of the Automation chapter

dsainati1:
This is a first draft of the Automation chapter, which contains explanations of `lia`, `trivial`, `try`, `repeat`, `first`, and most importantly `simp`. After explaining these, the chapter culminates with the `RegExp` exercise, moved here from `IndProp`.


LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+
+  - the logical connectives `∧`, `∨`, `¬`, and `→`,
+
+then invoking `lia` will either solve the goal or fail, meaning
+that the goal is actually false.  (If the goal is _not_ of this
+form, `lia` will fail.)
+::::
+
+```lean
+example : ∀ (m n o p : Nat),
```
chenson2018: At this late in the book, are we still writing hypotheses to the right of the colon? I thought it was decided that after the first couple of chapters we'd prefer the follow the style guide.

LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+%%%
+htmlSplit := .never
+file := "Automation"
+%%%
+
+::::full
+Up to now, we've used the manual part of Lean's tactic
+facilities.  In this chapter, we'll learn more about some of
+Lean's powerful automation features, including
+tactic combinators like `try` and `repeat`, decision procedures like `lia`,
```
chenson2018: Remember that when introducing a new tactic, we can write `{tactic}` before it so that it links the docs in the HTML. (Maybe you prefer below when each is properly introduced)

LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+`lia` can many of the cases of our old `Perm3_In` example.
+```lean
+theorem Perm3_In_better_with_lia : ∀ (α : Type) (x : α) (l₁ l₂ : List α),
+    Perm3 l₁ l₂ → x ∈ l₁ → x ∈ l₂ := by
+  intros α x l₁ l₂ hPerm hIn
+  induction hPerm
+  case perm3_swap12 a b c =>
+    rw [List.mem_cons, List.mem_cons, List.mem_cons] at *
+    obtain h | h | h | h := hIn
+    . lia /- was right; left; assumption -/
```
chenson2018, line 118: I find these and the last example above a pretty strange in that these aren't about arithmetic. I think most would write `simp [h]` here. (But maybe this is intentional and I need to read further...)

Edit: no, I got to the end and still find this odd!

LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+If the goal is a universally quantified formula made out of
+
+  - numeric constants, addition (`+` and `succ`), subtraction (`-` and `pred`)
+    and multiplication by constants (this is what makes it Presburger arithmetic),
+
+  - equality (`=` and `≠`) and ordering (`≤` and `<`), and
+
+  - the logical connectives `∧`, `∨`, `¬`, and `→`,
+
+then invoking `lia` will either solve the goal or fail, meaning
```
chenson2018, line 79: When `lia` fails, it will mention `grind`. So maybe at some place should can be a short explanation that it's a superset of the `lia` rules we're not going to discuss. (Or if you really want to hide this, it wouldn't be too hard to do so in a macro I think)

LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+
+::::terse
+The lemmas we've been using for rewriting are
+the same ones we'll give to `simp` for it to automatically
+solve goals involving those theorems.
+::::
+
+```lean
+namespace simp_lemmas_example
+def add (n : Nat) (m : Nat) : Nat :=
```
chenson2018: Why are we not using notation here? I'd expect to never be writing `succ` this far in. Is the point that you're try to exclude the core `simp` lemmas from being trigger for teaching purposes?

I think it's especially important here because part of the consideration for a simp normal form is notation. To be explicit, we have theorems like 
```lean
@[simp] theorem Nat.add_eq : Nat.add x y = x + y := rfl
```
in core precisely so that we flow from non-notation to notation.

LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+  rfl
+
+/- `add_zero` and `add_succ` are the `simp` lemmas for `add`. -/
+theorem add_succ : ∀ n m : Nat, add n (.succ m) = .succ (add n m) := by
+  intro n m
+  rfl
+end simp_lemmas_example
+```
+
+When we tag theorems with `@[simp]`, like so:
```
chenson2018: I'd swap the order here and use `@[simp]` on the above since it is much more common, then later explain you can add attributes afterwards. (Though they've seen it happen before previously it probably didn't stick since it wasn't emphasized)

LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+
+::::terse
+`simp only` uses only the provided theorems:
+::::
+```lean
+open simp_lemmas_example in
+theorem add_succ_nested_2 : ∀ n m : Nat,
+    add n (.succ (.succ m)) = .succ (.succ (add n m)) := by
+  intro n m
+  simp only [Nat.succ_eq_add_one, add_succ]
```
chenson2018: As a further effect of not using notation, `Nat.succ_eq_add_one` appearing is probably confusing to a beginner.

LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+`simp` makes our final example _much_ shorter.
+::::
+```lean
+theorem Perm3_In_shortest (α : Type) (x : α) (l₁ l₂ : List α) :
+    Perm3 l₁ l₂ → x ∈ l₁ → x ∈ l₂ := by
+  intros hPerm hIn
+  induction hPerm <;>
+  first
+  | simp at * <;> lia
+  | lia
```
chenson2018, line 493: I find this a little rough, I'd hope for something more like

```suggestion
  induction hPerm <;> simp_all <;> lia
```
Was it decided maybe that `simp_all` is a bit awkward to introduce?

Edit: I see `simp_all` is introduced below. Then maybe this could use `simp [*]` and get something still nicer?

LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+```
+
+::::full
+Like `apply` and `rw`, there's also a version of `simp` that can simplify in
+hypotheses, rather than the goal. The `simp [<lemmas>] at h` tactic
+runs the simplifier with at hypothesis `h`.
```
chenson2018, line 499: The phrasing here suggests that this is a different tactic, which is not true in the case of `simp`

LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+attribute [pp_nodot] RegExp.Char RegExp.App RegExp.Union RegExp.Star
+```
+
+Note that this definition is _polymorphic_: Regular
+expressions in `RegExp α` describe strings with characters drawn
+from `α` -- which in this exercise we represent as _lists_ with
+elements from `α`.
+
+::::full
+(Technical aside: We depart slightly from standard practice in
```
chenson2018, line 702: Do you mean here that this is different because the inductive type doesn't specify `α` is finite? In Lean the convention is for inductives not to carry `Prop`-valued typeclasss assumptions, enforcing this only at the theorems that use them. So this could give off a slightly wrong impression.

LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+                        `s₂ =~ (Star re)`
+                    ----------------------- (MStarApp)
+                    `(s₁ ++ s₂) =~ (Star re)`
+
+
+This directly corresponds to the following `inductive` definition:
+
+```lean
+namespace RegExp
+inductive ExpMatch {α : Type} : List α → RegExp α → Prop where
```
chenson2018, line 773: The constructors here should be updated to match the naming conventions.

LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+::::exercise (rating := 1) (name := "regexp_match_of_list")
+As a quick exercise, prove that every list matches `reg_exp_of_list` of itself:
+
+```lean
+theorem regexp_match_of_list α (l : List α) : l =~ reg_exp_of_list l := by
+  solution!
+    induction l with
+    | nil => constructor
+    | cons hd tl ih =>
+        simp [reg_exp_of_list]
```
chenson2018: The usage of `simp` here does not follow the convention for non-terminal `simp`, see [this section](https://leanprover-community.github.io/contribute/style.html#squeezing-simp-calls) of the style guide. (If it doesn't make sense from the relatively short explanation, I'm happy to explain over a call)

This is a core principle of correctly using `simp` that should be explained. Mathlib also has a very nice linter that even gives explicit line numbers of what tactics are breaking this rule. If we don't add a Mathlib dependency (still my preference), we should at least vendor it.

LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+
+
+:::dev "Daniel Sainati (@dsainati1)"
+How to make this a WORKINCLASS in verso?
+:::
+```lean
+theorem MStar1 α s (re : RegExp α) :
+    s =~ re →
+    s =~ Star re := by
+    intro h
```
chenson2018, line 920: I don't want to leave a bunch of nitpicky style comments, but I'll just mention once as an example that this should be

```suggestion
theorem MStar1 {α : Type} (s : List α) (re : RegExp α) : s =~ re → s =~ Star re := by
  intro h
```
or

```suggestion
theorem MStar1 {α : Type} (s : List α) (re : RegExp α) :
    s =~ re → s =~ Star re := by
  intro h
```
if the conversation about narrower columns is adapted. The rules here (all in the style guide) that I applied are:

- always providing a signature for hypotheses, even when they can be inferred
- tactics beginning at 2 spaces indent
- if needed, long theorem statements before tactics aligning at 4 spaces

LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+    rw [←List.append_nil s]
```
chenson2018:
```suggestion
    rw [← List.append_nil s]
```

LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+  ss.foldr (· ++ ·) [] =~ RegExp.Star re := by
+  -- ADMITTED
+  induction ss with
+  | nil => constructor
+  | cons s ss' ih =>
+    simp
+    constructor
+    · apply h; simp
+    · apply ih; intro s' hs'
+      apply h; simp; right; assumption
```
chenson2018: I don't think I'll flag them all, but this is another case of non-terminal simp usage

LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+proving the pumping lemma below.
+
+```lean
+theorem pumping_constant_ge_1 {α : Type} (re : RegExp α) :
+    pumpingConstant re ≥ 1 := by
+  induction re with
+  | EmptySet => simp [pumpingConstant]
+  | EmptyStr => simp [pumpingConstant]
+  | Char _ => simp [pumpingConstant]
+  | App re₁ _ ih1 _ => simp [pumpingConstant]; lia
```
chenson2018, line 1298: Since I've been commenting on non-terminal `simp`, note that `simp; lia` *is* allowed: because `lia` is a tactic that doesn't rely on the specific structure of rewrites that `simp` does, we consider it flexible and allow them to follow each other without `simp only`.

LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+
+These auxiliary lemmas might also be useful in your proof of the
+pumping lemma.
+
+```lean
+theorem napp_plus {α : Type} (n m : Nat) (l : List α) :
+    napp (n + m) l = napp n l ++ napp m l := by
+  induction n with
+  | zero => simp [napp]
+  | succ n ih => simp [Nat.succ_add, napp, ih]
```
chenson2018: The way `napp` and `Nat.succ_add` are used here is not great, these should be replaced by appropriate `simp` lemmas. 

See the proof of [List.replicate_append_replicate](https://leanprover-community.github.io/mathlib4_docs/Init/Data/List/Lemmas.html#List.replicate_append_replicate) in core for comparison.

LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+    pumpingConstant re₁ <= s₁.length ∨ pumpingConstant re₂ <= s₂.length := by
+    solution!
+      rw [app_length] at Hlen
+      apply add_le_cases
+      apply Hlen
+  . solution!
+      specialize ih₁ H
+      let ⟨s₁₂, s₁₃, s₁₄, h₁, h₂, h₃⟩ := ih₁
+      rw [h₁]
+      exists s₁₂; exists s₁₃; exists (s₁₄ ++ s₂)
```
chenson2018: I'm only skimming this proof, but this stood out

```suggestion
      exists s₁₂, s₁₃, s₁₄ ++ s₂
```

LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+      . intro m; specialize h₃ m
+        rw [←List.append_assoc]
+        constructor <;> trivial
+  . solution!
+      specialize ih₂ H
+      let ⟨s₂₁, s₂₂, s₂₃, h₁, h₂, h₃⟩ := ih₂
+      rw [h₁]
+      exists (s₁ ++ s₂₁); exists s₂₂; exists s₂₃
+      constructor
+      . rw [←List.append_assoc, ←List.append_assoc]
```
chenson2018: The repeated use of `append_assoc` (a simp lemma) in the opposite direction is suspicious, maybe some definition is not stated in simp nf?

LF/Automation.lean
```diff
@@ -0,0 +1,1535 @@
...
+        rw [←List.append_assoc]
+        assumption
+```
+::::
+
+::::exercise (rating := 3) (name := "weak_pumping_union_l")
+```lean
+theorem weak_pumping_union_l :  ∀ {α : Type} (s₁ : List α) (re₁ re₂ : RegExp α),
+  s₁ =~ re₁ →
+  (pumpingConstant re₁ <= s₁.length →
```
chenson2018:
```suggestion
  (pumpingConstant re₁ ≤ s₁.length →
```

chenson2018 (commented):
I only skimmed the long proofs at the end, as I'd been reviewing for about an hour and was quickly losing steam. Mostly small things, except for the bit about non-terminal `simp` usage.

LF/Automation.lean
dsainati1: @rogerburtonpatel was there a reason this was structured this way?

LF/Automation.lean
dsainati1: I think the reason for this is that `lia` does solve first order logic goals as well as arithmetic ones. Is there a more limited tactic for just first order goals that we should use here instead?

LF/Automation.lean
dsainati1: Yeah, I think I am not fully following the convention. Perhaps we should discuss. Or, if you are interested, you could add a paragraph to the `simp` section earlier in the file to explain the convention? We will need to explain it to the students at any rate.

LF/Automation.lean
dsainati1: This should be taken care of by the autoformatter eventually, right?

LF/Automation.lean
dsainati1: I am also not sure I follow the suggestion here

chenson2018: I mean that `Nat.succ_eq_add_one` is a theorem about notation that would ideally not appear at all here. I think it only did because of earlier not using notation for addition.

LF/Automation.lean
dsainati1: @bcpierce00 What was the purpose of this aside in the original Rocq text? Does it make sense to keep here?

LF/Automation.lean
```diff
@@ -0,0 +1,1631 @@
...
+The simplest version of our theorem uses `simp_all`:
+
+```lean
+theorem Perm3_In_shortest (α : Type) (x : α) (l₁ l₂ : List α) :
+    Perm3 l₁ l₂ → x ∈ l₁ → x ∈ l₂ := by
+  intros hPerm hIn
+  induction hPerm <;> simp_all <;> lia
+```
+
+## Idiomatic `simp` Usage
```
dsainati1, line 545: @chenson2018 let me know how this section reads to you

LF/Automation.lean
```diff
@@ -0,0 +1,1631 @@
...
+
+
+::::exercise (rating := 1) (name := "re_not_empty")
+Write a recursive function `re_not_empty` that tests whether a
+regular expression matches some string. Prove that your function
+is correct.
+
+:::solution
+```lean
+def re_not_empty {α : Type} (re : RegExp α) : Bool :=
```
dsainati1: Just to check my understanding: we should _not_ put a `simp` tag on this function? Is it better to write `simp [re_not_empty]` like in the proof below?

LF/Automation.lean
```diff
@@ -0,0 +1,1631 @@
...
+::::full
+There is not much reason to use `try` in completely manual proofs like
+these, but it is very useful together with the `<;>` combinator.
+::::
+
+```lean
+inductive silly : Nat → Prop where
+| silly1 n (h : n > 1) : silly n
+| silly2 n (h : 1 ∈ []) : silly n
+| silly3 n (h : exists m, n = m + 2) : silly n
```
berberman: more natural naming imo

```lean
inductive Silly : Nat → Prop where
| mk1 n (h : n > 1) : Silly n
| mk2 n (h : 1 ∈ []) : Silly n
| mk3 n (h : ∃ m, n = m + 2) : Silly n
```

LF/Automation.lean
```diff
@@ -0,0 +1,1631 @@
...
+::::terse
+Consider the proof below. Notice all the repetition and near-repetition...
+::::
+
+```lean
+theorem Perm3_In_old (α : Type) (x : α) (l₁ l₂ : List α) :
+    Perm3 l₁ l₂ → x ∈ l₁ → x ∈ l₂ := by
+  intros hPerm hIn
+  induction hPerm
+  case perm3_swap12 a b c =>
```
berberman: Let's use `induction ... with` and `| ...` for branches.

LF/Automation.lean
```diff
@@ -0,0 +1,1631 @@
...
+
+```lean
+theorem Perm3_In_better_with_try (α : Type) (x : α) (l₁ l₂ : List α) :
+    Perm3 l₁ l₂ → x ∈ l₁ → x ∈ l₂ := by
+  intros hPerm hIn
+  -- TODO: autoformatter needs to make this look decent
+  induction hPerm <;>
+  try rw [List.mem_cons, List.mem_cons, List.mem_cons] at * <;> lia
+  case perm3_trans _ _ _ _ _ ih₁2 ih₂3 =>
+    lia
```
berberman: how about:
```suggestion
theorem Perm3_In_better_with_try (α : Type) (x : α) (l₁ l₂ : List α) :
    Perm3 l₁ l₂ → x ∈ l₁ → x ∈ l₂ := by
  intro hPerm hIn
  induction hPerm with (try rw [List.mem_cons, List.mem_cons, List.mem_cons] at * <;> lia)
  | perm3_trans => lia
```

LF/Automation.lean
```diff
@@ -0,0 +1,1631 @@
...
+:::
+
+```lean
+theorem Perm3_In_better_with_first (α : Type) (x : α) (l₁ l₂ : List α) :
+    Perm3 l₁ l₂ → x ∈ l₁ → x ∈ l₂ := by
+  intros hPerm hIn
+  induction hPerm <;>
+  first
+  | rw [List.mem_cons, List.mem_cons, List.mem_cons] at * <;> lia
+  | lia
```
berberman: I wonder if indenting the body tactics would make it nicer:
```lean
induction hPerm <;>
  first
  | rw [List.mem_cons, List.mem_cons, List.mem_cons] at * <;> lia
  | lia
```

LF/Automation.lean
```diff
@@ -0,0 +1,1631 @@
...
+::::
+
+::::terse
+Consider the proof below. Notice all the repetition and near-repetition...
+::::
+
+```lean
+theorem Perm3_In_old (α : Type) (x : α) (l₁ l₂ : List α) :
+    Perm3 l₁ l₂ → x ∈ l₁ → x ∈ l₂ := by
+  intros hPerm hIn
```
berberman: Not specific to this PR, we should prefer `intro` over `intros` in general. IIRC `intros ...` is a legacy tactic equivalent to `intro ...` and it might get removed at some point. The only difference is that bare `intros` introduces many names while `intro` only introduces one, but IMO unnamed `intro`/`intros` should be avoided.

LF/Automation.lean
```diff
@@ -0,0 +1,1631 @@
...
+    rw [List.mem_cons, List.mem_cons, List.mem_cons] at *
+    obtain h | h | h | h := hIn
+    . lia /- was right; left; assumption -/
+    . lia
+    . lia
+    . lia
+  case perm3_swap23 a b c =>
+  -- Here, we solve _all_ goals- and eschew the `obtain` - with
+  -- the <;> tactic combinator, which we saw in the `Induction` chapter.
+    rw [List.mem_cons, List.mem_cons, List.mem_cons] at * <;> lia
```
berberman: Maybe we could introduce `all_goals` as well?

LF/Automation.lean
```diff
@@ -0,0 +1,1631 @@
...
+  -- the * here targets all hypotheses and the goal
+  simp only [List.mem_append] at *
+  cases h₁
+  . left; left; assumption
+  . right; assumption
+```
+
+:::dev "Daniel Sainati (@dsainati1)"
+Chris suggested using Mathlib's `linter.flexible` option to enforce proper `simp` usage.
+How do we feel about adding a Mathlib dependency for this?
```
berberman, line 627: Personally I usually disable this... I understand why it exists, but I'm not sure it's something we should emphasize right now, since IMO it's more about writing maintainable proofs instead of a basic concept.

LF/Automation.lean
```diff
@@ -0,0 +1,1631 @@
...
+    s₁ = s₀ ++ s₂ ++ s₃ ∧
+    s₂ ≠ [ ] ∧
+    (∀ m : Nat, s₀ ++ napp m s₂ ++ s₃ =~ Union re₁ re₂) := by
+  intro Hmatch IH Hlen
+  have H : pumpingConstant re₁ ≤ s₁.length := by
+    solution!
+      simp [pumpingConstant] at Hlen
+      lia
+  solution!
+    specialize IH H
```
berberman: My impression is that hypotheses should be in lowercase

LF/Automation.lean
```diff
@@ -0,0 +1,1631 @@
...
+    let ⟨s₁₁, s₁₂, s₁₃, h₁, h₂, h₃⟩ := IH
```
berberman: Do we distinguish `let`/`have`/`obtain`?

berberman (commented):
Some nitpicking comments about the code style (some of them apply to previous chapters as well.)

LF/Automation.lean
dsainati1: We have at this point taught all of these tactics, I think. Is there a particular one we should be using here instead of `let`?

LF/Automation.lean
dsainati1: We could. Is use of `all_goals` idiomatic in Lean? In Rocq my sense is that it's an antipattern to structure proofs this way, but that might be specific to Rocq.

LF/Automation.lean
chenson2018: For destructing a proposition, I'd lean towards `have`/`obtain` interchangeably. `let` has different unfolding behavior that I associate with more often being used with data.

LF/Automation.lean
chenson2018: I'd consider it fine to use here, you could probably make `<;>` work too.

LF/Automation.lean
chenson2018: This seems fine. I'd maybe give an example of doing the wrong thing with `simp; exact` and explain why that is bad (it relies on the exact hypothesis structure).

LF/Automation.lean
berberman: Yes, I think so. Marking as `@[simp]` means that we *always* want to unfold the function during `simp`. This is fine as long as the unfolded RHS does not introduce new recursive calls on larger terms or something that may lead to infinite loops, or we don't want to preserve the notations (which is rare). 

So here it wouldn't be harmful to have this function tagged, though, in my opinion, the good style, or the style we currently follow is to avoid simp on all definitions and rather write simp lemmas or unfold the definitions manually like what we do right now.

To reduce some boilerplates, we could do `induction re with (simp [re_not_empty])`. BTW I think this function should be written in lower camel case, e.g. `reNotEmpty` according to the conventions. IIUC theorem/proof uses snake case, type/proposition uses upper camel case, and other function/value uses lower camel case.

LF/Automation.lean
KihongHeo: I had a similar impression while reading it. I think it would be helpful for the reader if, when we explain the conditions under which lia applies above, we mention that it also uses “simple” logical facts in addition to arithmetic, and then explicitly remind the reader of that here as well.

LF/Automation.lean
```diff
@@ -0,0 +1,1644 @@
...
+::::terse
+Consider the proof below. Notice all the repetition and near-repetition...
+::::
+
+```lean
+theorem Perm3_In_old (α : Type) (x : α) (l₁ l₂ : List α) :
+    Perm3 l₁ l₂ → x ∈ l₁ → x ∈ l₂ := by
+  intro hPerm hIn
+  induction hPerm with
+  | perm3_swap12 a b c =>
```
KihongHeo: Do we have a style guideline for when to write parameter names explicitly such as a, b, and c, when to use _, and when to omit them altogether? It seems that this file currently uses a mix of all three styles. For example, at line 234, `| perm3_trans => lia`

LF/Automation.lean
```diff
@@ -0,0 +1,1644 @@
...
+As a culminating exercise for this book and as practice using the automation techniques we
+discussed above on a real proof,
+we examine the theory of regular expressions,
+eventually working up to a proof of the pumping lemma.
+::::
+
+## Definitions
+
+::::full
+Regular expressions are a natural language for describing sets of
```
KihongHeo: "natural language" sounds a bit odd here. Some readers might think "NO. RE is a formal language!"

LF/Automation.lean
xhalo32: I agree that we shouldn't worry about teaching proof maintainability across lean and mathlib versions.

chenson2018: I'm not advocating for an extended discussion of proof maintainability, but it seems like a simple enough good habit to instill (use `simp only` when in the middle of a proof).

mwhicks1: I am not following this discussion. Is it about using the linter or using the `simp only` pattern? I think we should be using the latter (and explaining why), but I am ambivalent about using the former.

I believe that "`simp only` in the middle" is a Mathlib "best practice" as @xhalo32 previously pointed out to me, so I have used in `Imp` and other porting work, since Mathlib best practices are linked from CONTRIBUTING.md. Our third tenet is that we are teaching proof engineering, and thus should follow idiomatic Lean, which we have interpreted to mean idiomatic Mathlib, too.

LF/Automation.lean
mwhicks1: I don't see anything in the Mathlib [naming](https://leanprover-community.github.io/contribute/naming.html) or [style](https://leanprover-community.github.io/contribute/style.html) guides. Do you have an opinion (does anyone)? In code, Lean will warn you if you name parameters you do not use. So it seems reasonable to me to make these `_`, to match that style.

Meta: The way I consider such questions is by checking:
1. Is there a Mathlib/Lean convention that recommends a particular approach?
2. If not, do we have reason to have an opinion? One way to support an opinion is by explicated principle; another way is data mining from Mathlib proofs (say). If we have no strong opinion, we don't need to be consistent. If we do have an opinion, it should be in [CONTRIBUTING.md](https://github.com/plclub/sf-in-lean/blob/main/CONTRIBUTING.md), which also links the Mathlib stuff.

LF/Automation.lean
```diff
@@ -576,6 +558,100 @@ theorem Perm3_In_shortest (α : Type) (x : α) (l₁ l₂ : List α) :
...
+theorems in the simp set is the _simp normal form_ of that expression.
+
+It's important for the stability of proofs using `simp` that all the theorems in the simp set
+progress towards this normal form. Accordingly, library designers often first consider
+what they want that normal form to look like, and then structure their theorem definitions
+accordingly. As a simple example, the simp normal form for lists prefers to use the `++`
+notation instead of `List.append`, so there is a `simp` theorem `List.append_eq`
+whose type is `List.append_eq {α : Type u} {as bs : List α} : as.append bs = as ++ bs`.
+
+The `simp` normal form appears on the right, while the expression in need of simplification
```
xhalo32: The right-hand-side doesn't necessarily always have to be in simp-normal form. [Here's an example with the lean prelude](https://live.lean-lang.org/#codez=LTAEBUAsFNQJwJYHNIBdiQIYDsAmwBnBXWAewDNQADAIVNIBsA6aARwH1tTV2DoHy7HLiqgEBUF1SgAcgDFQAI2gBjTAFc%2B1OoyZT2bAx0NTRa7EtiYADtYYJouMQFtnjhJlT8AnkwBQAMQqMCoA1qA6zCbcvPyCwoHBquGRejGGGRxSftAAHpjOdrAA3phKoABcEfQMAL6VoAAUZQC8TQCEZQBkXUoAlH2VbYrefqCgRIWgpNgM3qAA2qnRPHwCQngANNW6%2BplGnNzbqfqocOrQALpjE6RwcN5AA).

chenson2018: Yes, it would be more accurate to say the right hand side should make *confluent progress towards* a normal form. (It's worth noting that nothing enforces this programmatically, but core is generally pretty good about it)

bcpierce00:
I'm sure there will be a lot more to polish on this chapter, but let's go ahead and merge it (after dealing with any pending in-PR comments) for now.

bcpierce00 (approved):

PR #104, opened 2026-07-20
clean up Lists

KihongHeo:
- Fix minor things
- More discussion needed (see my comments below)

### DIFF
--- a/LF/Lists.lean
+++ b/LF/Lists.lean
@@ -68,6 +68,10 @@ Why is this namespace needed??
 So that the definitions here don't clash with the standard library.
 :::
 
+:::dev "Benjamin Pierce (bcpierce00)"
+Then let's explain it, if really needed.
+:::
+
 ```lean
 namespace Lists
 ```
--- a/LF/Lists.lean
+++ b/LF/Lists.lean
@@ -60,17 +60,11 @@ import LF.Induction
 import LF.UsingLean
 ```
 
-:::dev "Benjamin Pierce (bcpierce00)"
-Why is this namespace needed??
-:::
-
-:::dev "Daniel Sainati (dsainati1)"
-So that the definitions here don't clash with the standard library.
-:::
-
-:::dev "Benjamin Pierce (bcpierce00)"
-Then let's explain it, if really needed.
-:::
+::::full
+This chapter introduces basic data structures and functions for working with
+them. We place all these definitions in the `Lists` namespace to avoid name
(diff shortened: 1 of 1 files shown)


LF/Lists.lean
KihongHeo: I don't see any conflicting definitions with the standard library once I remove this namespace. Am I missing something? @dsainati1

LF/Lists.lean
```diff
@@ -183,7 +183,7 @@ The distinction is minor, but it is worth understanding that they
 are not the same. For instance, the following definitions are
 ill-formed:
 
-```display
+```lean +error
 -- Can't match on a pair with multiple patterns:
```
KihongHeo, line 185: In the HTML document, this sentence appears twice.
```
-- Can't match on a pair with multiple patterns:
-- Can't match on a pair with multiple patterns:
```

LF/Lists.lean
KihongHeo, line 900: In the html document,

```
-- FILL IN HERE
-- FILL IN HERE
```

The above sentence appears twice.

mwhicks1 (approved):

LF/Lists.lean
dsainati1: Removing this causes the build to break for me, because the syntax we define here makes uses of list notation in e.g. LF.lean to parse as a NatList. If you have a better way around that though I’m for it.

LF/Lists.lean
bcpierce00: Looks OK now

LF/Lists.lean
```diff
@@ -754,7 +754,7 @@ theorem member_add_same v t : member v (add v t) = true := by
     rw [BEq.refl]
     dsimp
 
-theorem member_add_diff v1 v2 t : (v1 == v2) = false -> member v1 (add v2 t) = member v1 t := by
+theorem member_add_diff v1 v2 t : (v1 == v2) = false → member v1 (add v2 t) = member v1 t := by
```
bcpierce00, line 755: This seems to be changing just a single -> to →.  If we're going to do this, we should do it globally.

bcpierce00:
This looks good to go, except for my comment about the stray →.  Please go ahead and merge once this is fixed.

bcpierce00 (approved):

LF/Lists.lean
KihongHeo: I see. Thanks.

LF/Lists.lean
KihongHeo: Fixed.

PR #105, opened 2026-07-20
Salvage the still-relevant parts of bcp-misc

bcpierce00:
A few miscellaneous improvements (that I forgot to merge earlier):

- CONTRIBUTING.md: record the FPiL rationale for deferring `grind` 
- LF/Lists.lean: rename `leb_n_Sn` -> `ble_n_Sn` and `leb_pred_n_n` ->
  `ble_pred_n_n`, so the names track `Nat.ble`, the function they are about.
- LF/Tactics.lean: Claude-generated `:::dev` note reminding us the remaining `leb` names (here and in `Logic`) should be renamed too -- they are also stated over `≤?`, which is notation for `Nat.ble`.


bcpierce00:
Everything here is minor.  I will merge in a little while -- probably no need for anyone else's eyes.

mwhicks1 (approved):

Issue #40, opened 2026-06-26
`inversion` case naming


ionathanch:
> One small UX issue I ran into is alts name handling. For example, in
> ```lean
> inductive Foo : Nat → Prop where
>   | mk : Foo 0
>
> example (h : Foo 0) : True := by
>   inversion h with
>   | mk => sorry
> ```
>
> `inversion` fails with `Invalid alternative name mk: Expected [anonymous]`. The issue is that we resolve alts using mvar tags instead of constructor names (we discard `ctorName` in `CasesSubgoal`.) It works in some cases by coincidence (when the goal tag happens to be the same as ctor name, e.g. `cons`).
>
> I think we have a design choice here: we're using meta cases API and not reusing elaborator's alt handling (`evalAlts`). In order to get this to work, either we need to duplicate a small part of that name normalization logic (e.g. stripping prefixes) locally after saving all `ctorName`s, or somehow reuse `Alt` construction in `ElimApp`. IMO the latter approach would be harder.

_Originally posted by @berberman in https://github.com/plclub/sf-in-lean/pull/37#pullrequestreview-4576607442_

ionathanch:
Fixed by #107

PR #107, opened 2026-07-20
fixes for `inversion` (#52, #40)

ionathanch:
Fixes for the `inversion` tactic that change behaviour:
* no longer fails on nonindexed inductives
* alternatives match to the constructor name rather than to the goal name, which may sometimes be missing if there is only one goal


dsainati1 (approved):
Does this also fix the second issue with the ‘silly’ inductive in the zulip thread

ionathanch:
yes, I'll add it to the tests

PR #108, opened 2026-07-20
Pass over section 1 of Basics

mwhicks1:
I went over section 1 of Basics, making the following changes:
1. I tried to be more careful about not assuming prior knowledge of certain terms (side effect, datatype, tuple type) and more directly identified inductive definitions as the "general mechanism" we use to define datatypes.
2. I tidied up some examples that had errors, to use the ```lean +error ...``` form
3. I (hopefully!) improved wording in a few places
4. Tidied typesetting
5. Added links to external resources 
6. Various comments added about further changes we could make

One thing I think we should do: We should add the namespaces part (1.8) to the very end of section 1, rather than where it is, which is breaking up the pedagogical flow. See what you think about that.

AI use: I had Claude check my approach to adding new terms. All the rest of the edits were mine.

### DIFF
--- a/LF/Basics.lean
+++ b/LF/Basics.lean
@@ -55,10 +55,11 @@ Here is what I did:
 
 ::::full
 The _functional style_ of programming is founded on simple, everyday
-mathematical intuitions: If a program has no side effects, then --
-ignoring efficiency -- all we need to understand about it is how it
+mathematical intuitions: If a program has no side effects, such
+as reading or writing files or network packets, then (ignoring efficiency)
+all we need to understand about it is how it
 maps inputs to outputs. That is, we can think of it as just a
-concrete method for computing a mathematical function. This direct
+concrete means for computing a mathematical function. This direct
 connection between programs and simple mathematical objects supports
 both formal correctness proofs and sound informal reasoning about
 program behavior. This is one sense of the word "functional" in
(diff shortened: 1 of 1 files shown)


LF/Basics.lean
```diff
@@ -83,56 +84,46 @@ some basic _tactics_ that can be used to prove properties of
...
+_datatypes_ -- types whose values are data, such as booleans,
+integers, and strings -- as primitives, Lean's extensive standard
+library _defines_ them, along with many common data structures
+besides, like lists and hash tables. It does so with a single
+powerful and general mechanism: the _inductive definition_.
+A type introduced this way is called an _inductive type_; the
+word "inductive" hints at the use of mathematical induction
+to prove statements about its values.
+
+To demonstrate how inductive definitions work, and make the
```
```suggestion
To demonstrate how inductive definitions work and make the
```

LF/Basics.lean
```diff
@@ -83,56 +84,46 @@ some basic _tactics_ that can be used to prove properties of
...
+point that the Lean library's use of them is routine and
+not magical, we will recapitulate most of the datatype definitions we
```
dsainati1: This phrasing to me begs the question of what a "magical" use of inductives would look like. Really what we are trying to say is that Lean can use inductives to define basically every datatype we care about.

dsainati1 (approved):

LF/Basics.lean
mwhicks1: I reworded it to say "To demonstrate how inductive definitions work, and illustrate their expressive power, we will recapitulate most of the datatype definitions we ..."

PR #109, opened 2026-07-20
Typesetting nits

bcpierce00:
(Can just be merged, I think, but see if you like the colors...)


.vscode/sf.code-snippets
```diff
@@ -0,0 +1,13 @@
+{
+  "Dev note (Benjamin Pierce)": {
```
mwhicks1: Should this be checked into the main repository with your specific name? Is there a way to make it general?

mwhicks1 (approved):

.vscode/sf.code-snippets
bcpierce00: I couldn't figure out a way to make it general.  Will remove fromn the repo.

PR #110, opened 2026-07-21
Slang: mention decide and other small improvements

xhalo32:
I haven't formatted the lines to fit within the margins as that seems silly to do manually.

### DIFF
--- a/TS/Slang.lean
+++ b/TS/Slang.lean
@@ -158,9 +158,9 @@ _Evaluating_ an arithmetic expression produces a number.
 def Aexp.eval (a : Aexp) : Nat :=
   match a with
   | num   n     =>  n
-  | plus  a1 a2 =>  eval a1 + eval a2
-  | minus a1 a2 =>  eval a1 - eval a2
-  | mult  a1 a2 =>  eval a1 * eval a2
+  | plus  a1 a2 =>  a1.eval + a2.eval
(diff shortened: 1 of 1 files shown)


HL/Imp.lean
```diff
@@ -607,6 +607,9 @@ def Bexp.eval (st : State) (b : Bexp) : Bool :=
     (and b1 b2).eval st = (b1.eval st && b2.eval st) := rfl
 ```
 
+It's worth noting that `≤` and `>` are `Prop`-valued, i.e. `a1.eval st ≤ a2.eval st` is a proposition, but {name}`Bexp.eval` returns a {name}`Bool` so Lean implicitly inserts a {name}`decide` coercion.
+You can observe the call to {name}`decide` by hovering over {name}`Bexp.eval_le` and {name}`Bexp.eval_gt`.
+
```
mwhicks1: Shouldn't this change be in Slang, instead, and then nothing needs to be said in Imp?

xhalo32: Good point! I also forgot some of the other changes that I made to Imp need to be changed in Slang also...

xhalo32:
@mwhicks1 I moved the mention of decide to Slang and changed the .eval calls like in https://github.com/plclub/sf-in-lean/pull/111

mwhicks1 (approved):

PR #116, opened 2026-07-22
Claude's fix for the inline {lean}`expr` issue

bcpierce00:


mwhicks1 (approved):
I have no idea if this is a good fix or not, but I'll approve so we can move forward. Perhaps invite a Verso expert to take a look at the whole file at some point?

bcpierce00:
One hesitation, not Claude-related: If we go very far in this direction, it's going to lead us to think about where the `{lean}` marker should be applied (also `{name}`)...  Which I think we should not spend time on right now...

PR #118, opened 2026-07-22
Finish full/terse balance pass and add note on structures

### LINKED ISSUE #75: LF/Basic has sections with no terse content
In my version of `_out/lf/terse/lean/LF/Basics.lean`, `add_succ` (and likely some other definitions) just don't make it to the output. I'm not exactly sure how and why that would happen, but it's definitely some kind of bug somewhere.

I could have done something wrong, but I used the `Makefile` on a clean clone of the repo.

hgoldstein95:
I'm just about done with my pass through basics to line things up. These last changes are mostly cosmetic except for the new section on structures that @mwhicks1 suggested.

Closes #75.

### DIFF
--- a/LF/Basics.lean
+++ b/LF/Basics.lean
@@ -40,12 +40,8 @@ be people that struggle with some of it.
 PRESENTATION ADVICE: Working with the .lean file directly in VS Code
 is recommended for the first few lectures, so students see exactly
 what's in the source file.
-:::
 
-:::dev "Michael Clarkson (clarksmr)"
-The first issue I had was that I don't have Lean installed. Since LF:Preface hasn't
-been ported, I had to figure out how to install it. New instructors will face the same issue.
-Here is what I did:
+If you don't have Lean installed yet:
 * Install Lean 4 through the VS Code extension.
 * Start a new terminal session to pick up environment changes.
 * Run `make`. I got many warnings about "expose"; are those expected?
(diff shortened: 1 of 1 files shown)


LF/Basics.lean
```diff
@@ -1193,10 +1201,25 @@ example : allZero (.bits .b0 .b0 .b0 .b0) = true  := by rfl
 end Playground
 ```
 
-:::dev "mwhicks1"
-Lean has structure types which provide tuple types directly.
-I could imagine reworking the `Nibble` definition to be a structure instead.
-Should we do that? Should we at least mention structures
+### Aside: Structures
```
berberman, line 1204: Now that we formally introduce `structure` in Basics, should we also move projections from Poly
https://github.com/plclub/sf-in-lean/blob/9da6bbfc2dd71359395b16612fb6d2290397f9e2/LF/Poly.lean#L822-L854

to Basics?

dsainati1: I don't think we can move them verbatim since the definition of products there is polymorphic. Are non-polymorphic pairs a useful enough type to give them their own introduction in Basics?

dsainati1:
Does this close https://github.com/plclub/sf-in-lean/issues/75 and if so can we add that to the issue description?

LF/Basics.lean
hgoldstein95: I don't feel strongly about it here vs. otherwise. @mwhicks1 just left a dev comment, and in the spirit of my rant from last meeting I figured I'd try to act on it rather than leaving it around.

Anyway, if I had to make a call, I think this is benign enough here, and we can go into more detail later.

dsainati1 (approved):

PR #120, opened 2026-07-23
Improve typesetting of devnotes and quiz labels

bcpierce00:
(Also remove the stray Material Icon Theme dependency.)


bcpierce00 (commented):
(Actually, the typesetting stuff was already merged.  So this is just removing the extension.)

PR #125, opened 2026-07-23
Move per-chapter notes from INPROGRESS.md into chapter :::dev notes

bcpierce00:
Migrate most of the notes out of INPROGRESS.md into :::dev notes at the top of each chapter.

There are still a few things in INPROGRESS... can we either delete those as stale or move them somewhere else?


bcpierce00:
Oops -- broke some comments. Will fix.

dsainati1 (approved):

PR #126, opened 2026-07-23
Explain field notation in UsingLean

### LINKED ISSUE #117: Clean up use of prefix and dot notations in early chapters
We mix use of postfix and prefix dot notation in Basics and Induction without explanation, which is confusing. Should go through these early chapters and make sure we only use prefix notation, and then add some text explaining the postfix notation in UsingLean

dsainati1:
Moves the explanation of field notation from Basics to UsingLean, and disables it for the first two chapters

Closes #117


LF/Poly.lean
```diff
@@ -1144,7 +1144,7 @@ We can use `filter` to give a concise version of the
 
 ```lean
 abbrev countoddmembers' (l : List Nat) : Nat :=
-  (filter odd l).length
+  (filter (·.odd) l).length
```
berberman: This looks unnatural to me... What do others think?

LF/Poly.lean
```diff
@@ -1084,13 +1084,13 @@ def filter {α : Type} (test : α → Bool) (l : List α) : List α :=
...
 ::::full
-For example, if we apply `filter` to the predicate `Nat.even`
+For example, if we apply `filter` to the predicate `·.even`
 and a list of numbers, it returns a list containing just the
 even members.
 ::::
 
 ```lean
-example : filter even [1, 2, 3, 4] = [2, 4] := by rfl
+example : filter (·.even) [1, 2, 3, 4] = [2, 4] := by rfl
```
xhalo32: I'm not a fan of this style when a plain `even` also works because it adds a lot of extra typing.

LF/IndProp.lean
```diff
@@ -967,16 +967,16 @@ theorem ev_4'' : Ev 4 := by
 /- In this way, we can also prove theorems that have hypotheses
     involving `Ev`. -/
 
-theorem ev_plus4 : ∀ n, Ev n → Ev (4 + n) := by
-  intro n Hn
+theorem ev_plus4 n : Ev n → Ev (4 + n) := by
+  intro Hn
```
berberman: Should we do:
```lean
theorem ev_plus4 {n : Nat} (h : Ev n) : Ev (4 + n) := by
```
instead?

LF/UsingLean.lean
```diff
@@ -546,16 +587,16 @@ talk about using `Nat.add_zero` and friends from now on.
        later files from breaking.
 
 ```lean
-theorem even_zero : even 0 = true := by rfl
+theorem even_zero : Nat.even 0 = true := by rfl
 
-theorem double_zero : double 0 = 0 := by rfl
+theorem double_zero : Nat.double 0 = 0 := by rfl
```
xhalo32, line 603: Would it make more sense to have these in the `Nat` namespace? Or what was the rationale behind this changes?

dsainati1: The rationale is to define some functions that Lean will print with field notation so that students can see how that works.

LF/UsingLean.lean
```diff
@@ -490,48 +488,91 @@ those proofs to make your logic clear.
...
+  | n + 1, m + 1 => eqb n m
 
 def minustwo (n : Nat) : Nat :=
   match n with
-  | .zero            => .zero
-  | .succ (.zero)    => .zero
-  | .succ (.succ n') => n'
+  | 0    => .zero
+  | 1    => .zero
+  | n' + 2 => n'
```
xhalo32: What's the reason that `eqb` and `minustwo` are not in the `Nat` namespace like `even`, `odd` and `double`?

xhalo32: Also should the `.zero`s be `0`s?

dsainati1: I wanted to have some in the namespace and some not so that students could see the difference between them. Would the idiomatic thing to do be to put all of these in the namespace?

LF/UsingLean.lean
```diff
@@ -490,48 +488,91 @@ those proofs to make your logic clear.
...
 ```lean
-def even (n : Nat) :=
+def Nat.even (n : Nat) :=
   match n with
-  | .zero           => true
-  | .succ .zero     => false
-  | .succ (.succ n) => even n
+  | 0     => true
+  | 1     => false
+  | n + 2 => even n
```
xhalo32, line 506: There's a little bit of magic hidden in `n + 2` as it's not immediately a pattern consisting of the constructors `n.succ.succ` or `(n + 1) + 1`. Do we want to mention what's going on here? (I don't remember what the feature is called from the top of my head)

LF/IndProp.lean
```diff
@@ -967,16 +967,16 @@ theorem ev_4'' : Ev 4 := by
 /- In this way, we can also prove theorems that have hypotheses
     involving `Ev`. -/
 
-theorem ev_plus4 : ∀ n, Ev n → Ev (4 + n) := by
-  intro n Hn
+theorem ev_plus4 n : Ev n → Ev (4 + n) := by
```
xhalo32:
```suggestion
theorem ev_plus4 (n : Nat) (h : Ev n) : Ev (4 + n) := by
```

LF/Poly.lean
bcpierce00: ... and removes quite a bit of readability!

LF/IndProp.lean
xhalo32: IndProp seems to currently favor that value arguments are explicit (and type arguments are implicit).

LF/UsingLean.lean
xhalo32: I see, that's a good idea. I didn't see it mentioned in the prose though. Nevermind, it's somewhat explained.

xhalo32:
> Okay, will revert even and odd to not live in the Nat namespace. How do we feel about double though?

That's not what at least I meant when suggesting to use a "plain `even`", it can still be `Nat.even` but just not forced to the dot-notation using `(·.even)`.

dsainati1:
Ahh okay. I moved all of these functions to live in the `Nat` namespace and removed usages of the `(·.even)`-style notation.

PR #132, opened 2026-07-24
IndProp versified!  Still needs to be carefully checked against origi…

bcpierce00:
…nal IndPropo.v


bcpierce00:
Closing in favor of #133

PR #133, opened 2026-07-24
Versify IndProp.lean and Maps.lean

bcpierce00:
IndProp.lean still needs to be carefully checked against original .v file

(Maps.lean is just for the sake of completeness -- probably doesn't matter one way or the other)

### DIFF
--- a/LF.lean
+++ b/LF.lean
@@ -14,7 +14,7 @@ import LF.Typeclasses
 import LF.Poly
 import LF.Tactics
 import LF.Logic
-import LF.IndPropVerso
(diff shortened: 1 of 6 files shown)


bcpierce00:
(Hm - I don't actually understand why this branch seems to include #132.)

bcpierce00:
OK, figured out what was happening and all is good now.  Please review this branch and forget about #132 (which I've deleted)

dsainati1 (approved):
I think the plan is to just delete `Maps` now that `Typeclasses` is merged in

PR #135, opened 2026-07-24
Tear down pre-versification infrastructure

bcpierce00:
Now that all the files are versified, we can rip out a bunch of cruft... :-)


bcpierce00:
(Should not be merged till after #133, (semi-)obviously!)

dsainati1 (approved):

PR #136, opened 2026-07-24
Lists: change abbrevs to definitions, rewrite proofs to use lemmas instead of rfl

xhalo32:
As discussed on Zulip, Lists.lean should use definitions instead of abbrevs. ~~This PR notably changes the following abbrevs into defs: `countoddmembers`, `sum`, `add` and `Bag`.~~ This PR removes `Bag` in favor of defining the functions on `NatList`s.

I moved some of the examples after the characterizing lemmas to make it possible to use them when proving via rewrites -- I also added a couple of examples before the graded ones to demonstrate the API. This is something we might want to do more widely in Lists.lean.

<details>
<summary>Outdated</summary>
I rewrote some proofs to use rewriting instead of `rfl` because it made sense to me for the model solutions to do it "the hard/tedious way". However, now some of the examples have quite long proofs which caused heartbeat timeouts with verso, so I made them a bit shorter.

Notice that a lot of the proofs now have to rewrite `← add_def` and `← empty_def` because the `[1, 2, 3]` notation expands to `::` instead of `add`. I'm not totally confident this is the best way to go. It might be better to just replace the notation with `add 1 (add 2 (add 3 empty))`...

I also added `count_add_def`, `member_add_def`, `removeOne_add_def`, `removeAll_add_def` in addition to the "positive" and "negative" lemmas. Tell me if this isn't ideal.
</details>

Finally, there are some missing `Nat.ble` characterizing lemmas which I added a dev notice about.

There are more dev notes about `included_cons_member`,  `count_cons_inc` and `nil_app`


LF/Lists.lean
```diff
@@ -648,45 +690,123 @@ A `bag` (or `multiset`) is like a set, except that each element
 can appear multiple times rather than just once.  One way of
 representing a bag of numbers is as a list.  The following definition
 introduces a new type, `Bag`, as an abbreviation for `NatList`.
+
+We define `Bag` as it's own new definition, so it needs an "API" so that we can work with it.
```
dsainati1:
```suggestion
We define `Bag` as its own new definition, so it needs an API so that we can work with it.
```

LF/Lists.lean
```diff
@@ -648,45 +690,123 @@ A `bag` (or `multiset`) is like a set, except that each element
...
+
 ```lean
-abbrev Bag := NatList
+def Bag := NatList
+
 namespace Bag
 ```
 
+::::full
+We define an "empty bag" and "adding an element to a bag" separately, and express the API using these definitions instead of `[]` and `::`.
```
dsainati1:
```suggestion
We define an "empty bag" and "adding an element to a bag" separately and express the bag's API using these definitions instead of `[]` and `::`.
```

LF/Lists.lean
```diff
@@ -574,7 +574,34 @@ def oddmembers (l : NatList) : NatList := solution!(
...
+    rw [oddmembers_cons, h, cond_false]
+
+example : oddmembers [1, 2] = [1] := by
+  rw [oddmembers_cons_odd _ _ rfl]
+  rw [oddmembers_cons_not_odd _ _ rfl]
+  rw [oddmembers_nil]
+
+theorem test_oddmembers : oddmembers [0, 1, 2, 3, 0] = [1, 3] := by
+  solution!
+    rw [oddmembers_cons_not_odd _ _ rfl]
```
dsainati1: I don't think students will have seen use of `rfl` as an argument before. Is this maybe a place to teach that?

xhalo32: Good question! I think this file uses quite a bit of equalities as assumptions so it's good to have here.

LF/Lists.lean
```diff
@@ -540,62 +540,105 @@ def foo (n : Nat) : NatList :=
...
+  rw [length_cons, length_cons, length_nil]
+
+example : countOddMembers [0, 2, 4] = 0 := by
+  solution!
+    rw [countOddMembers_def]
+    rw [oddMembers_cons_not_odd _ _ rfl]
+    rw [oddMembers_cons_not_odd _ _ rfl]
+    rw [oddMembers_cons_not_odd _ _ rfl]
+    rw [oddMembers_nil]
+    rw [length_nil]
```
berberman: Are we intentionally avoiding `rfl` because we want readers to manually rewrite here? I feel this is a bit tedious and diverges from the normal Lean experience...

berberman:
It seems that `gradeTheorem` names don't really match the definitions. Since they are placeholders for now, I guess we can fix them later.

LF/Lists.lean
xhalo32: My idea here is to favor using rewrites in the model solution. Also, otherwise the characterizing lemmas would be used much less. 

On line 810 in the file we mention that rfl can also be used.

One thing I've thought about is to e.g. introduce a variable in the bags in some of the examples to make it genuinely need rewrites instead of just rfl. Would that be good motivation?

LF/Lists.lean
```diff
@@ -209,7 +209,7 @@ def bad_fst (p : NatProd) : Nat :=
 -- Can't match on multiple values with pair patterns:
 def bad_sub (n m : Nat) : Nat :=
   match n, m with
-  | ⟨0,        _⟩       => 0
+  | ⟨0,        _⟩        => 0
```
xhalo32, line 175: This is not aligned on my screen (GitHub mobile)

xhalo32:
@berberman thanks for the fixes and review!

LF/Lists.lean
```diff
@@ -648,50 +690,140 @@ A `bag` (or `multiset`) is like a set, except that each element
...
+@[elab_as_elim, induction_eliminator]
+theorem inductionOn
+    {motive : Bag → Prop}
+    (s : Bag)
+    (empty : motive Bag.empty)
+    (add : ∀ n t, motive t → motive (Bag.add n t)) :
+    motive s := by
+  induction s with
+  | nil => exact empty
+  | cons n t ih => exact add n t ih
```
berberman: This trick makes `induction s` select a custom induction principle for `Bags`, so no need to rewrite `nil` back to `empty`. But why don't we just define a new inductive type and give it a nice notation? I feel like `def Bag := NatList` is causing more troubles (e.g. `@[match_pattern]`, carefully not to unfold into `nil` and `::`, etc.) and I don't see why this abbreviation is useful.

berberman: Hmm interesting, the student project failed to build because `Bag.add` there is a `sorry` placeholder, causing this induction principle invalid.

xhalo32: The point is to practice encapsulation. The match patterns are not causing trouble, however using `[1,2,3]` notation for bags is, because it exposes `::` and `nil`. This could be fixed with a notation type class but it's too early to introduce that. I suggest we switch from `[1,2,3]` to `add 1 (add 2 (add 3 empty)))`

bcpierce00:
What's needed before we can merge this PR?

LF/Lists.lean
berberman: But doing `induction` will reveal the underlying structure (`nil` and `cons`), causing the lemmas inapplicable unless first rewriting it backwards to list. Defining notations won't help. This custom induction solves this issue but IMO it's not worth it -- we could have just defined a new inductive type.

bcpierce00:
Please signal when you think this has converged, @xhalo32

LF/UsingLean.lean
```diff
@@ -505,7 +505,9 @@ def Nat.even (n : Nat) :=
   | 1     => false
   | n + 2 => even n
 
-def Nat.odd n := !(even n)
+def Nat.odd (n : Nat) := !(even n)
+
+theorem Nat.odd_def (n : Nat) : n.odd = !(n.even) := rfl
```
dsainati1, line 510: This is a pretty different style of definition than students will have seen up to this point. What is the benefit of defining things this way?

xhalo32: It's used in Lists.lean.

dsainati1: Oh this is another characterizing lemma! I don't even think I realized this since it's not a match statement. I think it could be worth just a sentence pointing this out.

xhalo32: I'm not sure where to put this sentence. There are some lemmas like `even_zero`, `double_zero`, `double_succ` below (which are not in the `Nat` namespace), but they don't have any prose explaining them.

dsainati1: Yeah it's probably fine actually. I think I just confused myself.

xhalo32:
Thanks @berberman again for the improvements, they all look good to me! We can also fix the style (implicit parameters etc) later.

The only thing broken now is the lf-terse build. It seems like they are caused by the `:::full` blocks, for example all content in the exercises up to `removeOne` get stripped from the terse build.

mwhicks1:
> It seems like they are caused by the `:::full` blocks, for example all content in the exercises up to `removeOne` get stripped from the terse build.

I ran into this also with some of the `TS/`* work I just did. The solution I implemented was to pull the exercises out of `::::full` to the top level. Is there any reason not to do that?

bcpierce00:
Yes: It makes the terse version ugly.  So we should do it as little as possible and make sure to leave a comment for the instructor saying which bits to skip over when we are forced to do it.

berberman:
Renamed some definitions in idiomatic Lean style -- this diverges from the Rocq version, but I think it's giving better presentation. Will need to update Poly.

LF/Lists.lean
```diff
@@ -327,16 +299,22 @@ Some notation for lists to make our lives easier:
...
 
+
 ```lean
 scoped infixr:65 " :: " => cons
 scoped macro (priority := high) "[ " elems:term,* "]" : term => do
   elems.getElems.foldrM (``(cons $(⟨·⟩) $(⟨·⟩))) (← ``(nil))
 ```
 
+We first define `::` as right-associative notation for {name}`cons`,
+and then define list notation, allowing us to write `[1, 2]` instead of `1 :: 2 :: []`. We also have an unpander that prints `cons 1 (cons 2 [])` to `[1, 2]`.
```
dsainati1: Should this say "expander"?

dsainati1 (approved):

LF/Lists.lean
xhalo32: It should say "unexpander", as it goes from expanded terms back to syntax

dsainati1:
```suggestion
and then define list notation, allowing us to write `[1, 2]` instead of `1 :: 2 :: []`. We also have an unexpander that prints `cons 1 (cons 2 [])` to `[1, 2]`.
```

berberman: I was lying... Actually we don't have one :upside_down_face:

PR #137, opened 2026-07-24
Contributing updates

xhalo32:
Fixes a couple of duplicated sections. Adds a section about when to use abbreviations that aligns with https://github.com/plclub/sf-in-lean/pull/136


CONTRIBUTING.md
```diff
@@ -270,46 +270,6 @@ Related notation introduced alongside tactics: anonymous constructor
   over the separate `case` syntax *and* over the bare `·` goal selector — i.e. prefer
   `cases h with | …` / `induction h with …`.
   Put each alternative on its own unindented line beginning with `|`.
-  
-* **`rewrite` before `rw`** (see tactic chart above) --
-  `rw [h]` is roughly `rewrite [h]; rfl`, which is too strong at
-  first: it hides the closing `rfl` and makes proofs step
-  confusingly (the goal vanishes when you step past the final `]`).
-  We introduce `rw` specifically in `Induction.lean` and use from
-  then on.
```
mwhicks1, line 282: I don't see the above recommendation in the updated text. Did you mean to delete it?

The rest looks great! I wonder, at this point: Should we move the Style guidelines into their own file, and reference that file from `CONTRIBUTING.md` ?

xhalo32: Good catch, that's a mistake

bcpierce00 (approved):

PR #138, opened 2026-07-24
Refine Poly

berberman:
1. Improved the wording explaining inductive parameters
2. Replaced `--` with em dash (IIRC verso doc recommends this)
3. Added `lean`/`name` inline tags
4. Cleaned up outdated comments
5. Improved induction style
6. Chose proper implicit arguments for theorems
7. Replaced `h`/`t` with `head`/`tail` in list deconstructions (as we usually use `h` for hypothesis)
8. Fixed missing terse in some sections
9. Fixed one incorrect quiz solution
10. Changed forall quantifiers to parameters in some theorems (which I think it's the preferred style)


LF/Poly.lean
```diff
@@ -13,19 +13,9 @@ htmlSplit := .never
...
-Using the built-in definition of `List.reverse` is dramatically more complicated
-than implementing our own reverse function, since it is implemented in terms of an auxiliary
-function.
-
-The associativity of `++` in Lean is different than Rocq. In Rocq the definition
-of `app_assoc` is `l ++ m ++ n = (l ++ m) ++ n`, but in Lean it's
-`l ++ m ++ n = l ++ (m ++ n)`.
-:::
+```lean -show
+variable (α β γ : Type)
```
dsainati1: If we are going to be using this here we probably need to explain it.

LF/Poly.lean
```diff
@@ -734,44 +625,37 @@ chapter, for practice with polymorphism. Complete the proofs below.
 You will likely find useful the following lemmas about append and length
 from Lean's standard library:
```
dsainati1: We should point out that these are exactly the characterizing lemmas for append.

LF/Poly.lean
```diff
@@ -849,21 +732,21 @@ example : (3, 5).1 = 3 := by rfl
 example : (3, 5).2 = 5 := by rfl
 ```
 
-The notation `α × β` is syntactic sugar for `Prod α β`.
+The notation {lean}`α × β` is syntactic sugar for {lean}`Prod α β`.
```
dsainati1: Agree with BCP's comment below, let's add a note here to tell people how to type the product symbol

LF/Poly.lean
```diff
@@ -951,19 +832,26 @@ example : unzip [(1, false), (2, false)] = ([1, 2], [false, false]) := solution!
...
-Did we literally see `Option Nat` or was it spelled some other way?
-:::
-
 ::::full
 Our last polymorphic type for now is _polymorphic options_.
-Lean's standard library provides `Option α`, with constructors
-`none` and `some x`. (We already saw `Option Nat` in the
-previous chapter.)  Let's briefly look at the definition:
+Lean's standard library provides {lean}`Option α`, with constructors
+{name}`none` and {lean}`some`. (We already saw {lean}`Option Nat` in the `Lists` chapter.)
```
dsainati1:
```suggestion
{name}`none` and {lean}`some`. (We already saw {lean}`NatOption in the `Lists` chapter.)
```

LF/Poly.lean
berberman: It's only for being able to use `` {lean}`α` `` in verso, and has no impact on the later Lean code. So I'm hesitant to add an explanation -- we don't actually use `variable` in the right way, thought it does show up in extracted projects, which is a bit concerning...

bcpierce00: We should really try hard to keep "can’t explain what this is or tell readers how to use it themselves" to an absolute minimum... like close to zero if possible.

dsainati1: Yeah if this is not a correct usage of variable and also shows up in student-visible output I think it’s probably not worth it just for use with verso rendering

berberman: Never mind, I figured out how to hide it from student projects.

dsainati1 (approved):

PR #145, opened 2026-07-25
Tidy CONTRIBUTING.md and split out STYLE.md

bcpierce00:


CONTRIBUTING.md
```diff
@@ -77,20 +61,17 @@ For discussions, we use a combination of tools.
   In-text comments can also be used for coordinating work on specific
   issues.
 
-We prefer _not_ holding discussions in annotations on PRs, because
-they tend to either get lost when the PR is merged or delay merging.
-Putting very local or short-term comments in this medium is fine -- or
-you can just make the change by directly editing the PR, if you think
-it's clear.
+- **On PRs:** We prefer _not_ holding longer discussions in annotations on 
+  PRs, because they tend to either get lost when the PR is merged or delay merging. Putting very local or short-term comments in this medium is fine -- or you can just make the change by directly adding commits to the PR, if it's clear whaet needs to be done.
```
xhalo32: Typo: whaet

STYLE-CHECKING.md
```diff
@@ -0,0 +1,126 @@
...
+### Lean conventions
+- **LEAN-1 — Mathlib alignment.** Follow the Mathlib style and naming
+  conventions, except where an SFL pedagogical convention overrides them.
+- **LEAN-2 — Idiomatic Lean.** Write definitions and proofs the way a working
+  Lean engineer would; depart from idiom only temporarily, only for pedagogy,
+  and say why.
+- **LEAN-3 — Tactic discipline.** Introduce and use tactics in the established
+  order; never use a tactic before the chapter has taught it.
+- **LEAN-4 — Naming and namespaces.** Follow the naming/namespace rules in
+  `CONTRIBUTING.md`.
```
xhalo32: Should this be STYLE.md?

xhalo32: Same with the occurrences of CONTRIBUTING.md below.

CONTRIBUTING.md
bcpierce00: Feel free to just fix things at this level instead of leaving comments...

xhalo32: Sorry, I was just reading and reviewing on mobile and don't have access to my machine at the moment. In the future, if I'm on an actual development-capable system I will just push a fix instead.

bcpierce00: Sounds good!

scripts/style_check.py
```diff
@@ -0,0 +1,255 @@
+#!/usr/bin/env python3
+# This file is maintained by Claude (AI-generated).
+"""
+style_check.py  —  Semi-automatic conformance checks for STYLE.md.
```
mwhicks1, line 4: Great idea!

CONTRIBUTING.md
```diff
@@ -3,23 +3,14 @@
...
 ## Top-level orientation
 ### Guiding Philosophy
 
-These are the tenets of the SFL effort, in order. Consult these tenets
-when making a change: If your change is supported by them, then
-make it; no need for excessive coordination. If it is not supported by
-at least one tenet, then either your change is out of scope or a tenet
-is missing. If you are not sure then have a discussion (see below), and
-refer to the tenets to drive a decision (potentially updating the tenets).
-
```
mwhicks1, line 22: I don't want this deleted.

CONTRIBUTING.md
```diff
@@ -123,11 +105,13 @@ Our CI uses a small GitHub Actions workflow:
...
 
 We also have branch protection enabled, which requires the following before merging:
-* At least one approval before merge is allowed
-* Linear history (use rebase and squash merge)
+* At least one approval from someone on the `sf-mergers` list before merge is allowed
+* No outstanding conversations on the PR
 * CI build succeeds
 
-## Cadence of code changes
+We use a linear history, with squash-and-rebase merges.
```
mwhicks1, line 119: Do we? I believe I have seen past violations of this claim. Are we properly enforcing it on GitHub (I'm not actually sure)?

CONTRIBUTING.md
```diff
@@ -167,54 +151,84 @@ review it. Please address these comments in a subsequent commit, either
 making appropriate changes or else responding in the file with your
 own comments.
 
+### Stacked PRs
```
mwhicks1, line 161: Do we need this section? Seems like standard Git-fu when working on a multi-person project.

CONTRIBUTING.md
```diff
@@ -167,54 +151,84 @@ review it. Please address these comments in a subsequent commit, either
...
 ## Tools for coordinating work
 
 We prefer to move fast rather than over-coordinate synchronously, but
 we also want to avoid conflicts when possible. We use the [GitHub
 issue tracker](https://github.com/plclub/sf-in-lean/issues) for
 recording large tasks that need to be done (small or local tasks can
 just be recorded in comments in the affected .lean file) and for
-keeping track of work in progress that other people should be careful
-not to step on.
+keeping track of work in progress, plus the status meta-issue for getting an overview of who is working where.
```
mwhicks1: Do we have this meta-issue anymore? If so: Link it from here.

mwhicks1 (changes requested):

CONTRIBUTING.md
bcpierce00: I'm pretty sure GitHub is set up to (and does) enforce this.

CONTRIBUTING.md
bcpierce00: maybe we don't need it, but it was new to me, especially the final bit, which has messed me up a few times in the past week.

CONTRIBUTING.md
bcpierce00: Yes, this is the new one.

CONTRIBUTING.md
bcpierce00: ... but now that I know it, I don't need it written down here.  So perhaps I was the last person that needed the information and we should delete it. :-)

scripts/style_check.py
bcpierce00: There will be some real design involved in setting up something that really works and is not too heavy, so I backed off from doing it now.  But my long-term intention is that stuff that does into STYLE.md can eventually be morphed or evolved into something like this.

(One big design issue, FYI, is how to persist decisions about things like writing style. We need a way to record things like "I know the Chicago Manual says there should be a comma here, but I have good reasons for omitting it and I don't want to be reminded about it every time I run this script.")

CONTRIBUTING.md
bcpierce00: OK, I will restore it.  It isn't clear to me (even on re-reading) that the paragraph adds any information, though, that readers will not get from the section header, the tenets themselves, and a few brain-tokens worth of easy inference.

bcpierce00 (commented):
@mwhicks1 - Mike, please go ahead and merge this when you've had a chance to look at my responses to your comments.

CONTRIBUTING.md
mwhicks1: I will move the text to after the tenets. In my experience, people read this stuff as throwaway and don't think carefully about how the tenets can be used during decision-making. They also don't always appreciate that the order matters, e.g., when breaking ties. And I wanted to emphasize moving faster if you are matching the tenets in the change you are making.

mwhicks1 (approved):

PR #151, opened 2026-07-26
Bunch of tidying and cleaning up typesetting

bcpierce00:
(And some notes for things yet to be fixed.)

### DIFF
--- a/LF/Basics.lean
+++ b/LF/Basics.lean
@@ -957,6 +957,13 @@ end RGB
 Top-level definitions can also be prefixed by a namespace,
 which opens the namespace temporarily for the body of the definition.
 
+:::dev "Claude" NOW
+Rendering bug in *student* and *terse* (solutions is fine): the leading
+`--- …` triple-dash comment lines in this block and the next one each render
+*twice* in a row (e.g. two consecutive `--- this works, because …` lines). The
+solutions build shows each once, so this is a Verso rendering quirk with
+`---`-style comments in the elided builds, not a source duplication.
+:::
 ```lean
 --- this works, because the definition is qualified by `RGB.`
 def RGB.myOtherBlue : RGB := myBlue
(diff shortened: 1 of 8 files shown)


LF/Basics.lean
```diff
@@ -957,6 +957,13 @@ end RGB
 Top-level definitions can also be prefixed by a namespace,
 which opens the namespace temporarily for the body of the definition.
 
+:::dev "Claude" NOW
```
mwhicks1, line 960: There are lots of these notes claiming rendering problems, added by Claude. Are they human-confirmed? Are some of them potentially stale (the name of this PR implies that typesetting issues are being addressed) ?

LF/Lists.lean
```diff
@@ -1055,7 +1055,10 @@ What's the best Lean markup for a displayed equation? The markup below is going
 :::
 
 _Theorem_: For all lists `l1`, `l2`, and `l3`,
-    `(l1 ++ l2) ++ l3 = l1 ++ (l2 ++ l3)`.
+
+```display
+(l1 ++ l2) ++ l3 = l1 ++ (l2 ++ l3).
+```
```
mwhicks1, line 1061: Is it weird for the `_Theorem_` part to not also be in the `display` block? Same question for what's below.

LF/Poly.lean
```diff
@@ -961,9 +963,12 @@ Lean's standard library provides `Option α`, with constructors
 `none` and `some x`. (We already saw `Option Nat` in the
 previous chapter.)  Let's briefly look at the definition:
 
+```display
 inductive Option (α : Type) : Type where
-   | none : Option α
-   | some (x : α) : Option α
+  | none : Option α
+  | some (x : α) : Option α
+```
```
mwhicks1: Why is this a `display` and not `lean` ?

mwhicks1 (approved):

LF/Basics.lean
bcpierce00: Some are human-confirmed, some I have still to look at because fixing them seems to involve real work. All are newly confirmed (by Claude) yesterday, so I think they are real.

LF/Lists.lean
bcpierce00: The `display` block is for inset material (i.e., I was thinking of TeX terminology, like math displays).  Perhaps it would be better to call it `inset`?

LF/Poly.lean
bcpierce00: Because it's defined in the standard library, I think

bcpierce00 (commented):
Mike, feel free to merge it when you are satisfied with your comments...

LF/Poly.lean
mwhicks1: Agreed, but: Can't we wrap in a namespace, like we've done for other standard library types? Feels better to show that it's Lean that really typechecks. I'll make this change unless you object, when I merge.

bcpierce00: This is fine, FTR.

PR #153, opened 2026-07-26
Add a :::test directive for checks that run but aren't published

mwhicks1:
New verso directive `::::test` for Lean checks that run but aren't published. Motivation:

`::::hide` drops its body at *elaboration*, so Lean code inside it is inert text that is never checked. A `:::test` body is elaborated for real (so it can fail the build) but renders nothing in HTML/TeX and is skipped by the saver, so it reaches neither the published book nor the extracted `.lean` projects. Contrast `:::ignore`, which elaborates *and* renders, and is dropped only from the extracted `.lean`, not the HTML.

All of this (non user facing) code was authored by Claude.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -129,7 +129,14 @@ build and passed through as a labelled comment in the generated `.lean` files
 `PotentialImprovement` notes are dropped from every build. The other annotation
 directives (`instructors`/`hide`/`grade`/`solution`) are noops (bodies dropped
 from every build); the tag name reserves each for a future build that treats it
-differently.
+differently.  "Dropped from every build" means dropped *at elaboration*: Lean
+code inside `::::hide` is inert text that is never checked — right for the
+parked, aspirational code the LF chapters' hide blocks hold, but the opposite of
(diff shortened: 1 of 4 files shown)


berberman:
Maybe it's worth adding one more directive, but I think this should work after #149:

````lean
:::ignore

```lean -show
-- ...
```

:::
````

What checks/commands will we run secretly?

mwhicks1:
> Maybe it's worth adding one more directive, but I think this should work after #149:
> 
> ```lean
> :::ignore
> 
> ```lean -show
> -- ...
> ```
> 
> :::
> ```

I find this unintuitive: Why would a `-show` directive _not_ have the text rendered? I would have expected the opposite.

> What checks/commands will we run secretly?

In STLC I use it to test that the notation is working. I don't want to show all of the examples in the actual text.

berberman:
`-show` means setting the `show` verso flag to `false` (the default value is `+show`). It's a common pattern in Lean, e.g. we also have `simp +contextual` in core Lean.

PR #154, opened 2026-07-26
More verso commands for enhanced rendering

mwhicks1:
Extended the `bnf` directive to allow an optional fourth column, as a production label. Also adjusted typesetting so that rendering markup intended for HTML only, distinguishing terminals from nonterminals, is stripped in the generated lean for cleaner plain-text. Finally, tweaked so that a description list's terms are emitted into the extracted `.lean`; useful for lists of examples (in prep for STLC).

All code written by Claude, human reviewed; no user-facing content.


bcpierce00:
Cool.

Issue #155, opened 2026-07-26
Use Em Dashes


berberman:
According to [verso doc](https://verso-user-manual.netlify.app/Verso-Markup/#verso-markup), we should write `—` (`\em`) instead of `--` in general, as verso won't do the conversion for us.

bcpierce00:
At some point we should do a global change.  But we'll need to do it carefully to (a) avoid changing -- in other contexts (inline comments, e.g.) and (b) avoid a million conflicts.

rogerburtonpatel:
Now this is a task for an LLM (joke)

PR #158, opened 2026-07-26
Types: bnf grammar, and notation encoding behind :::details

mwhicks1:
- The informal syntax becomes a ` ```bnf ` block, so terminals and non-terminals are distinguished in the book and the extracted `.lean` gets an aligned grammar rather than one long line.
- The `tm` syntax category and the delaborator move into `:::details`.  Both are encoding: a reader following the metatheory can skip them, and the prose above each block already says what they achieve.

Follows the pattern established in the Stlc chapter.

### DIFF
--- a/TS/Types.lean
+++ b/TS/Types.lean
@@ -38,7 +38,7 @@ challenging.
 
 Spending an entire 80-minute class on this chapter feels about right.
 Going through the proofs of progress and preservation very carefully,
-at the board, was critical.  Doing the quizzes together is good -- a few
+at the board, was critical.  Doing the quizzes together is good — a few
 more quizzes would be better.
 
 For this lecture, I (BCP) find it useful to make a physical 1-page
@@ -68,7 +68,7 @@ might be wonderful additions to the TERSE notes.
 :::
 
 ::::full
-Our next major topic is _type systems_ -- static program analyses that
+Our next major topic is _type systems_ — static program analyses that
 classify expressions according to the "shapes" of their results.  We'll
 begin with a typed version of the simplest imaginable language, to
 introduce the basic ideas of types and typing rules and the fundamental
@@ -92,7 +92,7 @@ New topic: _type systems_
 To motivate the discussion of type systems, let's begin as usual with a
 tiny toy language. We want it to have the potential for programs to go
 wrong because of runtime type errors, so we endow it with two kinds of
-data -- numbers and booleans -- where not every operation is defined on
+data — numbers and booleans — where not every operation is defined on
 both types of data. For example, program terms like `5 + true` and
 `if 42 then 0 else 1` use undefined operator/data-type combinations.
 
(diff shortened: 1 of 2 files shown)


TS/Types.lean
```diff
@@ -140,6 +146,18 @@ written inside `<{ … }>` -- for example `<{ if false then 0 else succ 0 }>` --
 mirroring the informal grammar above.  A bare identifier is spliced as a Lean
 term (so a variable `t` is written just `t`); `~e` escapes an arbitrary Lean
 expression, and `( … )` groups.
+
+You do not need to understand exactly how the declarations below work; every
+object language in this book is given its syntax the same way, so it is worth
+seeing the pattern once:
+
+- `declare_syntax_cat` adds a new non-terminal to Lean's grammar -- here `tm`,
```
berberman:
```suggestion
- `declare_syntax_cat` adds a new non-terminal to Lean's grammar — here `tm`,
```

TS/Types.lean
```diff
@@ -140,6 +146,18 @@ written inside `<{ … }>` -- for example `<{ if false then 0 else succ 0 }>` --
...
+  the terms of this chapter's language.
+- Each `syntax` directive declares one production of that non-terminal, with
+  annotations fixing precedence, and the last one declares the `<{ … }>`
+  brackets that let a `tm` appear where Lean expects a term.
+- `macro_rules` then translates each production into the corresponding
+  constructor of {name}`Tm`.
```
berberman:
```suggestion
- `macro_rules` then translates the resulting syntax forms into the
  corresponding constructors of {name}Tm.
```

TS/Types.lean
berberman, line 174: It might be wroth thinking about the precedence here. First, prefix chains like `succ succ 0` won't parse because the inner `succ 0` has precedence 75 while the slot requires 76. Next, nested if like `succ if true then 0 else 0` will be rejected because `if` has precedence 50. The following code
```
if true then 0
else if false then succ 0
else pred 0 
```
will also be rejected similarly. Is this what we want, given that the BNF grammar above says every recursive position is an unrestricted `t`? Should we instead do

```lean
syntax:75 ident ppHardSpace tm:50 : tm
syntax:50 "if " tm:50 " then " tm:50 " else " tm:50 : tm
```

to allow them?

berberman (approved):

TS/Types.lean
mwhicks1: I'm not sure, to be honest. I'll wait for others to weigh in!

TS/Types.lean
mwhicks1: Based on an earlier convo I believe we are assuming that `--` will translate to an em dash, so this is correct AFAICS. This is the same convention used elsewhere in the book.

berberman: #155

mwhicks1: Did this change in both Smallstep and Types, now.

TS/Types.lean
mwhicks1: I decided not to change anything, since what's there is the same as what was in the Rocq. So I put a dev note instead that we can revisit.

mwhicks1: In the meantime Claude discovered a delaborators bug so I pushed its fix for that too.

PR #163, opened 2026-07-27
Decompose SFLMeta.Save and Improve lean block performance

berberman:
Previously `` ```lean `` code block was elaborated up to four times: 1) original source, 2) teacher variant (after removing `-- SOLUTION`/`-- END SOLUTION`), 3) student variant (after replacing `solution!` with `sorry`), 4) terse variant (after replacing `workinclass!` with sorry). This PR makes teacher variant resue the original elaboration result by stripping off any line not in teacher code from the result. Moreover, it makes student variant elaborated only if the student source differs from the original source, while previously this elaboration always happened.

Besides, this PR splits the gigantic `SFLMeta.Save` into multiple `SFLMeta.Save.*` files and repalces fragile variant string with a new type.

With this change, the full build speeds up ~2x on my machine.


bcpierce00:
What problem does this fix?  (I'm sure the fix is needed if you think it is, so I'll also go ahead and approve it now.  But I'm interested...)

bcpierce00 (approved):

berberman:
Trying to fix https://github.com/plclub/sf-in-lean/blob/7fddc00221d7362ce269a9be3563b1e0be370bb6/LF/Basics.lean#L2578-L2588

I expect a small compilation overhead because previously we incorrectly skipped some elaborations, but I didn't expect the failure on Lists... I will investigate later.

dsainati1 (approved):
Very excited for this one

Issue #89, opened 2026-07-15
Chapter status 


bcpierce00:
# Chapter status

Once chapters have been converted to Verso, we need to give each one a final check, comparing the Verso version to the original `.v` file, line by line. (The _Beyond Compare_ app is great for this.) After that, we need to manually rewrite any Claude-drafted proofs, make sure the technical material is complete, and give it a final proofread.  This table records the status of each chapter, so that we only do each step once.

## Volume 1 — Logical Foundations (LF)

| Chapter | Leanify | Check&nbsp;.v | Humanify | Tech | Proofread | Comments |
| --- | :---: | :---: | :---: | :---: | :---: | --- |
| Preface | [x] | [x] | [x] | [x] | [x] | |
| Basics | [x] | [x] | [x] | [x] | [x] | |
| Induction | [x] | [x] | [x] | [x] | [ ] | TODO: Move Code Actions from UsingLean to Induction. |
| UsingLean | N/A | N/A | [x] | [x] | [ ] | New chapter. Needs careful reading. TODO: Expand discussion of dsimp (motivating with useful exercises, etc.)|
| Lists | [x] | [x] | [x] | [x] | [ ] | Revert implicit arguments in this chapter.|
| Poly | [x] | [ ] | [x] | [x] | [ ] |  |
| Tactics | [x] | [x] | [x] | [x] | [x] | Explain named arguments (ie. things like `f (x := 2)`|
| Logic | [x] | [ ] | [ ] | [ ] | [ ] | |
| IndProp | [x] | [ ] | [ ] | [ ] | [ ] | |
| Automation | N/A | N/A | [ ] | [ ] | [ ] ||
| Typeclasses | N/A | [ ] | [ ] | [ ] | [ ] | |
| ProofObjects | [ ] | [ ] | [ ] | [ ] | [ ] | |
| IndPrinciples | [ ] | [ ] | [ ] | [ ] | [ ] |   |
| Rel | [ ] | [ ] | [ ] | [ ] | [ ] |  |
| CurryHoward | [ ] | [ ] | [ ] | [ ] | [ ] |  (probably not translate) |
| Postscript | N/A | N/A | [x] | [x] | [x] |  |

## Volume 2 — Type Systems (TS)

| Chapter | Leanify | Check&nbsp;.v | Humanify | Tech | Proofread | Comments |
| --- | :---: | :---: | :---: | :---: | :---: | --- |
[...]

PR #164, opened 2026-07-27
Typeclasses: fold in the rest of the Maps material

mwhicks1:
- Copies the relevant content of the deprecated `LF/Maps.lean` into this chapter, with a few adjustments, taking the prose from the original Rocq `Maps.v`: the remaining `TotalMap` theorems (`apply_empty`, `update_neq`, `ext`, `update_shadow`, `update_same`, `update_permute`) and a new Partial Maps section (`PartialMap`, `→ₚ`, the six lifted lemmas, and map inclusion as `⊆` with `update_subset`). The last four total-map lemmas become exercises. This is what the Type Systems volume's STLC chapters need for weakening.
- I had Claude look at the proofs -- where it proposes a deviation I had it leave a :::dev note. For proofs of exercises I just had it do the proofs.
- Moves `Typeclasses` to the end of the LF book in `LF.lean`.
- Note: Exercises are deliberately not wrapped in `::::full`: their results are used later in the chapter, and `::::full` content is dropped from the terse build, which left the extracted terse project with undefined constants.

Amendment: I had Claude compare this file with Maps.v and carried over material from there that seemed to be missing, including examples and prose.


STYLE.md
```diff
@@ -32,14 +32,16 @@ should be kept in sync as chapters are rewritten.
 | `Tactics`         | `intros`, `apply` (and `apply … at`), `replace`, `symm`, `injection`, `injections`, `congr`, `assumption`, `contradiction`, `unfold`, `split` |
 | `Logic`           | `constructor`, `obtain`, `left`, `right`, `ext`, `by_cases`, `exfalso` |
 | `IndProp`         | `rcases`, `subst` |
-| `Typeclasses`     | `decide` |
 | `Automation`      | `lia`, `try`, `repeat`, `specialize`, `trivial`, `simp`, `generalize` |
+| `Typeclasses`     | `decide` |
 | `HL/Imp`          | *(none new)* |
```
dsainati1, line 37: I guess Typeclasses could go after Automation, if we want it to. This would deviate slightly from the order in the Rocq book, where Maps (which Typeclasses is replacing) was between IndProp and Auto. Is there a particular reason for this reordering?

mwhicks1: Oh that wasn't a deliberate choice on my part, I thought that was what we wanted -- I took the order from the "Chapter status" issue https://github.com/plclub/sf-in-lean/issues/89. With this ordering, the proofs use `simp`. If we move it as you suggest, we should change them but it's not a big deal to do that.

dsainati1: @bcpierce00 what do you think? There is no explicit dependency between Automation and Typeclasses in either order. Do you have any preference?

LF/Typeclasses.lean
```diff
@@ -535,11 +679,157 @@ example : ({ 1 ↦ 2, 1 ↦ 3 } : TotalMap Nat Nat)[1]! = 2 := rfl
...
+
+notation a " →ₚ " b " ; " m => PartialMap.update m a b
+```
+
+We can also hide the last case when it is empty:
+
+```lean
+notation a " →ₚ " b => PartialMap.update ∅ a b
+
+def examplepmap : PartialMap String Bool := "Church" →ₚ true ; "Turing" →ₚ false
```
xhalo32, line 743: Should we add this notation for total maps too?

LF/Typeclasses.lean
```diff
@@ -535,11 +679,157 @@ example : ({ 1 ↦ 2, 1 ↦ 3 } : TotalMap Nat Nat)[1]! = 2 := rfl
...
+One last thing: for partial maps, it's convenient to introduce a notion of map inclusion, stating that all the entries in one map are also present in another. Lean already has notation for this -- `m₁ ⊆ m₂` -- which we get by supplying a {name}`HasSubset` instance.
+
+```lean
+def Subset (m₁ m₂ : PartialMap α β) : Prop :=
+  ∀ (a : α) (b : β), m₁[a] = some b → m₂[a] = some b
+
+instance : HasSubset (PartialMap α β) where
+  Subset := PartialMap.Subset
+
+def subset_def (m₁ m₂ : PartialMap α β) :
```
xhalo32: The dev comment below is right, this should be a theorem rather than a def.

LF/Typeclasses.lean
```diff
@@ -535,11 +679,157 @@ example : ({ 1 ↦ 2, 1 ↦ 3 } : TotalMap Nat Nat)[1]! = 2 := rfl
...
+  · simp only [update_neq _ _ _ eq] at *
+    rw [h a' b' eq]
+    symm
+    assumption
+
+end PartialMap
+```
+
+:::dev "Claude" PotentialImprovement
+In the above proof, the name `eq` is bound twice -- first as the equation `m₁[a'] = some b'`, then
```
xhalo32: I agree the proposed version that doesn't shadow the hypothesis is better.

xhalo32 (approved):

STYLE.md
mwhicks1: I'm going to leave it this way, and we can change our minds later. Easy change to make.

LF/Typeclasses.lean
mwhicks1: I think it's already there? See line 462:
```
notation a " →ₜ " b " ; " m => TotalMap.update m a b
```

dsainati1 (approved):

PR #166, opened 2026-07-29
Typeclasses: add notation for getting elements and make PartialMap a one-field structure

xhalo32:
- I added indexing syntax `arr[i]` as `MyGetElem` which needs to be `open scoped MyGetElem` in subsequent chapters that need the syntax. 
- I changed `abbrev PartialMap` to a one-field structure to encapsulate the inner total map. Now there's a full API boundary between the two. To convert a partial map to a total map one should use `toTotal`.

~~Do we want an unexpander for the new notation?~~ Added the unexpander.

~~One unsolved issue is about `ext x : 1`. Have we explained this already somewhere or is this the file to do it?~~


LF/Typeclasses.lean
```diff
@@ -358,14 +358,14 @@ theorem List.elem_poly_eq_elem_nat (xs : List Nat) (n : Nat) : xs.elem_poly n =
...
-We'll define two flavors of maps: total maps, which include a "default" element to be returned when a key being looked up doesn't exist, and partial maps, which instead return an option to indicate success or failure. Partial maps are defined in terms of total maps, using None as the default element.
+We'll define two flavors of maps: total maps, which include a "default" element to be returned when a key being looked up doesn't exist, and partial maps, which instead return an option to indicate success or failure. Partial maps are defined in terms of total maps, using {name}`none` as the default element.
 
 ## Identifiers
 
 To define maps, we first need a type for the keys that we will use to index into our maps. Instead of using concrete types, we will use a type variables
 
 ```lean
-universe u v
+universe u v w
```
chenson2018: This is probably my fault, but I think in the rest of the book we avoid writing universes. It'd be okay for these to all be just `Type` I think.

LF/Typeclasses.lean
```diff
@@ -444,17 +444,66 @@ instance : EmptyCollection (TotalMap α β) where
...
 
-so that we can use the `∅` notation for this empty map. We'll also declare the instance
+so that we can use the `∅` notation for this empty map.
+
+### Getting Elements
+
+To access values in a map, which are functions, we can apply the map to a key:
+
+```lean
+example : (∅ : TotalMap Nat Nat) 1 = default := rfl
```
chenson2018, line 507: I find it confusing that we use function application and then dive into defining `MyGetElem` without explanation. I'd say something like (feel free to wordsmith):

> Here we have made use of the the fact that the type `TotalMap` is defined as simply a function type, so technically we can use function application. While this is possible, it goes against the spirit we have seen in previous chapters of defining interfaces to our types, like characterizing lemmas. We would like that the public interface we design for `TotalMap` is independent of this function type, which should be considered an implementation detail. In the ideal, we should be able to substitute in a different type with similar behavior (a hashmap for instance) and need to change very little of the theorems about `TotalMap`

LF/Typeclasses.lean
```diff
@@ -444,17 +444,66 @@ instance : EmptyCollection (TotalMap α β) where
...
+
+end TotalMap
+```
+
+Let's introduce notation for accessing elements using square brackets.
+We'll call the typeclass `MyGetElem` after {name}`GetElem` from the standard library.
```
chenson2018: If you want to mention the real `GetElem` at all, I'd postpone until the end of the section.

LF/Typeclasses.lean
```diff
@@ -444,17 +444,66 @@ instance : EmptyCollection (TotalMap α β) where
...
+To use the notation `m[a]` to access elements of a map `m`, we add a {name}`MyGetElem` instance for total maps.
+
+```lean
+namespace TotalMap
+variable [Inhabited β]
+
+instance : MyGetElem (TotalMap α β) α β where
+  getElem m a := m a
 
 theorem getElem_def (m : TotalMap α β) (a : α) : m[a] = m a :=
```
chenson2018: If you want to present the `simp` lemma here I think it'd be the reverse (e.g. notation on the right, I always look at [Nat.succ_eq_add_one](https://leanprover-community.github.io/mathlib4_docs/Init/Data/Nat/Basic.html#Nat.succ_eq_add_one) when I forget)

LF/Typeclasses.lean
```diff
@@ -444,17 +444,66 @@ instance : EmptyCollection (TotalMap α β) where
...
   rfl
+
+example : (∅ : TotalMap Nat Nat)[1] = default := rfl
```
chenson2018: I'm wondering if leading with the `default` examples is confusing versus something concrete?

LF/Typeclasses.lean
```diff
@@ -477,24 +526,35 @@ def example_map :=
 We'll also introduce a notation for updating maps
 
 ```lean
-notation a " →ₜ " b " ; " m => TotalMap.update m a b
+notation a:55 " →ₜ " b:55 " ; " m:55 => TotalMap.update m a b
+
+theorem update_def (m : TotalMap α β) (a : α) (b : β) :
+  a →ₜ b ; m = fun a' => bif a == a' then b else m[a'] := rfl
```
chenson2018, line 660: A couple of style points:
- the signature should be indented to four spaces
- I believe that in SFL folks have standardized on using `by rfl`, at least in the first volume where maybe we don't want to draw attention to the term versus the tactic

So this would be:

```suggestion
theorem update_def (m : TotalMap α β) (a : α) (b : β) :
    a →ₜ b ; m = fun a' => bif a == a' then b else m[a'] := by
  rfl
```

LF/Typeclasses.lean
```diff
@@ -679,36 +726,60 @@ instance : Singleton (KVPair α β) (TotalMap α β) where
...
 ## Partial Maps
 
 Lastly, we define _partial maps_ on top of total maps. A partial map with elements of type `β` is simply a total map with elements of type `Option β`, whose default element is {name}`none`.
+We use define partial maps as a one-field structure encapsulating the inner total map.
 
 ```lean
-end TotalMap
+structure PartialMap (α : Type u) (β : Type v) where
+  /-- The inner total map. Should not appear in the public API, use `PartialMap.toTotal` instead. -/
+  private inner : TotalMap α (Option β)
```
chenson2018: I'd remove the `private`. More syntax for the student to learn, and I don't want to consider the meaning versus the module system...

LF/Typeclasses.lean
```diff
@@ -679,36 +726,60 @@ instance : Singleton (KVPair α β) (TotalMap α β) where
...
+
+theorem getElem_def (m : PartialMap α β) (a : α) : m[a] = m.toTotal[a] := rfl
 ```
 
-:::dev "Chris Henson (chenson2018)"
-Making this an `abbrev` is a design decision: a type alias avoids duplicating all
-the typeclass instances (`GetElem`, `EmptyCollection`, ...) for partial maps. If
-this is confusing for any reason, feel free to change.
+:::dev "Niklas Halonen (xhalo32)" NOW
+Explain why we use a one-field structure (instead of a `def` for example)
```
chenson2018: If you adopt my explanation in my early comment about why we define notation at all you could tie back in here.

LF/Typeclasses.lean
```diff
@@ -724,65 +795,126 @@ We have not carried that over — nothing here needs it and it is not clear it i
 wanted. Decide whether to reinstate it.
 :::
 
-Updating a partial map at a key means storing {name}`some` value there, and we introduce a similar notation for it:
+Updating a partial map at a key means storing a {name}`some` value there.
+To update, we create a new partial map from `a →ₜ some b ; m.toTotal` by wrapping it in angle brackets, i.e. using the anonymous constructor syntax.
```
chenson2018, line 963: Again, more syntax, do we really *need* to teach  `⟨⟩` in this chapter? If so, I think the natural place is where the other ways to declare an instance are, since this is a common pattern.

LF/Typeclasses.lean
```diff
@@ -724,65 +795,126 @@ We have not carried that over — nothing here needs it and it is not clear it i
...
+
 @[ext]
-theorem ext (m₁ m₂ : PartialMap α β) (h : ∀ a : α, m₁[a] = m₂[a]) : m₁ = m₂ := funext h
+theorem ext (m₁ m₂ : PartialMap α β) (h : ∀ a : α, m₁[a] = m₂[a]) : m₁ = m₂ := by
+  rw [← toTotal_eq_iff]
+  have heq (a : α) : m₁.toTotal[a] = m₂.toTotal[a] := by
+    specialize h a
+    dsimp [getElem_def] at h
+    exact h
+  rw [TotalMap.ext heq]
```
chenson2018: Any reason to not simplify this proof?
```suggestion
  exact TotalMap.ext h
```

LF/Typeclasses.lean
```diff
@@ -724,65 +795,126 @@ We have not carried that over — nothing here needs it and it is not clear it i
...
     (a₁ →ₚ b₁ ; a₂ →ₚ b₂ ; m) = (a₂ →ₚ b₂ ; a₁ →ₚ b₁ ; m) := by
-  simp only [totalMap_eq]
-  exact TotalMap.update_permute m a₁ a₂ (some b₁) (some b₂) h
+  ext x : 1
+  dsimp [getElem_def, toTotal_update]
+  rw [TotalMap.update_permute h]
+```
+
+:::dev "Niklas Halonen (xhalo32)" NOW
+`ext x : 1` has to be explained somewhere.
```
chenson2018: The usages above can just `ext`. Even if it unfolds more than you want I'd just use the lemma explicitly, I'm not sure it's worth introducing since nowhere else uses this.

LF/Typeclasses.lean
```diff
@@ -444,17 +444,66 @@ instance : EmptyCollection (TotalMap α β) where
...
+Let's introduce notation for accessing elements using square brackets.
+We'll call the typeclass `MyGetElem` after {name}`GetElem` from the standard library.
+The reason we are defining a new typeclass is for simplicity -- {name}`GetElem` contains logic for when keys are valid which we don't need.
+
+The typeclass takes three type parameters: the collection, the keys that can be used to index and the elements that it returns.
+For example, let's say we have declared an instance `MyGetElem (List Bool) Nat Bool`.
+One can think of the instance as saying that _indexing into `xs : List Bool` with an index `i : Nat` returns a `Bool`_.
+The notation we introduce for this is: `xs[i] = MyGetElem.getElem xs i`.
+
+Don't worry about the `outParam`, `macro_rules` or the `app_unexpander`, they are minor technicalities.
```
chenson2018: I think for the `outParam` you have to at least do some hand-waving explanation, something like: "it's like a normal paramater with a hint to Lean that helps typeclass inference" .

(It's fine to sweep the unexpander under the rug though)

chenson2018 (commented):
The core changes here look good, these are mostly picky comments about the balance between how someone more experienced would write things versus limiting how many new concepts and syntaxes we introduce while teaching a beginner.

LF/Typeclasses.lean
xhalo32: The way I've learned it is that `_def` lemmas should basically never be `@[simp]` (some examples that aligns with this: [id_def](https://leanprover-community.github.io/mathlib4_docs/Init/Core.html#id_def), [Function.comp_def](https://leanprover-community.github.io/mathlib4_docs/Init/Core.html#Function.comp_def), [max_def](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/Defs/LinearOrder.html#max_def)). I haven't heard before that the reverse direction should be marked as `@[simp]`.

For this particular case I think we can leave it as is (and never mark `getElem_def` as `@[simp]`)?

LF/Typeclasses.lean
xhalo32: This mirrors `example : (∅ : TotalMap Nat Nat) 1 = default := rfl` (line 453) that was there already before. I'll leave a dev-note and let's come back to this later, if that's okay?

LF/Typeclasses.lean
chenson2018: Yes, this is all true, but not my question. I meant show the version that *should* be `simp` and comment on this. At a high level, I think we're missing an explanation of *why* we are doing this odd thing of hiding function application being a notation.

LF/Typeclasses.lean
xhalo32: I think the `rfl` versus `by rfl` is not so consistent (already starting with `theorem even_zero` in Basics). I'll make an issue to standardize this across all chapters.

xhalo32: https://github.com/plclub/sf-in-lean/issues/180

LF/Typeclasses.lean
xhalo32: I don't get the point why we would want the reverse direction to be `simp` because I don't see a need for it in the proofs. None of the standard library examples do this either. The `TotalMap.getElem_def` is only used in a couple of places, notably in `update_eq` and `update_neq`.

I do think that exposing `getElem_def` as a public theorem goes against the philosophy that we are trying to explain (as you have written [in another comment](https://github.com/plclub/sf-in-lean/pull/166#discussion_r3686516366)). Maybe we should just delete `getElem_def` and use `rfl` in the proofs of `update_eq/neq`? It doesn't count as defeq abuse because `update_eq/neq` are characterizing lemmas.

xhalo32: If we get rid of `getElem_def` completely, the proofs need to use e.g. `change`:
```
theorem update_eq (m : TotalMap α β) (a : α) (b : β) : (a →ₜ b ; m)[a] = b := by
  rw [update_def]
  change (bif a == a then b else m[a]) = b
  rw [ReflBEq.rfl, cond_true]
```
Can we make `getElem_def` private somehow?

chenson2018: No, I disagree. Let's explicitly provide both theorems and explain when they should be used, relating this to the prior material about characterizing lemmas.

`getElem_def`, which rewrites from notation into the plain function application, is sparingly used to open up the API barrier for these lemmas `update_(n)eq` that characterize its behavior.

The reverse direction, function application into notation, the `simp` direction, is there to ensure that if we somehow end up not in the notation (induction eliminators, building API etc.) that we get back to our abstraction. It may not be used here, but it's a one line theorem that I think really helps the explanation.

(If we reach a point here where we agree but you feel more comfortable with me adding this explanation later let me know, as this is important but I don't want to hold up other chapters)

LF/Typeclasses.lean
chenson2018: Yes, that's fine.

LF/Typeclasses.lean
xhalo32: Anonymous constructor syntax is introduced already in Lists.lean. Do you still want us to use `{ inner := ... }` instead of `⟨...⟩`? It's fine to me either way.

LF/Typeclasses.lean
xhalo32: Nice catch! I added `rw [← toTotal_eq_iff]` after I already had the `heq` style proof and didn't realize it was more straight-forward...

LF/Typeclasses.lean
xhalo32: But I think it's confusing that it uses Option.ext. `ext x a'` turns the goal into `(a →ₚ b₂ ; a →ₚ b₁ ; m)[x] = some a' ↔ (a →ₚ b₂ ; m)[x] = some a'` which is more complex than it needs to be.

What if we use
```
apply ext
intro x
```
instead of `ext x : 1`?

LF/Typeclasses.lean
xhalo32: Okay, let's do that. I can add the explanation and the new `@[simp]` lemma now (what should we call it). I can see your point that normalizing back to the notation can make simp "more useful" in some cases. I'm surprised I can't find anything concrete about this in the style guidelines or the lean reference...

LF/Typeclasses.lean
chenson2018: Hmm, I missed this being added, sorry! If the usage is light enough that I forgot, a student definitely will too.

So this is fine, but let's add a mention of it at the "First Example" section where I showed the different ways to define a structure.

LF/Typeclasses.lean
xhalo32: There's some kind of a quirk with rewriting function applications:
```
theorem getElem_def (m : TotalMap α β) (a : α) : m[a] = m a := by rfl

@[simp]
theorem apply_eq_getElem (m : TotalMap α β) (a : α) : m a = m[a] := by rfl

example (m : TotalMap α β) (a : α) : m a = m[a] := by
  -- simp -- doesn't work (infinite loop)
  -- dsimp -- doesn't work (nothing happens)
  -- rw [apply_eq_getElem] -- doesn't work: The pattern to be substituted is a metavariable (`?m ?a`) in this equality: ?m ?a = ?m[?a]
  rw [apply_eq_getElem m a] -- works
```
I don't know how to fix this to demonstrate how `apply_eq_getElem` might be used.

LF/Typeclasses.lean
chenson2018: This is what I meant by "use the lemma explicitly" but on second thought this is not great.

Let's leave what you had, but try to resolve this dev note in a follow up PR since it's an easy one. (I'd put an explanation both here and where `ext` is originally introduced)

LF/Typeclasses.lean
chenson2018: I'm not at a computer I can test with, for now I'd leave off the example and leave a dev note.

LF/Typeclasses.lean
xhalo32: I took your explanation and augmented it a bit.

LF/Typeclasses.lean
```diff
@@ -444,17 +445,107 @@ instance : EmptyCollection (TotalMap α β) where
...
+@[simp]
+theorem apply_eq_getElem (m : TotalMap α β) (a : α) : m a = m[a] := by rfl
 ```
 
-which allows us to use the notation `m[a]` to access elements of a map `m`.
+The reverse direction of {name}`getElem_def` is provided as a `simp`-lemma.
+
+```
+example (m : TotalMap α β) (a : α) : m a = m[a] := by
+  -- simp -- doesn't work (infinite loop)
```
berberman, line 600: Should we explain why it would be an infinite loop?

LF/Typeclasses.lean
```diff
@@ -724,65 +856,121 @@ We have not carried that over — nothing here needs it and it is not clear it i
...
+To prove extensionality, we employ injectivity of {name}`PartialMap`'s constructor {name}`mk` using {name}`mk.injEq`.
+
+```lean
+theorem toTotal_eq_iff (m₁ m₂ : PartialMap α β) : m₁.toTotal = m₂.toTotal ↔ m₁ = m₂ := by
+  rw [mk.injEq]
+  rfl
+
 @[ext]
-theorem ext (m₁ m₂ : PartialMap α β) (h : ∀ a : α, m₁[a] = m₂[a]) : m₁ = m₂ := funext h
+theorem ext (m₁ m₂ : PartialMap α β) (h : ∀ a : α, m₁[a] = m₂[a]) : m₁ = m₂ := by
```
berberman: Should the first two arguments be implicit?

LF/Typeclasses.lean
```diff
@@ -444,17 +445,107 @@ instance : EmptyCollection (TotalMap α β) where
...
+
+We'll call the typeclass `MyGetElem`.
+The typeclass takes three type parameters: the collection, the keys that can be used to index and the elements that it returns.
+For example, let's say we have declared an instance `MyGetElem (List Bool) Nat Bool`.
+One can think of the instance as saying that _indexing into `xs : List Bool` with an index `i : Nat` returns a `Bool`_.
+The notation we introduce for this is: `xs[i] = MyGetElem.getElem xs i`.
+
+:::instructors
+`MyGetElem` is named after {name}`GetElem` from the standard library.
+The reason we are defining a new typeclass is for simplicity -- {name}`GetElem` contains logic for when keys are valid which we don't need.
```
berberman:
```suggestion
The reason we are defining a new typeclass is for simplicity — {name}`GetElem` contains logic for when keys are valid which we don't need.
```

LF/Typeclasses.lean
```diff
@@ -444,17 +445,107 @@ instance : EmptyCollection (TotalMap α β) where
...
+:::
+
+Don't worry about the `outParam`, it's like a normal paramater with a hint to Lean that helps typeclass inference.
```
berberman: This sounds like outParam is a normal parameter.

LF/Typeclasses.lean
xhalo32: I have no idea why that happens so please explain. Notice that this is inside a dev comment

LF/Typeclasses.lean
xhalo32: It is a normal type parameter from the kernel's point of view, no?

LF/Typeclasses.lean
xhalo32: Yes, feel free to change, not at the computer at the moment.

xhalo32: I fixed this

LF/Typeclasses.lean
berberman: Yes, I mean to be precise we might want to say something like the parameter (elem) annotated with outParam works like normal parameter, not outParam itself. Sorry for the confusion.

xhalo32: Oh I see, thanks for clarifying! I updated it.

LF/Typeclasses.lean
berberman: Oh I missed the earlier discussion and your comment is in dev block. To see why, with `trace.Meta.Tactic.simp.rewrite` enabled and specifying `maxSteps` to `simp`, we can track what's going on:
```lean
[Meta.Tactic.simp.rewrite] apply_eq_getElem:1000:
      m a
    ==>
      m[a] 
[Meta.Tactic.simp.rewrite] apply_eq_getElem:1000:
      m[a]
    ==>
      (MyGetElem.getElem m)[a] 
[Meta.Tactic.simp.rewrite] apply_eq_getElem:1000:
      (MyGetElem.getElem m)[a]
    ==>
      (MyGetElem.getElem (MyGetElem.getElem m))[a] 
[Meta.Tactic.simp.rewrite] apply_eq_getElem:1000:
      (MyGetElem.getElem (MyGetElem.getElem m))[a]
    ==>
      (MyGetElem.getElem (MyGetElem.getElem (MyGetElem.getElem m)))[a] 
[Meta.Tactic.simp.rewrite] apply_eq_getElem:1000:
      (MyGetElem.getElem (MyGetElem.getElem (MyGetElem.getElem m)))[a]
    ==>...
```
So I think the issue here is that the LHS `m a` is too board, causing this lemma repeating itself. First we have `m a => m[a]`, but `m[a]` is defeq to `(MyGetElem.getElem m) a`, and nevertheless we require `m` to be a `TotalMap`, a `TotalMap` is  defeq to `α → β`. So this lemma applies again -- `(MyGetElem.getElem m) a => (MyGetElem.getElem m)[a]` which is `(MyGetElem.getElem (MyGetElem.getElem m))[a] ` and keep going.

xhalo32:
> If this way is better, then why didn't we do it for total maps too?  Just for the sake
of explaining two different mechanisms? We should explain our reasoning.

@bcpierce00 with `TotalMap` we introduced the `MyGetElem` notation for accessing elements (including, as you noted, a rather heavy explanation why we did it), and with `PartialMap` we introduced the one-field structure -- both introduce one new big thing. Do you think it would make sense to introduce both the notation and the encapsulation using a structure at the same time for `TotalMap`?

For me it isn't clear how we want to motivate encapsulation and its philosophy. Maybe @chenson2018 has more to say about this?

chenson2018:
I'm having trouble with the GitHub UI finding the context for this question. Is it why we use `def` in one place and a one-field structure in another? Or something else?

xhalo32:
@chenson2018  Here's the link to the [dev comment](https://github.com/plclub/sf-in-lean/pull/166/changes/dc016daee63d50be1376c5b0c38f954f0d486e75#diff-51fc7bc3b8e30cc2a9571355b707d432dcc6dd94ec79c99b402ad7511aa94680R860). Yes, I believe it's about that. I don't have an explanation and don't know if/why we should prefer `def` over `structure` for `TotalMap`.

chenson2018:
> @chenson2018 Here's the link to the [dev comment](https://github.com/plclub/sf-in-lean/pull/166/changes/dc016daee63d50be1376c5b0c38f954f0d486e75#diff-51fc7bc3b8e30cc2a9571355b707d432dcc6dd94ec79c99b402ad7511aa94680R860). Yes, I believe it's about that. I don't have an explanation and don't know if/why we should prefer `def` over `structure` for `TotalMap`.

I see. My response would be that the structure is stricter mechanism that would be painful to use everywhere. I thought it was fairly natural to show the progression, but agree there could be more explanation. (I've not looked again since my review to see what's different though)

LF/Typeclasses.lean
mwhicks1: Students will run across universes, though. When looking at `List` lemmas, for example, they will see things like: 
```
List.reverse.{u} {α : Type u} (as : List α) : List α
```
Are we explaining these things somewhere, maybe in `Poly` ? @dsainati1

chenson2018: I think that both of these are true:
- We have been (I thought intentionally) writing just `Type`, because there's not much value to universes for this material. Grepping the codebase this seems to be the single place, and it's because I did so out of habit accidentally.
- Some definitions in core Lean will print in the infoview with universes, like your example

To my recollection, the Rocq version almost completely avoided mentioning universes. I'd probably keep it similarly lightweight, taking your suggestion to mention it in `Poly` the first time it is seen, but not going much further than the Rocq explanation of it being a "way of preventing a paradox in which a type contains itself" and how this will display. (I don't see any such explanation currently)

LF/Typeclasses.lean
xhalo32: I made it an intructor note only.

LF/Typeclasses.lean
```diff
@@ -444,17 +477,126 @@ instance : EmptyCollection (TotalMap α β) where
...
+The reverse direction of {name}`getElem_def` is provided as a `simp`-lemma.
+
+```
+example (m : TotalMap α β) (a : α) : m a = m[a] := by
+  -- simp -- doesn't work (infinite loop)
+  -- dsimp -- doesn't work (nothing happens)
+  -- rw [apply_eq_getElem] -- doesn't work: The pattern to be substituted is a metavariable (`?m ?a`) in this equality: ?m ?a = ?m[?a]
+  rw [apply_eq_getElem m a] -- works
+```
+:::
```
xhalo32, line 632:
```suggestion
berberman (explanation of the infinite loop): To see why, with trace.Meta.Tactic.simp.rewrite enabled and specifying maxSteps to simp, we can track what's going on:

[Meta.Tactic.simp.rewrite] apply_eq_getElem:1000:
      m a
    ==>
      m[a] 
[Meta.Tactic.simp.rewrite] apply_eq_getElem:1000:
      m[a]
    ==>
      (MyGetElem.getElem m)[a] 
[Meta.Tactic.simp.rewrite] apply_eq_getElem:1000:
      (MyGetElem.getElem m)[a]
    ==>
      (MyGetElem.getElem (MyGetElem.getElem m))[a] 
[Meta.Tactic.simp.rewrite] apply_eq_getElem:1000:
      (MyGetElem.getElem (MyGetElem.getElem m))[a]
    ==>
      (MyGetElem.getElem (MyGetElem.getElem (MyGetElem.getElem m)))[a] 
[Meta.Tactic.simp.rewrite] apply_eq_getElem:1000:
      (MyGetElem.getElem (MyGetElem.getElem (MyGetElem.getElem m)))[a]
    ==>...

So I think the issue here is that the LHS m a is too board, causing this lemma repeating itself. First we have m a => m[a], but m[a] is defeq to (MyGetElem.getElem m) a, and nevertheless we require m to be a TotalMap, a TotalMap is defeq to α → β. So this lemma applies again -- (MyGetElem.getElem m) a => (MyGetElem.getElem m)[a] which is (MyGetElem.getElem (MyGetElem.getElem m))[a] and keep going.
:::
```

mwhicks1:
I have read over the whole thing and I think it still needs work, mostly on the narrative text, but potentially also to the formal development, as it's quite long. But it's better to get it merged and do that work as a follow-up PR.

mwhicks1 (approved):

PR #167, opened 2026-07-29
Pass over Logic chapter

dsainati1:


LF/Logic.lean
```diff
@@ -132,9 +122,9 @@ For example, we can name them:
 ::::
 
 ```lean
-def plus_claim : Prop := 2 + 2 = 4
+def PlusClaim : Prop := 2 + 2 = 4
 
-#check (plus_claim : Prop)
+#check (PlusClaim : Prop)
```
xhalo32: What's the role of the ascription `: Prop` here?

This occurs multiple times but I won't point out all of them. If we decide to remove the ascriptions, let's remove them from every `#check`

Should we also add `leanOutput` boxes below the `#check`s?

LF/Logic.lean
```diff
@@ -156,9 +146,9 @@ returns a proposition asserting that this number is equal to three:
 ::::
 
 ```lean
-def is_three (n : Nat) : Prop := n = 3
+def IsThree (n : Nat) : Prop := n = 3
 
-#check (is_three : Nat → Prop)
+#check (IsThree : Nat → Prop)
```
xhalo32: Same question about ascription here

LF/Logic.lean
```diff
@@ -212,25 +207,27 @@ For clarity, we will avoid relying on these implicit casts.
...
 2. `Nat → Prop`
 3. `∀ n : Nat, Prop`
 4. `Nat → Nat`
 5. Not typeable
-::::
 
+:::quizSolution
 ```lean
-#check (pred (succ zero) = zero : Prop)
+#check (Nat.pred 1 = 0 : Prop)
```
xhalo32: Same question about ascription...

LF/Logic.lean
```diff
@@ -387,7 +395,7 @@ theorem add_is_zero (n m : Nat) : n + m = 0 → n = 0 ∧ m = 0 := by
 
 So much for proving conjunctive statements.  To go in the other
 direction -- i.e., to _use_ a conjunctive hypothesis to help prove
-something else -- we can use `let` to obtain the components.
+something else -- we can use {tactic}`let` to obtain the components.
```
xhalo32: Isn't `obtain` more idiomatic for destructuring?

LF/Logic.lean
```diff
@@ -168,15 +158,15 @@ For instance, here's a (polymorphic) property defining the
 familiar notion of an _injective function_.
 
 ```lean
-def injective {α β} (f : α → β) : Prop :=
+def Injective {α β} (f : α → β) : Prop :=
```
berberman:
```suggestion
def Injective {α β : Type} (f : α → β) : Prop :=
```

xhalo32 (commented):
Looks good, I reviewed up to line 700. My only comments are about old things that you didn't change.

LF/Logic.lean
berberman: Should we write `#check (IsThree)` followed by `leanOutput`? P.S. parentheses are required to print it in arrow.

LF/Logic.lean
berberman: Maybe drop these `#guard_msgs`?

LF/Logic.lean
```diff
@@ -212,25 +207,27 @@ For clarity, we will avoid relying on these implicit casts.
...
 
 ```display
-pred (succ zero) = zero
+Nat.pred 1 = 0
 ```
 
 1. `Prop`
 2. `Nat → Prop`
 3. `∀ n : Nat, Prop`
 4. `Nat → Nat`
```
berberman:
```suggestion
1. {lean}`Prop`
2. {lean}`Nat → Prop`
3. {lean}`∀ n : Nat, Prop`
4. {lean}`Nat → Nat`
```

LF/Logic.lean
```diff
@@ -310,20 +310,28 @@ Which of the following is _not_ a proposition?
 4. `(3 + 2 == 4) = false`
```
berberman:
```suggestion
1. {lean}`3 + 2 = 4`
2. {leam}`3 + 2 = 5`
3. {lean}`3 + 2 == 5`
4. {lean}`(3 + 2 == 4) = false`
```

LF/Logic.lean
```diff
@@ -256,50 +253,53 @@ What is the type of the following expression?
...
 
 ```display
-fun n : Nat => succ (pred n) = n
+fun n : Nat => n.pred + 1 = n
 ```
 
 1. `Prop`
 2. `Nat → Prop`
 3. `∀ n : Nat, Prop`
 4. `Nat → Nat`
```
berberman:
```suggestion
1. {lean}`Prop`
2. {lean}`Nat → Prop`
3. {lean}`∀ n : Nat, Prop`
4. {lean}`Nat → Nat`
```

LF/Logic.lean
```diff
@@ -310,20 +310,28 @@ Which of the following is _not_ a proposition?
...
 
 # Logical Connectives
 
 ## Conjunction
 
-The _conjunction_, or _logical and_, of propositions `A` and `B` is written
-`A ∧ B`; it represents the claim that both `A` and `B` are true.
+:::ignore
+```lean -show
+variable (a b c : Prop) (n m : Nat)
```
berberman: ~~I think `{p q : Prop}` would be better for props~~

LF/Logic.lean
```diff
@@ -963,69 +977,69 @@ example : True := ⟨⟩
...
 ```lean
 -- SOLUTION
-def is_nil {α : Type} (xs : List α) : Prop :=
+def IsNil {α : Type} (xs : List α) : Prop :=
   match xs with
   | [] => True
   | _ :: _ => False
 
-theorem is_nil_nil {α} : @is_nil α [] = True := rfl
+theorem IsNil_nil {α} : @IsNil α [] = True := rfl
```
berberman: something like
```lean
theorem IsNil_nil {α : Type} : IsNil ([] : List α) = True := rfl
```
or more idiomatically:
```lean
theorem IsNil_nil {α : Type} : IsNil ([] : List α) := trivial
```

LF/Logic.lean
xhalo32: You changed the formatting of the `quizSolution`. Did you forget to change this?

LF/Logic.lean
```diff
@@ -241,7 +238,7 @@ What is the type of the following expression?
 ::::
 
 ```lean
```
xhalo32: Is this missing `:::quizSolution`?

LF/Logic.lean
```diff
@@ -256,50 +253,53 @@ What is the type of the following expression?
...
 2. `Nat → Prop`
 3. `∀ n : Nat, Prop`
 4. `Nat → Nat`
 5. Not typeable
-::::
 
+:::quizSolution
 ```lean
-#check (fun n : Nat => succ (pred n) : Nat → Nat)
+#check (fun n : Nat => (n.pred) + 1 : Nat → Nat)
```
xhalo32: Unnecessary parens around n.pred

LF/Logic.lean
```diff
@@ -963,69 +977,69 @@ example : True := ⟨⟩
...
+The {tactic}`contradiction` tactic takes care of all of this for us.
 
 :::::exercise (rating := 2) (name := "nil_is_not_cons") (level := Advanced) (manual := true)
 Use the same technique as above to show that `[] ≠ x :: xs`.
 Do not use the `contradiction` tactic.
 
 ```lean
 -- SOLUTION
-def is_nil {α : Type} (xs : List α) : Prop :=
+def IsNil {α : Type} (xs : List α) : Prop :=
```
berberman: I would put it in `List` namespace.

LF/Logic.lean
```diff
@@ -963,69 +977,69 @@ example : True := ⟨⟩
...
   match xs with
   | [] => True
   | _ :: _ => False
 
-theorem is_nil_nil {α} : @is_nil α [] = True := rfl
+theorem IsNil_nil {α} : @IsNil α [] = True := rfl
 
-theorem is_nil_cons {α} (x : α) (xs : List α) : is_nil (x :: xs) = False := rfl
+theorem IsNil_cons {α} (x : α) (xs : List α) : IsNil (x :: xs) = False := rfl
```
berberman: also:
```lean
theorem IsNil_cons {α : Type} {x : α} {xs : List α} : IsNil (x :: xs) = False := rfl
```

Or using `Not` if we've introuduced it:

```lean
theorem isNil_cons {α : Type} {x : α} {xs : List α} : ¬ IsNil (x :: xs) := by
  dsimp [IsNil, Not]
  trivial
```

LF/Logic.lean
```diff
@@ -963,69 +977,69 @@ example : True := ⟨⟩
...
+def DiscrFun (n : Nat) : Prop :=
   match n with
   | 0 => True
   | _ + 1 => False
 
-theorem discr_fun_zero : discr_fun 0 = True := rfl
+theorem DiscrFun_zero : DiscrFun 0 = True := rfl
 
-theorem discr_fun_succ n : discr_fun (n + 1) = False := rfl
+theorem DiscrFun_succ n : DiscrFun (n + 1) = False := rfl
```
berberman:
```suggestion
theorem DiscrFun_succ {n : Nat} : DiscrFun (n + 1) = False := rfl
```

LF/Logic.lean
```diff
@@ -1628,55 +1673,59 @@ Suppose we have
 
 ```display
 n m : Nat
-H1 : n = m
-H2 : b = 42
-trans_eq : ∀ (α : Type) (a b c : α), a = b → b = c → a = c
+h₁ : n = m
+h₂ : b = 42
+trans_eq : ∀ (α : Type) (x y z : α), x = y → y = z → x = z
```
berberman: `α` and `x y z` should be implicit to make it easier to use. IIRC it's defined in Tactics. We'll need to upate there.

STYLE.md
```diff
@@ -124,6 +124,15 @@ Follow the Lean library's naming conventions:
 - Types and propositions (including definitions returning `Prop`) use `UpperCamelCase`, e.g. `Aexp`, `IsValue`;
 - Other values and functions use `lowerCamelCase`, e.g. `isEven`, `doubleBin`.
 
+Also follow Lean's variable naming conventions:
+
+- Types are Greek letters like `α`, `β`
+- Propositions are lowercase `a`, `b`, `c`, etc
```
berberman, line 130: ~~I think `p`, `q`, `r`, etc. are more conventional. Use `a`, `b`, `c`, etc. for arbitrary values e.g. `Type`~~

LF/Logic.lean
xhalo32: Or `#check IsThree`?

LF/Logic.lean
xhalo32: The [style guide](https://leanprover-community.github.io/contribute/style.html) says that p, q are for predicates and a, b, c are for propositions.

LF/Logic.lean
berberman: `#check IsThree` would print `IsThree (n : Nat) : Prop` instead of `Nat → Prop`. Maybe I should document this trick... 
https://github.com/plclub/sf-in-lean/blob/b302ac56f008f9126acd538a2e9545dadf9e7a41/LF/Poly.lean#L130-L132

LF/Logic.lean
xhalo32: The we should put IsThree in Nat right?

LF/Logic.lean
xhalo32: I think the latter is better. I highly dislike `= False` and proving that with `rfl` which is defeq abuse to some extent.

This comment applies more widely to all instances of `= False`

LF/Logic.lean
xhalo32: In my opinion we should do the explicit to implicit refactoring in a separate pass rather than in this PR.

STYLE.md
xhalo32: As I noted earlier a, b, c are idiomatic according to mathlib style guidelines.

LF/Logic.lean
dsainati1: Changing this to a Lean block doesn't typecheck unfortunately. This is the sort of thing it would be nice to support as discussed in [#lean-software-foundations-contributors > Syntax Highlighting without Typechecking](https://leanprover.zulipchat.com/#narrow/channel/607217-lean-software-foundations-contributors/topic/Syntax.20Highlighting.20without.20Typechecking/with/613014764)

LF/Logic.lean
berberman: Thanks, I was confused.

dsainati1: I saw this and was confused, because if you look at things like `And.intro` or `Iff.intro` they use `a` and `b` for their `Prop`s.

LF/Logic.lean
berberman: I agree, but I guess we could put a comment here

LF/Logic.lean
dsainati1: We haven't taught `trivial` yet, that shows up in the Automation chapter. Will change to the first one though.

LF/Logic.lean
xhalo32: I was unclear: I meant that you changed the formatting of the lean code in the quiz solution, but forgot to change `succ (pred n)` in the quiz question. You seem to have fixed this since so all good.

LF/Logic.lean
berberman: How about `True.intro`, `⟨⟩`, or `constructor`?

LF/Logic.lean
```diff
@@ -150,13 +150,13 @@ returns a proposition asserting that this number is equal to three:
 ::::
 
 ```lean (name := IsThree)
-def IsThree (n : Nat) : Prop := n = 3
+def Nat.IsThree (n : Nat) : Prop := n = 3
 
-#check IsThree
+#check (Nat.IsThree)
```
xhalo32, line 151: I think the parens can be removed.

dsainati1: I was following @berberman's convention in https://github.com/plclub/sf-in-lean/pull/167#discussion_r3676967000 which IMO looks nicer

LF/Logic.lean
```diff
@@ -1036,14 +1036,16 @@ def DiscrFun (n : Nat) : Prop :=
...
 
-theorem DiscrFun_succ n : DiscrFun (n + 1) = False := rfl
+theorem DiscrFun_succ (n : Nat) : ¬ DiscrFun (n + 1) := by
+  dsimp [DiscrFun]; intro h; assumption
 
 theorem discr_example (n : Nat) : ¬ (0 = n + 1) := by
   intro h
-  have hd : DiscrFun 0 := by rw [DiscrFun_zero]; exact ⟨⟩
-  rw [h, DiscrFun_succ] at hd; exact hd
+  have hd : DiscrFun 0 := by exact DiscrFun_zero
```
xhalo32: Just `:= DiscrFun_zero` suffices, no need for by exact

LF/Logic.lean
xhalo32: Oh I see, I didn't know about that before! I totally agree it's better

STYLE.md
```diff
@@ -124,6 +124,16 @@ Follow the Lean library's naming conventions:
 - Types and propositions (including definitions returning `Prop`) use `UpperCamelCase`, e.g. `Aexp`, `IsValue`;
 - Other values and functions use `lowerCamelCase`, e.g. `isEven`, `doubleBin`.
 
+Also follow Lean's variable naming conventions:
```
dsainati1, line 127: Updated style guide to make explicit our naming conventions:

LF/Logic.lean
```diff
@@ -1035,7 +1077,25 @@ theorem nil_is_not_cons {α : Type} (x : α) (xs : List α) :
 
 The handy "if and only if" connective, which asserts that two
 propositions have the same truth value, is a structure containing
-the two implication directions. `P ↔ Q` is notation for `Iff P Q`.
+the two implication directions. {lean}`a ↔ b` is notation for {lean}`Iff a b`.
+
+::::full
+In Lean, {lean}`Iff` is a structure packaging two fields and a constructor, which allow
```
dsainati1, line 1112: Some new text about how `Iff` works in Lean.

LF/Logic.lean
```diff
@@ -2539,42 +2614,42 @@ theorem tr_rev_correct {α} : @tr_rev α = @List.rev α := by
 
 ::::full
 We have seen that it is not possible to test whether or not a
-proposition `P` holds while defining a Lean function. You may be
+proposition {lean}`a` holds while defining a Lean function. You may be
 surprised to learn that a similar restriction applies in _proofs_!
 In other words, the following intuitive reasoning principle is not
-derivable in Lean:
+derivable in Lean with the tools we've seen so far:
```
dsainati1, line 2770: Moderated text here slightly, since Lean is classical by default unlike Rocq.

LF/Logic.lean
```diff
@@ -168,19 +166,19 @@ For instance, here's a (polymorphic) property defining the
...
 
-theorem succ_inj' : injective succ := by
+theorem succ_inj' : Injective Nat.succ := by
   intro x y H; injection H
 ```
 
 The familiar equality operator `=` is a (binary) function that returns
-a `Prop`. The expression `n = m` is notation for `Eq n m`.
+a {lean}`Prop`. The expression `n = m` is notation for `Eq n m`.
 Because `eq` can be used with elements of any type, it is also
```
KihongHeo: Does it mean `Eq` rather than `eq`?

LF/Logic.lean
```diff
@@ -189,141 +187,184 @@ right at this moment, but they'll see `Sort` when hovering.
 ```lean
 #check (Eq : ∀ {α : Type}, α → α → Prop)
 
-#check pred
+#check Nat.pred
```
KihongHeo: I don't understand why `Nat.pred` suddenly appears here.

LF/Logic.lean
KihongHeo: How about using `obtain` as we introduced `obtain` just right before?

LF/Logic.lean
```diff
@@ -441,13 +482,13 @@ example (n m : Nat) (h : n + m = 0) : n * m = 0 := by
...
-context we need just `A` or just `B`.  In such cases we can use
+Another common situation is that we know {lean}`a ∧ b` but in some
+context we need just {lean}`a` or just {lean}`b`.  In such cases we can use
 an underscore pattern `_` to indicate that the unneeded conjunct
 should just be thrown away.
 
 ```lean
-theorem proj1 (P Q : Prop) (h : P ∧ Q) : P := by
+theorem proj1 (a b : Prop) (h : a ∧ b) : a := by
   let ⟨hP, _⟩ := h
```
KihongHeo: Similarly, `obtain`?

LF/Logic.lean
```diff
@@ -517,35 +558,35 @@ propositions as arguments and yields a proposition.
...
 two proof obligations, each showing that the conclusion holds
-under a different assumption -- `P` in the first subgoal and `Q`
+under a different assumption - {lean}`a` in the first subgoal and {lean}`b`
 in the second.
 ::::
 
-Rather than performing case analysis via `cases`, we can also use `obtain`
-to match on the two possible injections, much like with `let`.
+Rather than performing case analysis via {tactic}`cases`, we can also use {tactic}`obtain`
+to match on the two possible injections, much like with {tactic}`let` and `∧`.
```
KihongHeo: let -> obtain?

LF/Logic.lean
```diff
@@ -1304,65 +1371,67 @@ Fundamental connectives we've been using since the beginning:
 ::::full
 The logical connectives that we have seen provide a rich vocabulary
 for defining complex propositions from simpler ones.
-To illustrate, let's look at how to express teh claim that an element `x`
-occurs in a list `l`.
+To illustrate, let's look at how to express teh claim that an element {lean}`x`
```
KihongHeo: teh -> the

LF/Logic.lean
```diff
@@ -1304,65 +1371,67 @@ Fundamental connectives we've been using since the beginning:
...
     /- (Notice the use of the empty pattern to discharge the last case.) -/
 ```
 
-We can also reason about more generic statements involving `In`.
+We can also reason about more generic statements involving {lean}`List.In`.
 
 ```lean
-theorem In_map (α β : Type) (f : α → β) (xs : List α) (x : α) (h : In x xs) :
-    In (f x) (List.map f xs) := by
+theorem In_map (α β : Type) (f : α → β) (xs : List α) (x : α) (h : List.In x xs) :
```
KihongHeo: `In_map` or `List.In_map`?

LF/Logic.lean
```diff
@@ -963,69 +998,74 @@ example : True := ⟨⟩
...
 
 ```lean
-def discr_fun (n : Nat) : Prop :=
+def DiscrFun (n : Nat) : Prop :=
   match n with
   | 0 => True
   | _ + 1 => False
 
-theorem discr_fun_zero : discr_fun 0 = True := rfl
+theorem DiscrFun_zero : DiscrFun 0 := by constructor
```
xhalo32:
```suggestion
theorem discrFun_zero : DiscrFun 0 := by constructor
```
I think theorem names should always start with lowercase. This comment applies to many other cases, like `DiscrFun_succ` below

LF/Logic.lean
```diff
@@ -1051,7 +1109,7 @@ constructor:
 
 #check (fun α β : Prop => α ↔ β : Prop → Prop → Prop)
 
-theorem iff_sym (P Q : Prop) (h : P ↔ Q) : (Q ↔ P) := by
+theorem iff_sym (a b : Prop) (h : a ↔ b) : (b ↔ a) := by
```
xhalo32:
```suggestion
theorem iff_sym (a b : Prop) (h : a ↔ b) : b ↔ a := by
```
There are more instances with unnecessary parentheses below

LF/Logic.lean
```diff
@@ -1181,12 +1247,12 @@ example n : (∃ m, n = m + 4) → (∃ o, n = o + 2) := by
 
 ::::::full
 :::::exercise (rating := 1) (name := "dist_not_exists")
-Prove that "`P` holds for all `x` implies "there is no `x` for which
-`P` does not hold." (Hint: `cases` works on existential assumptions!)
+Prove that "{lean}`a` holds for all {lean}`x` implies "there is no {lean}`x` for which
+{lean}`a` does not hold." (Hint: `cases` and `let` work on existential assumptions!)
```
xhalo32:
```suggestion
{lean}`a` does not hold." (Hint: `cases` and `obtain` work on existential assumptions!)
```

LF/Logic.lean
```diff
@@ -1410,35 +1479,35 @@ theorem In_map_iff (α β : Type) (f : α → β) (xs : List α) (y : β) :
...
-_properties_ of their arguments. For instance, if `P` has type
-`Nat -> Prop`, then `P n` says that property `P` holds of `n`.
+_properties_ of their arguments. For instance, if `p` has type
+{lean}`Nat -> Prop`, then `p n` says that property `p` holds of {lean}`n`.
 
-Drawing inspiration from `In`, write a recursive function `All`
-stating that some property `P` holds of all elements of a list
-`xs`. To make sure your definition is correct, prove the `All_In`
+Drawing inspiration from {lean}`List.In`, write a recursive function `All`
+stating that some property `a` holds of all elements of a list
```
xhalo32:
```suggestion
stating that some property `p` holds of all elements of a list
```

LF/Logic.lean
```diff
@@ -1410,35 +1479,35 @@ theorem In_map_iff (α β : Type) (f : α → β) (xs : List α) (y : β) :
 ::::::full
 :::::exercise (rating := 3) (name := "All")
 We noted above that functions returning propositions can be seen as
-_properties_ of their arguments. For instance, if `P` has type
-`Nat -> Prop`, then `P n` says that property `P` holds of `n`.
+_properties_ of their arguments. For instance, if `p` has type
+{lean}`Nat -> Prop`, then `p n` says that property `p` holds of {lean}`n`.
```
xhalo32:
```suggestion
{lean}`Nat → Prop`, then `p n` says that property `p` holds of {lean}`n`.
```

LF/Logic.lean
```diff
@@ -1463,23 +1532,23 @@ theorem All_In α (P : α → Prop) (xs : List α) :
 :::::exercise (rating := 2) (name := "combine_odd_even")
 Complete the definition of `combine_odd_even` below. It takes as arguments
 two properties of numbers, `Podd` and `Peven`, and it should return
-a property `P` such that `P n` is equivalent to `Podd n` when `n` is odd
+a property `a` such that `a n` is equivalent to `Podd n` when `n` is odd
```
xhalo32:
```suggestion
a property `p` such that `p n` is equivalent to `Podd n` when `n` is odd
```

LF/Logic.lean
```diff
@@ -1579,42 +1657,43 @@ we want to prove the following:
...
 If we really wanted, we could in fact do it for both rewrites.
 
 ```lean
 example (x y z : Nat) : x + (y + z) = (z + y) + x := by
-  rw [add_comm x (y + z)]
-  rw [add_comm z y]
+  rw [Nat.add_comm x (y + z)]
+  rw [Nat.add_comm z y]
 ```
 ::::
```
xhalo32: Should we wrap this in `namespace Nat` or use `open Nat`? I don't like that this change introduces extra keystrokes

xhalo32: Seems like there are more cases of new `Nat.` further down the file

LF/Logic.lean
```diff
@@ -2103,121 +2182,115 @@ theorem orb_true_iff (b1 b2 : Bool) :
...
-of the `beq_list` function below. to make sure that your definition
-is correct, prove the lemma `beq_list_true_iff`.
+of some type {lean}`α`, we can define a function `beqList` for testing
+equality of lists with elements in {lean}`α`. Complete the definition
+of the `beqList` function below. to make sure that your definition
+is correct, prove the lemma `beqList_true_iff`.
 
 ```lean
-def beq_list {α : Type} (beq : α → α → Bool) (xs1 xs2 : List α) : Bool := solution!(
+def beqList {α : Type} (beq : α → α → Bool) (xs1 xs2 : List α) : Bool := solution!(
```
xhalo32:
```suggestion
def List.beq {α : Type} (beq : α → α → Bool) (xs1 xs2 : List α) : Bool := solution!(
```

or we could put this in `namespace List`

LF/Logic.lean
```diff
@@ -2103,121 +2182,115 @@ theorem orb_true_iff (b1 b2 : Bool) :
...
     (h : ∀ x1 x2, beq x1 x2 = true ↔ x1 = x2) :
-    ∀ xs1 xs2, beq_list beq xs1 xs2 = true ↔ xs1 = xs2 := by
+    ∀ xs1 xs2, beqList beq xs1 xs2 = true ↔ xs1 = xs2 := by
   solution!
     intro xs1; induction xs1
     case nil =>
       intro xs2; cases xs2
       case nil =>
-        rw [beq_list_nil_nil]; constructor
+        rw [beqList_nil_nil]; constructor
```
xhalo32:
```suggestion
        rw [beqList_nil_nil]
        constructor
```
applies to other instances too

LF/Logic.lean
```diff
@@ -2103,121 +2182,115 @@ theorem orb_true_iff (b1 b2 : Bool) :
...
+`Tactics.forall_exists_challenge`, to the {lean}`List.All` property defined above.
 
-Copy the definition of `forallb` from Tactics here so that this file can be
+Copy the definition of {lean}`forallb` from Tactics here so that this file can be
 graded on its own.
 
 ```lean
-def Logic.forallb {α : Type} (test : α → Bool) (xs : List α) : Bool := solution!(
-  match xs with
+def Logic.forallb {α : Type} (test : α → Bool) (l : List α) : Bool := solution!(
```
xhalo32:
```suggestion
def List.allb {α : Type} (test : α → Bool) (l : List α) : Bool := solution!(
```
or `List.all` ([standard library](https://leanprover-community.github.io/mathlib4_docs/Init/Data/List/Basic.html#List.all))

LF/UsingLean.lean
```diff
@@ -600,22 +600,22 @@ talk about using `Nat.add_zero` and friends from now on.
...
 
 :::::exercise (rating := 2) (name := "double_add")
 ```lean
-theorem double_add (n : Nat) : n.double = n + n := by
+theorem Nat.double_add (n : Nat) : n.double = n + n := by
   solution!
     induction n with
     | zero =>
-      rw [double_zero]
+      rw [Nat.double_zero]
```
xhalo32: Can't we just write `rw [double_zero]` because the theorem is in the `Nat` namespace? Same applies to other instances

xhalo32:
I think this is going in a good direction. My major points (all about style though):
- Theorem names should start with a lowercase letter
- If possible, let's avoid having to type `Nat.` over and over again. We should consider putting examples and theorems in the `Nat` namespace to avoid the extra keystrokes.

dsainati1:
I don't think keystrokes are really something to optimize for; imo it's more important not to clutter the namespace with junk theorems.

LF/UsingLean.lean
dsainati1: I removed unnecessary uses of `Nat.` within theorems, but I don't want to add junk theorems and examples into the Nat namespace.

xhalo32:
For junk theorems I agree, but `example`s don't clutter the namespace.

dsainati1:
So you want to `open Nat in` before `example`s? That seems reasonable

LF/Logic.lean
dsainati1: Hmm, even after adding an `open Nat in` to the top of this, removing the `Nat.` still gives me a type error.

bcpierce00:
#Note: Daniel could use a hand getting this finished -- on vacation till 8/7

xhalo32:
> So you want to `open Nat in` before `example`s? That seems reasonable

I was thinking more just surrounding the relevant section in `namespace Nat` ... `end Nat`

xhalo32:
I'm fine with getting this merged and moving the discussion about whether to have examples in relevant namespaces (or to use `open`) elsewhere.

xhalo32 (approved):

LF/Logic.lean
```diff
@@ -2359,42 +2448,91 @@ However, we cannot prove this equality by reflexivity, as the two sides
...
+n m : Nat
+α✝ : Type
+e1 e2 x✝ y✝ : α✝
+α β : Type
+x x' y : α
+l l' : List α
+f g : α → β
+p : α → Prop
+a b : Prop
+⊢ a ∧ b = b ∧ a
```
berberman, line 2470: These variables defined for verso presentation should not appear in the context...

LF/Logic.lean
berberman: Because `Nat.add_comm` has `protected` modifier. `open` won't bring in `protected` definitions, and I think they are supposed to be prefixed with `Nat.`.

LF/Logic.lean
berberman: Converted to dev comments. I'll try to fix this in SFLMeta.

berberman (approved):

PR #169, opened 2026-07-30
added the TotalMap update lemmas

meluge:
I've added 4 lemmas about updating TotalMap: `update_neq, update_shadow, update_same, update_permute.' They are needed as the proofs for Hoare.lean uses them. I have the lemmas also in Hoare.lean to make the proofs go through but I'll refactor to use these ones in Typeclasses.lean after the PR.

### DIFF
--- a/LF/Typeclasses.lean
+++ b/LF/Typeclasses.lean
@@ -469,9 +469,74 @@ theorem update_eq (m : TotalMap α β) (a : α) (b : β) : (a →ₜ b ; m)[a] =
   rfl
 ```
 
-:::dev
-exercises here...
-:::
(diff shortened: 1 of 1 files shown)


mwhicks1 (changes requested):
I believe these lemmas were added a couple of days ago. Notice the conflict with `LF/Typeclasses.lean` identified here. Please work from the most recent `Typeclasses` version.

meluge:
Oh sorry, I did not see that version. Thanks!

xhalo32:
Also see https://github.com/plclub/sf-in-lean/pull/166 which does some major changes to Typeclasses.lean. I can help fix any problems it causes in Hoare.lean!

PR #181, opened 2026-07-31
StlcProp -- initial conversion

mwhicks1:
As usual: 
- Used `to_verso` to make initial conversion from the Rocq
- Claude translated the definitions and did the proofs (which we will replace), I reviewed
- I went over all the text to make sure it reads Ok and follows the original, and deleted stale comments


TS/StlcProp.lean
```diff
@@ -0,0 +1,2302 @@
...
+```
+
+# Progress
+
+::::full
+The _progress_ theorem tells us that closed, well-typed
+terms are not stuck: either a well-typed term is a value, or it
+can take a reduction step.  The proof is a relatively
+straightforward extension of the progress proof we saw in the
+{ref "Types"}[Types] chapter.  We give the proof in English first, then
```
Robertboy18, line 173: Small mismatch here: the Lean proof is shown before the English proof now, but this still says the English proof comes first. Could we update the sentence?

TS/StlcProp.lean
```diff
@@ -0,0 +1,2302 @@
...
+      exact .ite _ _ _ _ _ (ihc _ _ h₁) (iht _ _ h₂) (ihe _ _ h₃)
+```
+
+::::full
+The substitution lemma can be viewed as a kind of "commutation
+property."  Intuitively, it says that substitution and typing can
+be done in either order: we can either assign types to the terms
+`t` and `v` separately (under suitable contexts) and then combine
+them using substitution, or we can substitute first and then
+assign a type to ` `x:=v` t`; the result is the same either
```
Robertboy18: I think the substitution notation got mangled during conversion here. Should this read `[x := v]t`?

TS/StlcProp.lean
```diff
@@ -0,0 +1,2302 @@
...
+An exercise in the {ref "Types"}[Types] chapter asked about the _subject
+expansion_ property for the simple language of arithmetic and
+boolean expressions.  This property did not hold for that language,
+and it also fails for STLC.  That is, it is not always the case that,
+if `t ⟶ t'` and `empty ⊢ t' ⦂ T`, then `empty ⊢ t ⦂ T`.
+Show this by giving a counter-example that does _not involve
+conditionals_.
+
+:::solution
+COMMENT: `untagged (* .. *) comment at prose position in the Rocq source — probably an error there; decide whether it is book prose or an author note`
```
Robertboy18: This `COMMENT:` line looks like a conversion note rather than part of the solution. Could we remove it or turn it into a `:::dev` note so readers don’t see it?

rogerburtonpatel:
There's a problem here. On the line `| var Γ x T₁ h =>` of the first big proof, `progress`, the proof goal displays an infinite nest of parentheses. Debugging this with `set_option pp.deepTerms true` above the theorem leads to the language server hanging. There is a loop somewhere in the prettyprinter. Did anyone check this out? I can make a fix, but it is troubling...

rogerburtonpatel:
I'm proceeding with reviewing the file, but it is a bit concerning that an issue this obvious was not checked. Even the types on hover have infinite parenthesis with a `...` in them.

mwhicks1:
> On the line `| var Γ x T₁ h =>` of the first big proof, `progress`

@rogerburtonpatel I thought the idea was that you would delete the proofs entirely and do them from scratch? If you do that, does the same problem arise? I.e., is it related to a busted delaborator, perhaps?

rogerburtonpatel:
> > On the line `| var Γ x T₁ h =>` of the first big proof, `progress`
> 
> @rogerburtonpatel I thought the idea was that you would delete the proofs entirely and do them from scratch? If you do that, does the same problem arise? I.e., is it related to a busted delaborator, perhaps?

It is a busted delaborator; I fixed it... it made rewriting the proofs impossible as every variable appeared as 

((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((...)))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))

rogerburtonpatel:
I've now continued. Let me push just the delab fix

bcpierce00:
#Note: What do we need to do to get this merged?

TS/StlcProp.lean
```diff
@@ -0,0 +1,2348 @@
+import SFLMeta
+import TS.Stlc
+import LF.CustomTactics
+open Verso.Genre Manual
+open SFLMeta
+
+set_option maxHeartbeats 1000000
```
chenson2018: Is this actually needed or an AI artifact?

Edit: I see the `dev` comment below, but don't put much stock in it without human confirmation. It's a common thing AI will automatically do when writing Lean.

dsainati1:
@rogerburtonpatel in the interest of getting this merged in its in-progress state, can we either mark all the remaining AI proofs as such or replace them with sorry?

rogerburtonpatel:
Yes, let me mark them as AI. `sorry`ing them actually would make things harder for us down the line in this case

rogerburtonpatel:
Man, many of these have the `exact` pattern which we decided to keep before but are now getting rid of (phew). So many will be marked as in progress between AI and human since I fixed everything but that.

rogerburtonpatel:
Done

dsainati1 (approved):
This is in progress but let's get it merged for now

bcpierce00:
Are all the comments here reflected in the chapter text?

PR #184, opened 2026-08-01
Rework presentation at the start of Typeclasses

mwhicks1:
I have changed the introductory narrative to motivate typeclasses better by bringing forward the `List.elem` example. I then follow with the development of the `Inhabited` typeclass to explain how you make one (with a forw. ref to the use of `Inhabited` later). Finally I swing back to `List.elem` and complete the example to show how the `Beq` typeclass solves the problem we uncovered there. The rest of the chapter proceeds as before.

I also added a forward reference to this chapter from `Lists`, which first introduces `Beq`.

Claude helped me work out this re-do and did much of the writing at my direction.

### DIFF
--- a/LF/Typeclasses.lean
+++ b/LF/Typeclasses.lean
@@ -26,64 +26,141 @@ htmlSplit := .never
 file := "Typeclasses"
 %%%
 
-
-So far we have discussed *parametric polymorphism*, where in Lean we can declare a type variable
+Chapter {ref "Poly"}[Poly] introduced *parametric polymorphism*, declaring a type variable with no
+constraint on it:
 
 ```lean
 variable (α : Type)
 ```
 
-without any further information about the details of the type `α`. Here we are able to work
-with a type like {InlineLean.lean}`List α`, and can write functions like {name}`List.reverse` or
-{name}`List.length` that only operate on the structure of a list independently of its contents.
-Along with these we have corresponding proofs like {name}`List.length_reverse` that are similarly
-structural in nature.
+This lets us work with a type like {InlineLean.lean}`List α`, writing functions like
+{name}`List.reverse` and {name}`List.length`, and proofs like {name}`List.length_reverse`, that use
+only the list's structure and never inspect any particular `a : α`.
 
-This is somewhat limiting, as we cannot inspect or non-trivially use any particular `a : α`. What we
-might want in some situations rather than `α` being completely generic is to partially specify its
-behavior. In Lean, this happens through a form of "ad hoc polymorphism" called *typeclasses*. This
-concept originated in the Haskell programming language, and is analogous to features you may be
-familiar with in other languages such as traits in Rust.
+Sometimes, though, we want less freedom: rather than `α` being completely generic, we want to
+partially specify its behavior. In Lean, this is done through a form of "ad hoc polymorphism" called
+*typeclasses*. The concept originated in Haskell, and is analogous to features you may know from
+other languages, such as traits in Rust.
 
-# First Example: Inhabited Types
+# Why We Need Typeclasses
 
-Suppose we wanted to specify that a type has at least one inhabitant — i.e.,
-that it is not empty. We have previously seen `structure`, which would allow us to express
-this as
+Consider the following function, which checks whether a natural number occurs in a list of natural
+numbers:
+
+```lean
+@[irreducible]
+def List.elem_nat (a : Nat) (xs : List Nat) : Bool :=
+  match xs with
+  | [] => false
+  | b :: tl => bif a == b then true else elem_nat a tl
+
+unseal List.elem_nat in
+theorem List.elem_nat_nil (a : Nat) : [].elem_nat a = false := rfl
+
+unseal List.elem_nat in
+theorem List.elem_nat_cons (a b : Nat) (xs : List Nat) :
+    (b :: xs).elem_nat a = bif a == b then true else elem_nat a xs := rfl
+```
+
+```lean
+#eval [0, 1].elem_nat 0
+#eval [0, 1].elem_nat 1
+#eval [0, 1].elem_nat 2
+```
+
+What if we want this to work for lists of *any* element type, not just `Nat`? Parametric
+polymorphism suggests simply replacing `Nat` with a type variable `α` — but that produces a puzzling
+error:
+
+:::dev
+@dsainati - The Verso compilation is not actually removing this from the generated file despite th e
+-keep flag. It is, however, stripping the message guard, so this results in an error. Once
+we figure out how we are handling these cases, uncomment this.
+
+```lean -keep
+/--
+error: failed to synthesize instance of type class
+  BEq α
+
+Hint: Type class instance resolution failures can be inspected with the `set_option trace.Meta.synthInstance true` command.
+-/
+#guard_msgs in
+def List.elem_poly {α : Type} (a : α) (xs : List α) : Bool :=
+  match xs with
+  | [] => false
+  | b :: tl => bif a == b then true else elem_poly a tl
+```
+:::
+
+Lean can't find a needed instance of the typeclass {name}`BEq`. This is Lean trying to use
+typeclasses to work out how `==` should behave on a value of unspecified type `α` — not every type
+gets `==` for free. As {ref "Lists"}[Lists] noted when we first used it, `==` on `Nat` comes from
+the `BEq` typeclass, an interface each type has to implement for itself.
+
+We could sidestep typeclasses entirely, and just have the caller supply the equality test to use:
+
+```lean -keep
+def List.elem_poly {α : Type} (eq : α → α → Bool) (a : α) (xs : List α) : Bool :=
+  match xs with
+  | [] => false
+  | b :: tl => bif eq a b then true else elem_poly eq a tl
+
+#eval [0, 1].elem_poly Nat.beq 0
+```
+
+This works, but it's tedious: every caller has to know, and remember to supply, the right equality
+function. Typeclasses automate exactly this — instead of the programmer passing the function
+explicitly, Lean searches for one on its own. We specify what to search for by declaring a `class`
+with the needed function as a field. Here is where we end up:
+
+```lean
+@[irreducible]
+def List.elem_poly {α : Type} [BEq α] (a : α) (xs : List α) : Bool :=
+  match xs with
+  | [] => false
+  | b :: tl => bif a == b then true else elem_poly a tl
+
+unseal List.elem_poly in
+theorem List.elem_poly_nil [BEq α] (a : α) : [].elem_poly a = false := rfl
+
+unseal List.elem_poly in
+theorem List.elem_poly_cons [BEq α] (a b : α) (xs : List α) :
+    (b :: xs).elem_poly a = bif a == b then true else elem_poly a xs := rfl
+```
+
+The `[BEq α]` argument is filled in automatically, based on whatever `α` the caller uses, and it's
+what {name}`List.elem_poly` uses internally wherever it writes `==`. This is the *ad hoc
+polymorphism* we mentioned above: {name}`List.elem_poly` isn't generic over every type, only over
+types that support `==`. We'll see exactly how `[BEq α]` gets filled in below — starting with how
+to define a typeclass in the first place.
+
+# Defining Your Own Typeclasses
+
+`BEq` comes from Lean's standard library. Let's define a typeclass of our own, to see the
+mechanism — instances and synthesis — that made `==` resolve automatically above.
+
+Suppose we want to specify that a type has at least one inhabitant, i.e., that it is not empty. A
+`structure` (chapter {ref "Lists"}[Lists]) can express this directly:
 
 ```lean
 structure HasOneStruct (α : Type) where
   one : α
 ```
 
-As discussed previously, a structure is just an inductive type that comes with a single constructor, in this case {name}`HasOneStruct.mk`, along with a few different syntaxes for convenience:
+A value of {InlineLean.lean}`HasOneStruct Nat`, such as {InlineLean.lean}`HasOneStruct.mk (1 : Nat)`,
+witnesses that `Nat` is inhabited: it contains an element, namely `1`. But `HasOneStruct Nat` is
+just a wrapper around a `Nat`, the same way {InlineLean.lean}`List Nat` is — nothing makes Lean
+produce that witness *automatically*, the way it found a `BEq Nat` instance for us above.
 
-```lean
-def nat_hasOneStruct_mk : HasOneStruct Nat := HasOneStruct.mk 1
-
-def nat_hasOneStruct_brace : HasOneStruct Nat := {one := 1}
-
-def nat_hasOneStruct_where : HasOneStruct Nat where
-  one := 1
-
-def nat_hasOneStruct_anonymous : HasOneStruct Nat := ⟨1⟩
-```
-
-All of these are the same value, a structure of type {InlineLean.lean}`HasOneStruct Nat` that we
-can thing of as carrying a piece of data (a term that is not a proposition), the number `1`, as evidence that this type is not empty. We can use this in proofs such as
-
-```lean
-theorem nat_hasOneStruct_eq_one : nat_hasOneStruct_mk.one = 1 := rfl
-```
-
-to see how this structure provably contains the value that we have placed within it. In Lean, typeclasses work very similarly, and in fact are implemented as structures. The only difference in defining a typeclass is that we instead use the `class` keyword:
+Typeclasses solve this. In Lean, they're implemented as structures and declared the same way, but
+with `class` in place of `structure`:
 
 ```lean
 class HasOne (α : Type) where
   one : α
 ```
 
-While the definition is essentially the same, including the constructor {name}`HasOne.mk`, the usage of typeclasses is different, and what allows us to use them to express ad hoc polymorphism. Instead of using `def` to define inhabitants of this type, we will instead use the `instance` keyword:
+The difference is in how we provide values of this type. Instead of `def`, we use `instance`:
 
 :::dev
 From GitHub (@chenson2018): Part of the point here is that we don't usually refer to instances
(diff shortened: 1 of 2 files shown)


LF/Typeclasses.lean
```diff
@@ -26,64 +26,141 @@ htmlSplit := .never
...
 variable (α : Type)
 ```
 
-without any further information about the details of the type `α`. Here we are able to work
-with a type like {InlineLean.lean}`List α`, and can write functions like {name}`List.reverse` or
-{name}`List.length` that only operate on the structure of a list independently of its contents.
-Along with these we have corresponding proofs like {name}`List.length_reverse` that are similarly
-structural in nature.
+This lets us work with a type like {InlineLean.lean}`List α`, writing functions like
+{name}`List.reverse` and {name}`List.length`, and proofs like {name}`List.length_reverse`, that use
```
dsainati1:
```suggestion
{name}`List.reverse` and {name}`List.length` and proofs like {name}`List.length_reverse`, which use
```

LF/Typeclasses.lean
```diff
@@ -26,64 +26,141 @@ htmlSplit := .never
...
+only the list's structure and never inspect any particular `a : α`.
 
-This is somewhat limiting, as we cannot inspect or non-trivially use any particular `a : α`. What we
-might want in some situations rather than `α` being completely generic is to partially specify its
-behavior. In Lean, this happens through a form of "ad hoc polymorphism" called *typeclasses*. This
-concept originated in the Haskell programming language, and is analogous to features you may be
-familiar with in other languages such as traits in Rust.
+Sometimes, though, we want less freedom: rather than `α` being completely generic, we want to
```
dsainati1:
```suggestion
Sometimes, though, we want less freedom: rather than leaving `α` completely generic, we want to
```

LF/Typeclasses.lean
```diff
@@ -26,64 +26,141 @@ htmlSplit := .never
...
+partially specify its behavior. In Lean, this is done through a form of "ad hoc polymorphism" called
+*typeclasses*. The concept originated in Haskell, and is analogous to features you may know from
```
dsainati1:
```suggestion
*typeclasses*. The concept originated in Haskell and is analogous to features you may know from
```

LF/Typeclasses.lean
```diff
@@ -26,64 +26,141 @@ htmlSplit := .never
...
+other languages, such as traits in Rust.
 
-# First Example: Inhabited Types
+# Why We Need Typeclasses
 
-Suppose we wanted to specify that a type has at least one inhabitant — i.e.,
-that it is not empty. We have previously seen `structure`, which would allow us to express
-this as
+Consider the following function, which checks whether a natural number occurs in a list of natural
+numbers:
```
dsainati1:
```suggestion
Consider the following function, which checks whether a natural number occurs in a list:
```

LF/Typeclasses.lean
```diff
@@ -26,64 +26,141 @@ htmlSplit := .never
...
 ```lean
-def nat_hasOneStruct_mk : HasOneStruct Nat := HasOneStruct.mk 1
+#eval [0, 1].elem_nat 0
+#eval [0, 1].elem_nat 1
+#eval [0, 1].elem_nat 2
+```
 
-def nat_hasOneStruct_brace : HasOneStruct Nat := {one := 1}
+What if we want this to work for lists of *any* element type, not just `Nat`? Parametric
+polymorphism suggests simply replacing `Nat` with a type variable `α` — but that produces a puzzling
```
dsainati1:
```suggestion
polymorphism suggests simply replacing `Nat` with a type variable `α`, but that produces a puzzling
```

LF/Typeclasses.lean
```diff
@@ -26,64 +26,141 @@ htmlSplit := .never
...
+  | b :: tl => bif a == b then true else elem_poly a tl
 ```
+:::
 
-All of these are the same value, a structure of type {InlineLean.lean}`HasOneStruct Nat` that we
-can thing of as carrying a piece of data (a term that is not a proposition), the number `1`, as evidence that this type is not empty. We can use this in proofs such as
+Lean can't find a needed instance of the typeclass {name}`BEq`. This is Lean trying to use
+typeclasses to work out how `==` should behave on a value of unspecified type `α` — not every type
+gets `==` for free. As {ref "Lists"}[Lists] noted when we first used it, `==` on `Nat` comes from
+the `BEq` typeclass, an interface each type has to implement for itself.
```
dsainati1:
```suggestion
Lean is trying to use typeclasses to work out how `==` should behave on a value of type `α`, 
but it can't find the instance of the typeclass {name}`BEq` that it needs, since not every type can be checked for equality. As {ref "Lists"}[Lists] noted when we first used it, `==` on `Nat` comes from the `BEq` typeclass, an interface each type has to implement for itself.
```

LF/Typeclasses.lean
```diff
@@ -26,64 +26,141 @@ htmlSplit := .never
...
+
+We could sidestep typeclasses entirely, and just have the caller supply the equality test to use:
```
dsainati1:
```suggestion
We could sidestep typeclasses entirely and just have the caller supply the equality test to use:
```

LF/Typeclasses.lean
```diff
@@ -26,64 +26,141 @@ htmlSplit := .never
...
+def List.elem_poly {α : Type} (eq : α → α → Bool) (a : α) (xs : List α) : Bool :=
+  match xs with
+  | [] => false
+  | b :: tl => bif eq a b then true else elem_poly eq a tl
+
+#eval [0, 1].elem_poly Nat.beq 0
+```
+
+This works, but it's tedious: every caller has to know, and remember to supply, the right equality
+function. Typeclasses automate exactly this — instead of the programmer passing the function
```
dsainati1:
```suggestion
function. Typeclasses automate exactly this; instead of the programmer passing the function
```

LF/Typeclasses.lean
```diff
@@ -26,64 +26,141 @@ htmlSplit := .never
...
+explicitly, Lean searches for one on its own. We specify what to search for by declaring a `class`
+with the needed function as a field. Here is where we end up:
```
dsainati1:
```suggestion
with the needed function as a field:
```

LF/Typeclasses.lean
```diff
@@ -26,64 +26,141 @@ htmlSplit := .never
...
+
+unseal List.elem_poly in
+theorem List.elem_poly_cons [BEq α] (a b : α) (xs : List α) :
+    (b :: xs).elem_poly a = bif a == b then true else elem_poly a xs := rfl
+```
+
+The `[BEq α]` argument is filled in automatically, based on whatever `α` the caller uses, and it's
+what {name}`List.elem_poly` uses internally wherever it writes `==`. This is the *ad hoc
+polymorphism* we mentioned above: {name}`List.elem_poly` isn't generic over every type, only over
+types that support `==`. We'll see exactly how `[BEq α]` gets filled in below — starting with how
```
dsainati1:
```suggestion
types that support `==`. We'll see exactly how `[BEq α]` gets filled in below, starting with how
```

dsainati1 (approved):
LGTM, just some writing nits for Claude

LF/Typeclasses.lean
```diff
@@ -127,21 +207,74 @@ set_option pp.all true in
...
+which reveals {InlineLean.lean}`@HasOne.one Nat instHasOneNat` and
+{InlineLean.lean}`@HasOne.one Int instHasOneInt` on the right-hand side of each equality. The
+`#synth` command runs the same search directly:
 
 ```lean
 #synth HasOne Nat
 ```
 
-for instance prints {name}`instHasOneNat`, informing us that this is the instance found that has type {InlineLean.lean}`HasOne Nat`. In general, the assumption is that for typeclasses like {name}`HasOne` that include data, as opposed to solely proofs, that these these instances should be unique to avoid ambiguity.
+prints {name}`instHasOneNat`. For a typeclass like {name}`HasOne` that carries data (as opposed to
```
xhalo32: We could use `leanOutput` here. For an example [see Poly.lean](https://github.com/plclub/sf-in-lean/blob/614820201c14922eb6dbd0636bdd141a546f2a59/LF/Poly.lean#L135-L141)

LF/Typeclasses.lean
```diff
@@ -26,64 +26,141 @@ htmlSplit := .never
...
-that it is not empty. We have previously seen `structure`, which would allow us to express
-this as
+Consider the following function, which checks whether a natural number occurs in a list of natural
+numbers:
 
 ```lean
-structure HasOneStruct (α : Type) where
-  one : α
-```
+@[irreducible]
```
chenson2018: This was written some time ago, my is understanding now that we don't write `irreducible` this far into the book? cc @dsainati1

LF/Typeclasses.lean
```diff
@@ -26,64 +26,141 @@ htmlSplit := .never
...
+def List.elem_poly {α : Type} (a : α) (xs : List α) : Bool :=
+  match xs with
+  | [] => false
+  | b :: tl => bif a == b then true else elem_poly a tl
 ```
+:::
 
-All of these are the same value, a structure of type {InlineLean.lean}`HasOneStruct Nat` that we
-can thing of as carrying a piece of data (a term that is not a proposition), the number `1`, as evidence that this type is not empty. We can use this in proofs such as
+Lean can't find a needed instance of the typeclass {name}`BEq`. This is Lean trying to use
```
chenson2018: The result of moving this up is that the terminology it uses isn't defined first, right? We've not explained what an instance or a typeclass is at this point.

LF/Typeclasses.lean
```diff
@@ -26,64 +26,141 @@ htmlSplit := .never
...
+Suppose we want to specify that a type has at least one inhabitant, i.e., that it is not empty. A
+`structure` (chapter {ref "Lists"}[Lists]) can express this directly:
+
+```lean
+structure HasOneStruct (α : Type) where
+  one : α
 ```
 
-to see how this structure provably contains the value that we have placed within it. In Lean, typeclasses work very similarly, and in fact are implemented as structures. The only difference in defining a typeclass is that we instead use the `class` keyword:
+A value of {InlineLean.lean}`HasOneStruct Nat`, such as {InlineLean.lean}`HasOneStruct.mk (1 : Nat)`,
```
chenson2018: This now works:
```suggestion
A value of {lean}`HasOneStruct Nat`, such as {lean}`HasOneStruct.mk (1 : Nat)`,
```

LF/Typeclasses.lean
```diff
@@ -26,64 +26,141 @@ htmlSplit := .never
...
+witnesses that `Nat` is inhabited: it contains an element, namely `1`. But `HasOneStruct Nat` is
+just a wrapper around a `Nat`, the same way {InlineLean.lean}`List Nat` is — nothing makes Lean
+produce that witness *automatically*, the way it found a `BEq Nat` instance for us above.
```
chenson2018: We continue to use the word "instance" here without having properly defined it. Even if it's without an example first, we need to say something that indicates that it is a particular value of the typeclass.

LF/Typeclasses.lean
```diff
@@ -96,28 +173,31 @@ instance instHasOneNat : HasOne Nat where
...
+Now {name}`HasOne.one` alone, with no instance named, is enough — Lean infers which instance we
+mean from the expected type:
 
 ```lean
 example : HasOne.one = (1 : Nat) := rfl
 example : HasOne.one = (-1 : Int) := rfl
 ```
 
-and based on the type annotations Lean will infer which of these instances you intended to use. So these proofs are equivalent to
+These elaborate to the same terms as naming the instances explicitly:
```
chenson2018: Are we using the word "elaborate" in earlier chapters? I think it is confusing to a beginner without context.

LF/Typeclasses.lean
```diff
@@ -26,64 +26,141 @@ htmlSplit := .never
...
-structure HasOneStruct (α : Type) where
-  one : α
-```
+@[irreducible]
+def List.elem_nat (a : Nat) (xs : List Nat) : Bool :=
+  match xs with
+  | [] => false
+  | b :: tl => bif a == b then true else elem_nat a tl
 
-As discussed previously, a structure is just an inductive type that comes with a single constructor, in this case {name}`HasOneStruct.mk`, along with a few different syntaxes for convenience:
```
chenson2018, line 59: It's hard to see in the diff, but this section seems to have been dropped completely for "`HasOneStruct Nat` is a just a wrapper around a `Nat`, the same way `List Nat` is".

I find this explanation more confusing in mentioning `List`. The point I was trying to make is that typeclasses are just inferred structures, and I wanted to reinforce the different ways you work with them. I think these little concrete examples are helpful here (`nat_hasOneStruct_*`, ` nat_hasOneStruct_eq_one`).

LF/Typeclasses.lean
```diff
@@ -26,64 +26,141 @@ htmlSplit := .never
...
+#guard_msgs in
+def List.elem_poly {α : Type} (a : α) (xs : List α) : Bool :=
+  match xs with
+  | [] => false
+  | b :: tl => bif a == b then true else elem_poly a tl
 ```
+:::
 
-All of these are the same value, a structure of type {InlineLean.lean}`HasOneStruct Nat` that we
-can thing of as carrying a piece of data (a term that is not a proposition), the number `1`, as evidence that this type is not empty. We can use this in proofs such as
```
chenson2018, line 73: The terminology "data" is now used without this explanation of its meaning.

chenson2018 (changes requested):
I like the core of this change. The most important comments below are what I think are some regressions in clearly defining termonology before using it and some removing of examples I felt were helpful.

LF/Typeclasses.lean
mwhicks1: I prefer the dash -- it is a stronger connection between the two parts of the sentence. (Or a colon could work.)

LF/Typeclasses.lean
mwhicks1: I think there is a bigger problem here which I'll fix separately.

LF/Typeclasses.lean
mwhicks1: Good idea!

LF/Typeclasses.lean
mwhicks1: We first show a typeclass instance, and use that term, [back in Lists](https://github.com/plclub/sf-in-lean/blob/main/LF/Lists.lean#L401). One might argue that that first mention is a little sparse too. In any case, I'll add some text back in the motivating example that introduces the term.

LF/Typeclasses.lean
mwhicks1: I will drop it here from `elem_nat` too. If our goal was to drop at some point in the book, we can't wait longer than the last chapter!

LF/Typeclasses.lean
mwhicks1: The first mention of it got dropped in the refactor; I added back the inline definition at first reference.

LF/Typeclasses.lean
mwhicks1: Good catch. Removed the term.

LF/Typeclasses.lean
mwhicks1: It did get dropped because I found it a bit noisy: We don't need to show readers all the different ways to create structure objects in a chapter on typeclasses. We have already covered `structure` objects before this point.

But I do agree with you that with the relocation of text, the `List` mention around `List Nat` is a bit odd, so I've reworked that. See what you think of the new version.

chenson2018: I can understand this reasoning. My only complaint with this change is that you now switch styles without much of a reason:

```lean
def nat_hasOneStruct : HasOneStruct Nat := ⟨1⟩

instance instHasOneNat : HasOne Nat where
  one := 1
```

Let's have them both use the `where` version for consistency.

chenson2018 (approved):
One comment above to address and I pushed a small wording fix/addition, but otherwise this seems fine.

LF/Typeclasses.lean
```diff
@@ -24,66 +24,159 @@ open InlineLean hiding lean
...
+
+#eval [0, 1].elem_poly 0
+```
+
+As {ref "Lists"}[Lists] noted when we first used it, `==` on `Nat` comes from the `BEq` typeclass,
+and it's what {name}`List.elem_poly` uses internally wherever it writes `==`. This is done using 
+*instance implicits*, where we place a desired typeclass assumption in square bracket.
+The `[BEq α]` constraint is saying that an instance of {name}`BEq` must be provided at
+call sites for the _particular_ type `α` that is used. In the example {lean}`[0, 1].elem_poly 0`,
+this type is `Nat`, and the automatically chosen instance contains {name}`Nat.beq`.
```
xhalo32: This is technically incorrect, the instance `BEq Nat`, which comes from `DecidableEq`, does not contain `Nat.beq`. @chenson2018 said that we intentinally ignore this, but I think we should pick the words here more carefully.

[Here's a demonstration](https://live.lean-lang.org/#codez=LTAEAMF5PUEsDOoB2B7ALhAcgQ3QOgCMBTAR1kIE9QATYgMzmTnTlWVBIGMcBXBYvHQByJHQBOcAG7EaoeuNQBbUABFiXODRyEANsQCipAFABiBJWToAFqABCR0LnTGzXaxoDWoAJJWiZMIA%2BmRBJKSgIL5WoNY4SDaI8jhc6AA0nLyYAsRKCaicgkqICEwA5vKKKs6uNsSo4rlOeAGkIW3hoADeOJygAFzN6AC%2BA0OtoL2EoJCgABS90JwAlAOzVMagkWDi9LrbtKjECMjCmADuDZ6bsTgysWNzzhNTM6Do4rzEq7MLM%2BurfrrailJQABxucXug0AAETdV6DOyoVC6YYZP6zD5fH7zaaYz7fUCAJMJJm9pkDONQmB8ClMANygHgCJC9AA8dIAfIz4sc%2BuyuaCIVtiAAPFKYRJIayuOj0UAAGUQBGI%2BiUQWQeHmvUGzlWcxFSEGioQmF1YyRKLWNyUeHcoANoHOLGlWwAPqAANoAXRmXPoOF0Ahu7vJg3Q%2B0gXMIcDli3W7w8HCxghVAlAKty6s1vXDMoYCqV%2BAzarBKMo7W6gEbgMYAFUoYOIozmZDG1ZJbfsyN0eu1oEreodRqVfcBnctQOtttsDqdNmDnp9keSgeI89D7wjUZj6YirzqSYJ6ZXR8zpd05ZbLZzujzcuNytVQTP1C61cGdYbow9DgilZ9f0Gft5kHAsTRHc0uytLYbXQO0Z2dedvV9ZcgzdPp%2BjDTdOG3OM%2Bn3d5D1TFNH2fUlc2MAB6YBjHvItHw1TAcGQORaOLJ8y1AfhBCgGAMkILIIFYkiywrZ5wlgGgjhOM58FceViBEJBGHEMCcliVBzkEws2OfUSWk6XoHVmITMwY0kDVgRpSF4OBGiQcAxNCUJxMyTBwBMksRJbOYAHb%2FlAHzVkMpBjO0%2Bjs3tBAJKk04CGMYAKNcKQcEkHR9HhMZnFGLoQPvIZhlcUUcHBdKhxNOjTy8iJHN3SK3g8rNGLqikNi2JgaF4VI2A4eC5zQ5gsN2G80K4dgkBFOq4FsSMbi2GhBU9BrdLIDIGoYr1ZtAcRNI9KaNq2LZRuQMpxE27bPRq9owjIH0ojQJrZTIQqxRKwQyofSrzwrXz%2FMC8yQtAj61TM4K1kpG52s61h2Dq2cXVAd0BuQob5yO8bJumjlNvmuBwUWsLPovUhVoJ4G8H2g7zt26wNso6iAGVisEUbwRSxAYZIdBzmIYgOHc0n2K%2BshYCYuR%2BfKnSy3AWSEue4qwVKwGKs8oXqv0q86tCiXhPPf6wda%2BBmKh7rYYQ%2Fq4EGvZUbG%2B0Md9bGFo9JaqpJ7XCYprYqb2zajpOs6dsu5yboOe7SUekxjCK16xmd1X5j8pY%2FtBrWgcF6hQZaygIaNrqYd6%2BHEYt5GrZGm2JodKb7YO2hHdjonXdT58Pa2nbvaAA) of `List.elem_nat` versus `List.elem_poly_eq` versus `List.elem_poly` and how `Nat.beq` and `==` play a role.

chenson2018: I changed the wording here once already for the same reason. You can say

```suggestion
this type is `Nat`, and the automatically chosen instance corresponds to {name}`Nat.beq`.
```

if that is better still.

chenson2018: (Also to be clear: I said *avoid* not ignore. The original version I wrote of this was carefully worded to sidestep mentioning the particular instance)

mwhicks1: I made the change to "corresponds to". But: This line that I've edited is not the only place that `Nat.beq` is mentioned as if it corresponds to `==`. I was following Chris's lead on this since I'm not an expert! I'm going to copy Niklas' comment to this spot in the text in case we choose to revisit more generally.

LF/Typeclasses.lean
mwhicks1: Sounds good.

PR #185, opened 2026-08-01
Fixed paren divergence in STLC unparser; HT @Robert Joseph

rogerburtonpatel:
Fixes a bug. This is a good case for being able to push directly to main for minor fixes.

### DIFF
--- a/TS/Stlc.lean
+++ b/TS/Stlc.lean
@@ -600,7 +600,9 @@ partial def delabTmInner : DelabM (TSyntax `stlcTm) := do
           let t ← withAppArg delabTmInner
           `(stlcTm| [$x := $s] $t)
         else
-          match ← delab with
+          let partialSubst := e.getAppFn.constName? == some `Stlc.subst
+          match ← withOptions (fun o => if partialSubst then o.setBool `pp.notation false else o)
+                    delab with
           | `($i:ident) => `(stlcTm| $i:ident)
           | e => `(stlcTm| ~$e)
   (⟨·⟩) <$> annotateTermInfo ⟨stx.raw⟩
@@ -1768,6 +1770,12 @@ variable (x : String)
 /-- info: Stlc.Tm.var x : Tm -/
 #guard_msgs in
 #check Tm.var x
+
+variable (s : Tm)
+
+/-- info: subst x s : Tm → Tm -/
+#guard_msgs in
+#check subst x s
 ```
 :::
 


TS/Stlc.lean
```diff
@@ -600,7 +600,9 @@ partial def delabTmInner : DelabM (TSyntax `stlcTm) := do
           let t ← withAppArg delabTmInner
           `(stlcTm| [$x := $s] $t)
         else
-          match ← delab with
+          let partialSubst := e.getAppFn.constName? == some `Stlc.subst
+          match ← withOptions (fun o => if partialSubst then o.setBool `pp.notation false else o)
+                    delab with
```
berberman, line 605: I found the overall design of this delaborator confusing, especially the special handling for `subst` before it was defined and the interaction between this function and `delabSubst` below. For this PR, disabling `pp.notation` affects the entire fallback `delab`, which could be too board: e.g. `subst x <{ true }>` now prints `subst x Tm.tru : Tm → Tm` instead of `subst x <{ true }>`. Can we check the arity in `delabSubst` and drop this workaround? Also, it would be useful to have more test cases.

berberman (changes requested):

TS/Stlc.lean
rogerburtonpatel: Agreed that the design is confusing. It should be redone. 

I think we should focus on the optimal delab later. I was fixing a bug that led to divergence in printing and led to the proof state being unreadable. There is a better delab than this one; my feeling is to make it as its own project rather than trying to turn this bug fix into the perfect solution. 

The arity check sounds reasonable on paper, and tests are a must. I regret to suggest work without being able to contribute as I'm at capacity with SPLV right now :-(

bcpierce00: @rogerburtonpatel Should we just delete this PR, or turn it into an issue, or...?

#Note: Delete?

bcpierce00:
@rogerburtonpatel Should we just delete this PR, or turn it into an issue, or...?

david-christiansen:
This code definitely has an issue like that fixed in #256, and @berberman suggested a better way of fixing it here.

But I'd probably want to fix it a different way. Instead of setting `pp.notations` to `false` on unsaturated applications of `subst`, I'd set up the delaborator to only fire on saturated applications. This can be done by adding something like `guard <| ( <- getExpr).isAppOfArity 3` to the top (`delabTy` and `delabTm` already have arity checks based on `match_expr`, but the `subst` delaborator doesn't need to check what's applied, just the arity). When the guard fails, Lean's delaborator will try the next registered delaborator in the chain, which is the default application delaborator in this case.

As @berberman said, the downside of setting `pp.notations` to `false` here is that the blast radius is too big - it turns off notations for the whole sub-tree, rather than just the one case.

bcpierce00 (commented):
#Note: Delete? Rework following D.T.C.?

bcpierce00:
#Note: I'll delete this PR soon unless someone says otherwise.

PR #186, opened 2026-08-02
LF: Add notations appendix chapter

xhalo32:
Adds an appendix to LF that contains a reference of all of the notations used across the three volumes so far.


bcpierce00 (approved):

bcpierce00:
Now that we've had the discussion with David, I wonder whether we actually want to merge this -- i.e., whether we want to go to the trouble of making a real Notation appendix -- or whether it has served its purpose and going further is too much work for now.  (I tend to think the latter...)

mwhicks1:
I think that its benefit to students is probably not worth the maintenance version.

xhalo32:
I agree this has served its purpose and keeping it up-to-date is not worth it right now.

PR #188, opened 2026-08-02
LF: remove stale Induction solution-marker notes

Robertboy18:
I initially read these markers as a rendering bug, but after checking `STYLE.md` and the generated files, they are doing exactly what this exercise needs: they hide the whole declarations so students write the statements too. The two nearby developer notes recommending `solution!` are therefore stale.

This PR now only removes those two notes. It leaves every solution marker and Lean declaration unchanged.

Tested with `make lf`. I also checked the generated student, solutions, and terse `Induction.lean` files: student and terse contain `-- FILL IN HERE`, while solutions retain the declarations without the marker lines.

### DIFF
--- a/LF/Induction.lean
+++ b/LF/Induction.lean
@@ -1485,13 +1485,6 @@ It would be wise to do some `example` proofs to check that your
 definition of {name}`normalize` works the way you intend before you
 proceed. They won't be graded, but do fill in a few below.
 
-:::dev "Claude" BeforeNextRelease
-Same `-- SOLUTION` mishandling as elsewhere in this chapter, milder here: the
-block keeps surviving content (`attribute [irreducible] …`) after
-`-- END SOLUTION`, so student/terse don't error, but the *solutions* build
-leaks the literal `-- SOLUTION` / `-- END SOLUTION` comment lines into the
-displayed code. Prefer `solution!` over the comment markers.
-:::
 ```lean
 -- SOLUTION
 /- normalize_test_zero -/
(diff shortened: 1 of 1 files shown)


LF/Induction.lean
```diff
@@ -1485,26 +1485,17 @@ It would be wise to do some `example` proofs to check that your
...
 /- normalize_test_2 -/
-example : normalize (.b0 .z) = .z := by rfl
+example : normalize (.b0 .z) = .z := solution!(by rfl)
 /- normalize_test_3 -/
-example : normalize (.b0 (.b0 .z)) = .z := by rfl
+example : normalize (.b0 (.b0 .z)) = .z := solution!(by rfl)
 /- normalize_test_4 -/
-example : normalize (.b1 (.b0 .z)) = .b1 .z := by rfl
--- END SOLUTION
+example : normalize (.b1 (.b0 .z)) = .b1 .z := solution!(by rfl)
```
berberman: I think we are using `SOLUTION` markers here intentionally as we want the students to fill out the definitions as well.

LF/Induction.lean
```diff
@@ -1524,40 +1515,33 @@ its own inductive proof — that will allow the main proof to make
...
     incr (doubleBin b) = .b1 b := by
-  cases b with
-  | z    => rw [doubleBin_z, incr_z]
-  | b0 n => rw [doubleBin_b0, incr_b0]
-  | b1 n => rw [doubleBin_b1, incr_b0]
+  solution!
+    cases b with
+    | z    => rw [doubleBin_z, incr_z]
+    | b0 n => rw [doubleBin_b0, incr_b0]
+    | b1 n => rw [doubleBin_b1, incr_b0]
```
berberman: Similarly, students are supposed to come up with the definition of this lemma.

berberman (commented):
I was reviewing on my phone and could be wrong, so please double check the surrounding text to make sure that we don't accidently drop `--SOLUTION`.

Robertboy18:
You're right. I checked `STYLE.md` and the original Rocq chapter, and these markers intentionally hide the whole declarations so students can write them. `solution!` would expose the statements. I'll restore the markers and keep only the removal of the obsolete developer notes. Thanks for catching this.

LF/Induction.lean
Robertboy18: You are right. I restored this whole block to the `SOLUTION` markers, so the student and terse outputs still replace it with `-- FILL IN HERE`. The PR now only removes the stale developer note.

LF/Induction.lean
Robertboy18: Agreed. I restored these markers too, so students still have to supply both declarations. The generated student and terse files contain `-- FILL IN HERE`; this PR now only removes the obsolete note.

berberman (approved):
Thanks for the cleanup. (The solution marker leakage was fixed in #163)

PR #191, opened 2026-08-04
Editing pass over Typeclasses

mwhicks1:
A nearly full pass over typeclasses led me to make some adjustments to presentation, mostly small ones at sentence level, but sometimes by adding a little more Lean code or slightly adjusting the order it is presented. Specifics:
- A bit more careful comparison between `List.elem_poly` (with and without typeclasses) and `List.elem_poly_eq` 
- Adjusted the presentation of `GetElem` to be a little bit gentler, justifying why we'd want a typeclass for added syntax (we don't use one for `update`), and presenting the implementation in clearer steps.
- Added more markup and references (including a `deftech` to the Logic chapter -- very small chnage there)
- A couple of bigger comments for later consideration:
- 1/ Is `HasOne` the best example? It connects to `Inhabited` which is nice, but I think it's hard to understand to understand in a vacuum. I'm not convinced the lead-in connection to `structure` is helpful either (it's a mechanism, which is not so interesting when you are learning a new concept, only after you have a sense of it, to strengthen what you've learned)
- 2/ Will we be able to fix the typeclass resolution error at the outset? If not, we need to change some later text not to refer to it.


LF/Typeclasses.lean
```diff
@@ -78,10 +78,14 @@ polymorphism suggests simply replacing `Nat` with a type variable `α`, but that
 error:
 
 :::dev
-@dsainati - The Verso compilation is not actually removing this from the generated file despite th e
+@dsainati - The Verso compilation is not actually removing this from the generated file despite the
 -keep flag. It is, however, stripping the message guard, so this results in an error. Once
 we figure out how we are handling these cases, uncomment this.
 
+mwhicks1: Until then, readers reach the sentence below — "that produces a puzzling error" —
```
chenson2018: The current setup works in the sense that, by having `-keep` with a `#guard_msgs` we see the error written out in both both the extracted file and the HTML. I'm still not happy with this, because what we really want here is `-keep +error`, so that in the HTML that this appears just like an ordinary error, with the advantage that it will exactly highlight where the error occurs. I gather we don't do this because then in the extracted file we don't see the error message.

cc @berberman Can you confirm that this isn't user error on my part given recent changes of yours? If not, could we improve this so that `-keep +error` also inserts a `#guard_msgs` in the extracted version? (This is the sort of thing I'm proposing we bring up in the meeting with David, this feels entirely too complex)

LF/Typeclasses.lean
```diff
@@ -160,8 +171,17 @@ to define a typeclass in the first place.
 `BEq` comes from Lean's standard library. Let's define a typeclass of our own, to see the
 mechanism — classes, instances, and synthesis — that made `==` resolve automatically above.
 
-Suppose we want to specify that a type has at least one inhabitant, i.e., that it is not empty. A
-`structure` (chapter {ref "Lists"}[Lists]) can express this directly:
+:::dev "mwhicks1" PotentialImprovement
+As a programmer, I wouldn't imagine defining a `structure` to "specify" that a type has at
+least one inhabitant — that only makes sense as a constraint on something else. For example,
+if I'm defining a map and I want a default value, then requiring an element of the value's
+type as a parameter implicitly states that there is at least one; it doesn't work in a
```
chenson2018, line 165: I don't think I follow. To me it is very intuitive that this represents a specification of having an inhabitant,  e.g. as a proposition `Nonempty α ↔ ∃ a : α, True`.  Just like every type, that specification is not satisfied until we have an inhabitant (an instance). The parameter analogy you make is exactly what `Inhabited.mk (default : α) : Inhabited α` is, which we use we declaring an instance. Maybe just spelling this out makes fills the gap here?

I don't have a better example in mind than this for a first typeclass to explain, but am not attached to what I wrote here. If you have an example from a Rocq resource that you feel is smoother I have no objection, but it seemed like the simplest example I could think of.

LF/Typeclasses.lean
```diff
@@ -160,8 +171,17 @@ to define a typeclass in the first place.
 `BEq` comes from Lean's standard library. Let's define a typeclass of our own, to see the
 mechanism — classes, instances, and synthesis — that made `==` resolve automatically above.
 
-Suppose we want to specify that a type has at least one inhabitant, i.e., that it is not empty. A
-`structure` (chapter {ref "Lists"}[Lists]) can express this directly:
```
chenson2018, line 164: General comment on:

>  I'm not convinced the lead-in connection to structure is helpful either (it's a mechanism, which is not so interesting when you are learning a new concept, only after you have a sense of it, to strengthen what you've learned)

My point in mentioning the connection to `structure` is to demystify typeclasses a bit by showing they are (pretty much) something we've already worked with. My experience has been that students can go a startling amount of time without making this connection, and because there are so many different ways to declare a structure/instance that they just memorize these and are generally very confused that these are just ordinary terms. I just want a big flashing sign that says "hey there is no magic here, you're just defining structures and the only difference is that Lean will sometimes try to infer them!". 

You have a lot more experience than me teaching this course though: if you think that this should be moved, altered, or removed entirely, please go for it.

LF/Typeclasses.lean
```diff
@@ -685,16 +730,16 @@ example : exampleMap'["quux"] = false := rfl
...
 Even if you don't work the following exercises, make sure you thoroughly understand the statements of the lemmas!
 
-(Some of the proofs require the functional extensionality axiom, which was discussed in the Logic chapter.)
+(Some of the proofs require the functional extensionality axiom {name}`funext`, discussed in the {ref "Logic"}[Logic] chapter.)
 
 First, the empty map returns its default element for all keys:
 
 ```lean
-theorem getElem_empty (a : α) : (∅ : TotalMap α β)[a] = default := rfl
+theorem getElem_empty (a : α): (∅ : TotalMap α β)[a] = default := rfl
```
chenson2018: This space should not be removed

```suggestion
theorem getElem_empty (a : α) : (∅ : TotalMap α β)[a] = default := rfl
```

LF/Typeclasses.lean
```diff
@@ -178,8 +198,8 @@ def nat_hasOneStruct : HasOneStruct Nat where
 example : nat_hasOneStruct.one = 1 := rfl
 ```
 
-But nothing makes Lean produce this witness *automatically*: we had to write `⟨1⟩` ourselves,
-unlike the `BEq Nat` instance that Lean found for us on its own above.
+But nothing makes Lean produce this witness _automatically_: we had to write `where one := 1`
+ourselves, unlike the `BEq Nat` instance that Lean found for us on its own above.
```
chenson2018, line 189: This line has bugged me a bit since the previous PR, because the equivalent for instances isn't automatic either. If I were a beginner I might be confused at being told typeclasses will solve this problem then writing the exact same line `where one := 1` in an instance a bit below. I think your example of passing around `eq` better demonstrates the point, while this blurs the line between writing and inferring instances.

chenson2018 (commented):
I mostly just tried to respond to your bigger questions and skim for any big issues. (I didn't really see many, these look like improvements to me)

LF/Typeclasses.lean
berberman: I'm confused by the dev comments here -- they look outdated to me. The current policy for project extraction is:

- `-keep` -> `sf_experiment`
- `+error` -> `sf_expect_failure`
- `-keep +error` -> `sf_expect_failure`

`sf_expect_failure` works for the combination because it always reverts the environment (which slightly diverges from verso's semantic but shouldn't be harmful.) No `#guard_msgs` needed. If something doesn't compile, then there's a regression.

For this `:::dev` wrapped code block, I believe

````lean
```lean -keep +error
def List.elem_poly {α : Type} (a : α) (xs : List α) : Bool :=
  match xs with
  | [] => false
  | b :: tl => bif a == b then true else elem_poly a tl
```
````

should work.

**Comment by KihongHeo** [comment]
Hi. I will also review the file today. BTW, I got the following error when running `make serve`
```
sf-in-lean/LF/Typeclasses.lean:498:133: No term def with key "extensional"
```
Anyone has the same problem?

bcpierce00:
I seem to have been broken this PR -- sorry and will fix tomorrow...

LF/Typeclasses.lean
xhalo32: I agree with Chris that `structure` is often the right construction for a "specification". I like to think of structures as sigma types which in Prop coincide with existence.

We already discuss `struct`-like structures (like in other programming languages like Rust) in Basics.lean (`NibbleStruct)` and Lists.lean (`NatProd'`). `HasOneStruct` makes perfect sense to me as a specification structure, and the way the narrative continues into `class HasOne` feels very natural.

bcpierce00:
OK, the PR is fixed.  It wasn't actually me that broke it -- the {tech} stuff was broken because the definition it was referring to is in another file.  For the moment, I've just converted to italics, since this does not seem trivial to fix.

LF/Typeclasses.lean
mwhicks1: What I am saying is this: A `structure` is just a type definition. In this case, the type is a one-field structure around a polymorphic type variable `α`. From a type-theoretic perspective this is the same as just the type `α` itself. To prove that a type `α` is inhabited you have to produce a value of type `α`. Wrapping `α` in a structure doesn't change that: you just now have to produce a structure containing the `α` value. In sum: the narrative makes it seem like using a `structure` is solving a problem, but it isn't.

For that matter, even typeclasses are not directly solving the problem: The search mechanism is going to locate an instance (a structure) you had to create. In some cases those instances might get created automatically through other typeclass specifications, but those other specifications you had to write (transitively), I think.

In sum: A typeclass is a kind of structure whose values (instances) can be filled in automatically by Lean if they exist. Like in the BEq example: You don't want to have to find and fill in `eq` functions. You want Lean to do that for you.

For that to work, you need to be able to specify what you want Lean to look for. Structures are a nice, uniform way to do that. In particular, the "thing" you are looking for might be multiple things, which you can package together in a structure (what's a good example of a multi-field typeclass?). Moreover, the thing you want may have semantic meaning not immediately evident from the type -- the name of the structure can convey that.

Having written all this out, I'm wondering whether dumping `HasOne` in favor of just using our own `BEq` would make sense. It's not very much more complicated and with the recent revisions it is well set-up.

Thoughts?

LF/Typeclasses.lean
mwhicks1: I am suggesting a reordering: Introduce `class` first to solve the problem we set out: Finding values like `eq` automatically. After introducing that, it's a fast-follow (or even during the presentation) that we say that `class` is just a structure that is subject to search via the typeclass mechanism, whereas normal structures are not. So we stay on point in terms of the problem we have laid out, but when introducing the solution we connect to something they have already seen.

LF/Typeclasses.lean
mwhicks1: This is part of the issue I'm driving at in my earlier comments: structures do not solve the problem we care about, which is to find an instance for us automatically. So this part of the text is hard to understand.

LF/Typeclasses.lean
chenson2018: Mike, I think if this conversation were happening synchronously you'd find we're (mostly) on the same page. Let's discuss tomorrow.

This all seems orthogonal to the question "is `Inhabited` a good first example". If you want to substitute in more detailed explanation of `BEq` followed up by `LawfulBEq`, this to me seems to cover the same concepts, and might make more sense to lead with given how this chapter has shuffled around. It also seems like we still need examples/exercises though, maybe keep the `Has*` stuff for this at the end?

LF/Typeclasses.lean
chenson2018: I think I just disagree here. See the first section of the [TPIL chapter on typeclasses](https://leanprover.github.io/theorem_proving_in_lean4/Type-Classes/#type-classes) for comparison.

LF/Typeclasses.lean
mwhicks1: Sounds good. Thanks for all this engagement! I have tried to put my prescriptions into action in a separate PR on top of this one. I'm keeping `HasOne` but giving it a use-case that motivates a slightly different name. See https://github.com/plclub/sf-in-lean/pull/198. I feel like we are pretty close to the right place now.

LF/Typeclasses.lean
mwhicks1: In my PR https://github.com/plclub/sf-in-lean/pull/198 I kept your ordering. I think it works.

LF/Typeclasses.lean
mwhicks1: I believe that my rework in PR https://github.com/plclub/sf-in-lean/pull/198 has addressed this point.

bcpierce00 (commented):
OK, finished my pass. This chapter still needs a lot of work, but we should move toward merging these changes, which are taking us in the right direction!

LF/Typeclasses.lean
bcpierce00: Sounds like this conversation has converged and we should go ahead and merge both this and #198?

bcpierce00 (approved):

LF/Typeclasses.lean
mwhicks1: Sure! If you look and approve that one, let's merge them both. As you say, there's more work to do here but this might be a good place to pause for now.

bcpierce00 (approved):

bcpierce00 (approved):

bcpierce00 (approved):

PR #192, opened 2026-08-04
Qualified import with `open` and parentheses for handling `||` precedence

yiyunliu:
The PR bundles two minor fixes.

- Basics.lean uses `export` for qualified imports even though `open` already has the qualified import form. The PR replaces all mentioning of `export` in Basics.lean with `open (...)`.
- In Lean, `||` has lower precedence than `=`. One of the exercises has the form `a || b = true` and it becomes `a || decide (b = true)`, where Lean auto-inserts `decide` to convert the `b = true` `Prop` into `bool`. There might be more instances of this problem, so it could be useful to ask Claude to hunt for similar issues

### DIFF
--- a/LF/Basics.lean
+++ b/LF/Basics.lean
@@ -457,7 +457,7 @@ Now that we've defined some basic functions on booleans, let's see how to
 _prove_ some simple properties of those functions. Here is a simple rule
 about `&&`:
 
-- `MyBool.true && b = b`
+- `(MyBool.true && b) = b`
 
 This is an example of a _proposition_, a logical _claim_ that we can try to prove.
 It says that `MyBool.true && b` is equal to `b` for every `MyBool` `b`.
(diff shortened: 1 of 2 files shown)


bcpierce00:
Claude found no more instances of this problem in checked Lean code, but there were a couple of instances in text:

[LF/Basics.lean:460](vscode-webview://1n70lfuqe4irh6agi646e7nmcc4au4ks2pjif94m7hsr708njp8f/LF/Basics.lean#L460) — the informal proposition `MyBool.true && b = b` is a plain (non-elaborated) monospace span; the very next lines restate it and [line 467](vscode-webview://1n70lfuqe4irh6agi646e7nmcc4au4ks2pjif94m7hsr708njp8f/LF/Basics.lean#L467) gives the properly-parenthesized Lean (MyBool.true && b) = b.

[LF/Basics.lean:2607](vscode-webview://1n70lfuqe4irh6agi646e7nmcc4au4ks2pjif94m7hsr708njp8f/LF/Basics.lean#L2607) and [2614](vscode-webview://1n70lfuqe4irh6agi646e7nmcc4au4ks2pjif94m7hsr708njp8f/LF/Basics.lean#L2614) — h : true && c = true || c inside proof-narration comments (informal).

Want to give these a quick fix also?

bcpierce00 (approved):

yiyunliu:
> Want to give these a quick fix also?

sure! will push the fixes once i'm back to my laptop

berberman (approved):
LGTM. To give a bit of context, initially I brought up `export` as I wanted to explain why `true`/`false` can be referred without `open Bool`; but it seems that we got confused in #64 and described `export` as the command that introduces names into the current namespace. I agree `export` should be avoided in this book at all.

yiyunliu:
pushed the two fixes @bcpierce00 suggested

Issue #165, opened 2026-07-28
Write Preface.lean and Postscript.lean in branch `add-preface-postscript`


rogerburtonpatel:
This is a **writing-only** task; there is no programming involved. 
**Lean knowledge (or availability to gain it) is needed to complete this task well.** 

The task is to write the Lean-specific versions of `Preface.v` and `Postscript.v`. These are text-only files that live in the current repository and describe the utility and history of Rocq, as well as the directions of the software foundations series. Naturally, these files need to be ported to discuss Lean. 

To begin tackling this port, I rewrote the files to Verso format. I realized then that the text of the chapter would be better written by a long-time Lean user. 

To complete this, switch to branch `add-preface-postscript`, locate `Preface.lean` and `Postscript.lean`, rewrite them to match the history and philosophy of Lean.

ionathanch:
I assume also closed by #227, as mentioned in #34

Issue #147, opened 2026-07-25
Bib chapters — porting plan and open questions


bcpierce00:
We need to convert the Bib files from the original SF Logical Foundations and Programming Language Foundations volumes. We should do it the "Verso way."

### Original SF source

`old/orig-lf-files/Bib.v` is a short, hand-written list (~6 entries), **not** a real BibTeX file. Each entry uses a custom `\BIB{key}` macro, e.g.:

```
\BIB{Wadler 2015} Philip Wadler. "Propositions as types."
  Communications of the ACM 58, no. 12 (2015): 75-84.
  {https://dl.acm.org/citation.cfm?id=2699407}
```

The PLF volume (`old/orig-plf-files/Bib.v`) has a similar handful of entries.

### How citations work in Verso (v4.32.0)

Verso's Manual genre has a built-in citation system, in `.lake/packages/verso/src/verso-manual/VersoManual/Bibliography.lean`:

- **References are Lean values, not a bib file.** Verso's own docs describe the module holding them as playing "the role of a `.bib` file in LaTeX". You write `def`s in a module that is imported where needed, e.g.:

  ```lean
  def wadler2015 : Article where
    title   := inlines!"Propositions as Types"
    authors := #[inlines!"Philip Wadler"]
    journal := inlines!"Communications of the ACM"
    year    := 2015
    month   := none
    volume  := inlines!"58"
    number  := inlines!"12"
    pages   := some (75, 84)
    url     := some "https://dl.acm.org/citation.cfm?id=2699407"
  ```

- There are exactly **four entry types**: `InProceedings`, `Thesis`, `ArXiv`, `Article`. Fields are Verso inline elements (built with `inlines!"..."`), not plain strings, so they may contain markup.
- **Cite in prose** with three roles:
  - `{citet foo}[]` — textual, renders "Author (Year)"
  - `{citep foo}[]` — parenthetical, renders "(Author, Year)"
  - `{citehere foo}[]` — full inline citation

  By default a cite renders as a **margin note** (Marginalia); `citehere` inlines the full reference instead.
- **The reference list is auto-generated**: the traverse pass collects every cited entry into the `Manual.Bibliography` state, deduplicated and sorted by author/year, so only works actually cited appear. Both HTML and TeX/PDF output are supported.

### No `.bib` importer

[...]

PR #195, opened 2026-08-04
Convert bibliography for Verso

luisacicolini:
Addresses #147 - the bibliography is converted to Verso. 

When #165 is also addressed we will be able to directly use these citations (afaiu they are basically only used there). 

Co-authored by Claude.


Bib.lean
```diff
@@ -0,0 +1,143 @@
...
+namespace SFLMeta
+
+/-!
+Shared bibliography entries for all SF-in-Lean volumes, transcribed from the
+original SF sources `old/orig-lf-files/Bib.v` and `old/orig-plf-files/Bib.v`.
+-/
+
+namespace Bib
+
+def aydemir2008 : InProceedings where
```
chenson2018, line 14: Can we add one of these references somewhere to confirm that they render properly? Even an LF preface that is empty other than a single reference seems fine.

SFLMeta/Bib.lean
```diff
@@ -0,0 +1,143 @@
...
+
+def bertot2004 : Article where
+  title   := inlines!"Interactive Theorem Proving and Program Development: Coq'Art: The Calculus of Inductive Constructions"
+  authors := #[inlines!"Yves Bertot", inlines!"Pierre Castéran"]
+  journal := inlines!"Springer-Verlag"
+  year    := 2004
+  month   := none
+  volume  := inlines!""
+  number  := inlines!""
+  pages   := none
```
chenson2018: `pages := none` is redundant since `Article` has this as a default value.

SFLMeta/Bib.lean
```diff
@@ -0,0 +1,143 @@
...
+  url     := some "https://tinyurl.com/z3o7nqu"
```
chenson2018: You should be able to omit all the `some` for readability since there is a coercion from `α` to `Option α`

chenson2018 (commented):
Very small nitpicks, LGTM otherwise.

luisacicolini:
thanks @chenson2018! All addressed :)

Bib.lean
chenson2018: @luisacicolini Does the reference you added actually appear the preview?

chenson2018 (approved):
LGTM, though I left one more question about the example reference.

bcpierce00:
Does Bib.lean have to live in SFLMeta?  Could it live at the top level instead, for example?

bcpierce00 (approved):

Bib.lean
luisacicolini: still working on it - I just pushed some changes, hopefully that works

bcpierce00 (approved):

bcpierce00 (approved):

PR #198, opened 2026-08-05
Typeclasses: motivate HasOne/DefaultValue with a real consumer, headOr

mwhicks1:
- Give the class a use: List.headOr_ex/List.headOr rewrite an explicit defaultValue parameter into an instance implicit, mirroring elem_poly_eq/elem_poly.
- Renamed HasOne (and the throwaway HasOneStruct) to DefaultValue, matching the concept to its motivating use.
- Spelled out, in prose, that getting Lean to fill in the parameter automatically takes two separate things: marking it as searchable (an instance implicit), and giving Lean something well-defined to search for (naming the concept via a structure/class).

Separate improvement:
- When introducing MyGetElem, pointed out that `==`/BEq already overloads notation via instance search the same way `m[a]`/MyGetElem is about to.

Text constructed in collaboration with Claude.

### DIFF
--- a/LF/Typeclasses.lean
+++ b/LF/Typeclasses.lean
@@ -172,22 +172,26 @@ def List.headOr_ex {α : Type} (defaultValue : α) (xs : List α) : α :=
 #eval ([] : List Nat).headOr_ex 0
 ```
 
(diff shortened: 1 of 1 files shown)


bcpierce00 (commented):
Looks great.  Please merge!  (I left some comments too, naturally...)

bcpierce00 (approved):

PR #206, opened 2026-08-07
Get ready for alpha testers

bcpierce00:
At the team meeting this week, we agreed on the desirability of increasing the pool of readers / alpha-testers for SFL.  With this in mind, I've added an ALPHATESTERS.md at the top level, changed the README to point to it, and removed all non-NOW :::dev comments from the student version so that they don't bother most readers.

### DIFF
--- a/ALPHATESTERS.md
+++ b/ALPHATESTERS.md
@@ -1,0 +1,19 @@
+# SF-in-Lean: Instructions for Alpha-Testers
+
+- clone the [repo](https://github.com/plclub/sf-in-lean)
+- type `make student` in the top-level directory
+- open _out/lf/student/html-multi/index.html
+- follow the installation instructions in the `Preface` to install VSCode and Lean if necessary
+- start reading!
+
+If you see clear improvements (typos, obvious clarifications, etc.), please feel free to just make them directly to the `.lean` files in the `LF`, `HL`, or `TS` subdirectories
+
+If you have more substantive remarks, please add them to the source code in the form of `dev` comments, like this:
+```
+    :::dev "Your Real Name (@your_github_handle)"
+    ... Your suggestions ...
+    :::
+```    
+Make a github PR for each chapter as you finish reading it.  
+
+We are actively working on all chapters, so if you're reading a chapter over a long period, make sure to pull changes often.
--- a/README.md
+++ b/README.md
@@ -3,36 +3,32 @@
 This repository contains the work-in-progress sources for [Software
 Foundations](https://softwarefoundations.cis.upenn.edu/) in Lean.
 
-## Status
+## Status and how to contribute
 
-SF-in-Lean is _not_ ready for ordinary readers yet.  We'll make a
-posting on the Lean Zulip when we've got something suitable for public
-consumption. We aim to have a complete and polished draft of most
+SF-in-Lean _is_ ready for adventurous alpha-testers.  See ALPHATESTERS.md for more.
+
+The SF-in-Lean team is also looking for contributors who have the time and interest to make a more significant commitment.  If you are interested in joining us, please email [Benjamin Pierce](bcpierce@cis.upenn.edu).
+
+SF-in-Lean is _not_ ready for ordinary readers yet. We aim to have a 
+complete and polished draft of most
 chapters of _Logical Foundations_ and _Programming Language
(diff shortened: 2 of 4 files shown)


ALPHATESTERS.md
```diff
@@ -0,0 +1,19 @@
...
+- type `make student` in the top-level directory
+- open _out/lf/student/html-multi/index.html
+- follow the installation instructions in the `Preface` to install VSCode and Lean if necessary
+- start reading!
+
+If you see clear improvements (typos, obvious clarifications, etc.), please feel free to just make them directly to the `.lean` files in the `LF`, `HL`, or `TS` subdirectories
+
+If you have more substantive remarks, please add them to the source code in the form of `dev` comments, like this:
+```
+    :::dev "Your Real Name (@your_github_handle)"
```
Robertboy18, line 13: Should this omit the `@`, or should `devNoteAuthorText` accept both forms? It adds `@` during rendering, so this example currently produces `@@your_github_handle`.

README.md
```diff
@@ -3,36 +3,32 @@
...
-posting on the Lean Zulip when we've got something suitable for public
-consumption. We aim to have a complete and polished draft of most
-chapters of _Logical Foundations_ and _Programming Language
-Foundations_ in time for Fall 2026 courses. 
+SF-in-Lean _is_ ready for adventurous alpha-testers.  See ALPHATESTERS.md for more.
 
-If you would like to be notified when chapters are ready for
-alpha-testing, please email [Benjamin
-Pierce](mailto:bcpierce@cis.upenn.edu).
+The SF-in-Lean team is also looking for contributors who have the time and interest to make a more significant commitment.  If you are interested in joining us, please email [Benjamin Pierce](bcpierce@cis.upenn.edu).
```
Robertboy18, line 10: Could this keep the `mailto:` prefix? As written, clicking Benjamin’s name follows a relative link instead of opening an email.

README.md
```diff
@@ -3,36 +3,32 @@
...
+chapters of _Logical Foundations_ and _Programming Language
+Foundations_ in time for Fall 2026 courses. We'll make a
+posting on the Lean Zulip when appropriate.
 
 ## Quick start: building and viewing the book
 
 To build everything and preview the HTML locally:
 
     make serve
+    open http://localhost:8000
```
Robertboy18, line 23: Could we keep this instruction as prose, as before? `make serve` stays in the foreground, so this command cannot run afterward in the same terminal, and `open` is macOS-specific.

README.md
```diff
@@ -3,36 +3,32 @@
...
+The SF-in-Lean team is also looking for contributors who have the time and interest to make a more significant commitment.  If you are interested in joining us, please email [Benjamin Pierce](bcpierce@cis.upenn.edu).
 
-We are also not set up yet to consider PRs from outside the
-translation team.  If you are interested in joining the team, please
-email Benjamin Pierce and we can discuss.
+SF-in-Lean is _not_ ready for ordinary readers yet. We aim to have a 
+complete and polished draft of most
+chapters of _Logical Foundations_ and _Programming Language
+Foundations_ in time for Fall 2026 courses. We'll make a
+posting on the Lean Zulip when appropriate.
```
mwhicks1, line 16: I don't think we will have a polished draft of most LF chapters by the time we start. Sad to say. Rather, we will have a pretty-good draft of it which we will refine as we teach the class. Should we adjust what we say here?

README.md
mwhicks1: I don't think that omitting the `mailto` is going to stop spam, if that's what you are worried about, @bcpierce00 . :)

ALPHATESTERS.md
mwhicks1: I see people writing `@` throughout the text now. Maybe we stick with that pattern for uniformity; perhaps we can change the rendering code instead?

mwhicks1 (approved):

ALPHATESTERS.md
bcpierce00: Fixed rendering to accept both

README.md
bcpierce00: No, it was just a typo

README.md
bcpierce00: OK.

README.md
bcpierce00: OK

PR #210, opened 2026-08-07
Induction polish

mwhicks1:
Full pass through Induction:
- Switched to fully using declaration-header style (as was promised by Basics)
- Some reformatting to help the linear flow
- Lots of Verso annotations
- Sentence level tweaks

I also merged with `claude/induction-review-fixes` so those changes are in too.

### DIFF
--- a/LF/Basics.lean
+++ b/LF/Basics.lean
@@ -2229,6 +2229,9 @@ The same statement can be written using an explicit universal quantifier, as we 
 mul_zero : ∀ (n : Nat), n * zero = zero
 ```
 
+Writing statements in declaration-header style shortens proofs because Lean
+automatically adds declared variables to the context, rather than requiring
+them to be added with `intro`.
 The declaration-header style is conventional in Lean, and we will generally use it from now on.
 ::::
 
@@ -2272,7 +2275,6 @@ Sometimes simple calculation and rewriting are not enough...
 
 ```lean +error
 example (n : Nat) : (succ n == zero) = false := by
-  intro n
   /-
     We can't rewrite by any lemmas here because `n` is unknown!
   -/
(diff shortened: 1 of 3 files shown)


LF/Basics.lean
```diff
@@ -2229,6 +2229,9 @@ The same statement can be written using an explicit universal quantifier, as we
 mul_zero : ∀ (n : Nat), n * zero = zero
 ```
 
+Writing statements in declaration-header style shortens proofs because Lean
+automatically adds declared variables to the context, rather than requiring
+them to be added with `intro`.
```
berberman:
```suggestion
them to be added with {tactic}`intro`.
```

LF/Induction.lean
```diff
@@ -915,26 +901,17 @@ respectively, which are equal by the induction hypothesis.
 _Qed_.
 ::::
 
-::::hide
-```
- MMG: the proof above makes no use of lemmas, so it's hard for
```
Robertboy18, line 920: Was this note meant to be resolved here? The example above still only uses the definition of `+` and the induction hypothesis, while `add_comm_informal` asks students to use earlier lemmas such as `zero_add` and `succ_add`. It might help to show one example of that first.

mwhicks1: I just decided that this (showing the use of lemmas in informal proofs) was not that important, i.e., it is easy enough to generalize your thinking from what's there.

Robertboy18: Makes sense, thanks!

LF/Induction.lean
```diff
@@ -356,7 +355,7 @@ there if we just go on like this.
 ::::full
 To prove interesting facts about numbers, lists, and other
 inductively defined sets, we often need a more powerful reasoning
-principle: _induction_.
+principle: *induction*.
```
bcpierce00: I changed boldface back to italic here, but I'm happy to discuss using boldface in a principled and consistent (and sparing) way...

bcpierce00 (approved):

PR #212, opened 2026-08-08
LF: introduce bif before first use

### LINKED ISSUE #205: `bif` is never introduced
`bif` is used in the solution for the `oddmembers` exercise in `Lists.lean`, but I believe up until that point it's never been introduced. I cross checked with an LLM and apparently it was introduced in Basics.lean in the older version but somehow got removed.

Robertboy18:
## Summary

I ran into `bif` for the first time in the `oddMembers` exercise and realized the syntax had not been introduced. This adds a short explanation immediately before that definition, including what happens for `true` and `false`.

Closes #205.

### DIFF
--- a/LF/Lists.lean
+++ b/LF/Lists.lean
@@ -558,6 +558,10 @@ theorem test_nonZeros : nonZeros [0, 1, 0] = [1] := by
 :::gradeTheorem "0.5" test_nonZeros
 :::
 
+The next definition uses `bif`, Lean's conditional for Boolean tests.
+The expression `bif b then x else y` evaluates to `x` when `b` is
+`true` and to `y` when `b` is `false`.
+
 ```lean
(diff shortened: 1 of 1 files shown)


LF/Lists.lean
```diff
@@ -558,6 +558,10 @@ theorem test_nonZeros : nonZeros [0, 1, 0] = [1] := by
 :::gradeTheorem "0.5" test_nonZeros
 :::
 
+The next definition uses `bif`, Lean's conditional for Boolean tests.
+The expression `bif b then x else y` evaluates to `x` when `b` is
+`true` and to `y` when `b` is `false`.
```
berberman:
```suggestion
{name}`true` and to `y` when `b` is {name}`false`.
```

berberman (approved):

mwhicks1 (approved):

PR #213, opened 2026-08-08
Basics: rework the last exercise

berberman:
The previous exercise was a bit of awkward with large numeric literals and  techniques we haven't introduced yet such as `Ne`, `dsimp`, `<;>`, etc.

This PR replaces it with a simpler exercise -- students define and test a small set of airport operations, write characterization lemmas, and then use those lemmas in larger proofs. I have tried making one about practicing more on structures, but it turned out to be awkward because the claims couldn't be stated nicely -- they would need the proofs to use `dsimp`/`contradiction` for operating equational hypotheses and structure projections.

I used gpt-5.6 to revise the wording.

### DIFF
--- a/LF/Basics.lean
+++ b/LF/Basics.lean
@@ -2706,17 +2706,18 @@ Add grading attributes.
 Now that we have learned the basic features of Lean 4, let's close the chapter
 with an exercise that brings them together.
 
-One fascinating aspect of using a theorem prover like Lean is that we can do
-more than implement a system — we can also state precisely what we expect the
-system to behave and prove that our implementation satisfies those expectations.
+In a theorem prover like Lean, we can do
+more than just implement a system — we can also state precisely how we expect the
+system to behave and prove that our implementation satisfies that expectation.
 
-In this exercise, we will model part of an airport database.
-The airport stores one database entry for each traveler.
-An entry records where the traveler is in the airport process,
-together with information about their current bag.
+In this exercise, we will model part of a database
+storing information about travelers passing through an airport.
+The database contains one entry per traveler, recording
+information about where the traveler is in the airport process and the contents of their luggage.
 
-We will implement several operations on these entries and prove
-general properties about their behavior.
+We will implement several operations on these entries, state
+properties that describe how the database should
+behave, and prove that the implementation satisfies them.
 :::
 
 :::terse
@@ -2730,6 +2731,8 @@ namespace Airport
 ```
 
 :::full
+First, we describe the possible states of a bag.
+
 For simplicity, we assume that a bag either contains a battery, which causes
 it to fail inspection, or contains only ordinary items:
 :::
@@ -2745,7 +2748,7 @@ inductive BagContent : Type where
 ```
 
 :::full
-The screening status records whether the bag has not yet been inspected,
+A bag's screening status records whether the bag has not yet been inspected,
 has been cleared, or has been blocked:
 :::
 
@@ -2761,7 +2764,7 @@ inductive ScreeningStatus : Type where
 ```
 
 :::full
-A traveler can be in one of three stages:
+Next, we consider the possible stages of the airport process a traveler can inhabit:
 - they have not yet purchased a ticket;
 - they have a ticket but have not yet checked in;
 - they have checked in, in which case the
@@ -2811,12 +2814,15 @@ theorem buyTicket_checkedIn (bagContent : BagContent)
 attribute [irreducible] buyTicket
 ```
 
-Here is our first general property: buying a ticket twice has the same
-effect as buying it once.
-
 :::full
-N.B. An operation is called _idempotent_
-if performing it twice has the same effect as performing it once.
+An operation is called _idempotent_
+when performing it twice has the same effect as performing it once.
+The first property we will prove about our system is that
+purchasing a ticket is an idempotent operation.
+:::
+
+:::terse
+Here is our first general property: buying a ticket twice is idempotent.
 :::
 
 :::exercise (rating := 2) (name := "buy_ticket_idempotent")
--- a/LF/Basics.lean
+++ b/LF/Basics.lean
@@ -2764,7 +2764,7 @@ inductive ScreeningStatus : Type where
 A traveler can be in one of three stages:
 - they have not yet purchased a ticket;
 - they have a ticket but have not yet checked in;
-- they have checked in,in which case the
+- they have checked in, in which case the
   database also stores the screening status of their bag.
 :::
 
@@ -2840,7 +2840,7 @@ theorem buyTicket_idempotent (t : Traveler) :
 ```
 :::
 
-A traveler can check in only after buying a ticket.
+A traveler can check in only after buying a ticket,
 and their bag is marked as needing inspection.
 Calling checkIn in any other state does nothing.
 
--- a/LF/Basics.lean
+++ b/LF/Basics.lean
@@ -1372,6 +1372,16 @@ inductive Nat : Type where
   | succ (n : Nat)
 ```
 
+The following lines make ordinary numerals such as 0, 1, and 2 work with our {name}`Nat` type.You can ignore the details for now.
+
+```lean
+def ofNat : _root_.Nat → Nat
+  | .zero => .zero
+  | .succ n => .succ (ofNat n)
+
+instance (n : _root_.Nat) : OfNat Nat n := ⟨ofNat n⟩
+```
+
 :::full
 We'll define some shorthands for numbers, putting them in the `Nat` namespace
 so we don't need to use `.` notation everywhere.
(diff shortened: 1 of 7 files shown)


LF/Basics.lean
```diff
@@ -1372,6 +1372,16 @@ inductive Nat : Type where
   | succ (n : Nat)
 ```
 
+The following lines make ordinary numerals such as 0, 1, and 2 work with our {name}`Nat` type.You can ignore the details for now.
+
+```lean
+def ofNat : _root_.Nat → Nat
```
xhalo32, line 1389: Can we avoid `_root_`? Not sure if it's used elsewhere in the file but I think it's better to `end+namespace` instead.

LF/Basics.lean
```diff
@@ -2676,358 +2686,330 @@ theorem and_eq_or (b c : Bool) : (b && c) = (b || c) → b = c := by
...
-example : letterComparison B F = gt := by rfl
-```
-
-::::exercise (rating := 1) (name := "letter_comparison")
-```lean
-theorem letterComparison_Eq : ∀ l : Letter,
-    letterComparison l l = eq := by
-  solution!
-    intro l; cases l <;> rfl
+The simplification rules for {name}`buyTicket`:
```
xhalo32: Is "simplification" the best term here?

xhalo32: I don't remember if we have introduced characterizing lemmas before this point

LF/Basics.lean
```diff
@@ -2676,358 +2686,330 @@ theorem and_eq_or (b c : Bool) : (b && c) = (b || c) → b = c := by
...
-def modifierComparison (m1 m2 : Modifier) : Comparison :=
-  match m1, m2 with
-  | plus, plus => eq
-  | plus, _ => gt
-  | natural, plus => lt
-  | natural, natural => eq
-  | natural, _ => gt
-  | minus, minus => eq
-  | minus, _ => lt
+theorem buyTicket_idempotent (t : Traveler) :
```
xhalo32, line 2809: have we explained what idempotent means?

LF/Basics.lean
mwhicks1: Like 1531 (on the `main` version) introduces simplification rules (aka characterizing lemmas -- we give both terms there), so we are good.

LF/Basics.lean
```diff
@@ -1372,6 +1372,16 @@ inductive Nat : Type where
   | succ (n : Nat)
 ```
 
+The following lines make ordinary numerals such as 0, 1, and 2 work with our {name}`Nat` type.You can ignore the details for now.
```
Robertboy18: Now that the late-days exercise is gone, do we still need this `OfNat` setup? I couldn’t find another Lean example that uses numeric literals with this `Nat`. The later `factorial_bad` example still fails because `HSub` is missing, before Lean reaches the termination check.

LF/Basics.lean
```diff
@@ -2676,358 +2686,330 @@ theorem and_eq_or (b c : Bool) : (b && c) = (b || c) → b = c := by
...
+and their bag is marked as needing inspection.
+Calling checkIn in any other state does nothing.
+
+:::exercise (rating := 1) (name := "check_in")
 
-::::exercise (rating := 2) (name := "grade_comparison")
-Here, we will need to access the fields of the `Grade` structure.
-The field names are `letter` and `modifier`, so for a grade `g`,
-we can write `g.letter` and `g.modifier` to access these fields.
+Define `check_in`.
```
Robertboy18: Should this be `checkIn` to match the function name below?

LF/Basics.lean
```diff
@@ -2676,358 +2686,330 @@ theorem and_eq_or (b c : Bool) : (b && c) = (b || c) → b = c := by
...
+    (screeningStatus : ScreeningStatus) :
+    checkIn (.checkedIn bagContent screeningStatus) = .checkedIn bagContent screeningStatus := solution!(by rfl)
 
-```lean
-theorem lowerLetter_F_is_F : lowerLetter F = F := by rfl
+attribute [irreducible] checkIn
 ```
 
-::::exercise (rating := 2) (name := "lower_letter_lowers")
+Buying a ticket and then checking in always leaves the traveler checked in,
```
Robertboy18: This sounds like a claim about any traveler, but the theorem only starts from `.noTicket`. Could we narrow the prose to say that a traveler with no ticket becomes checked in after these two operations?

LF/Basics.lean
```diff
@@ -2676,358 +2686,330 @@ theorem and_eq_or (b c : Bool) : (b && c) = (b || c) → b = c := by
...
+{name}`inspectBag` and {name}`replaceBag` commute when the traveler has not checked in.
+:::
+
+:::exercise (rating := 2) (name := "inspect_replace_commute")
 ```lean
-theorem grade_lowered_once : ∀ (lateDays : NatPlayground.Nat) (g : Grade),
-    (Nat.ble lateDays 9 = false) →
-    (Nat.ble lateDays 17 = true) →
-    applyLatePolicy lateDays g = lowerGrade g := by
+theorem inspectBag_replaceBag_comm_ticketed
```
Robertboy18: The prose covers any traveler who has not checked in, which includes both `.noTicket` and `.ticketed`, but this theorem only covers `.ticketed`. Also, `screeningStatus` is unused. Should we add the `.noTicket` case, or narrow the prose and remove that parameter?

LF/Basics.lean
berberman: This `Nat` is used in the next chapter (Induction) , and I don't think this looks better:
```lean
end NatPlayground

def ofNat : Nat → NatPlayground.Nat
  | .zero => .zero
  | .succ n => .succ (ofNat n)

instance (n : Nat) : OfNat NatPlayground.Nat n := ⟨ofNat n⟩

namespace NatPlayground
```

FYI TPiL also has similar declarations, though they are hidden:
https://github.com/leanprover/theorem_proving_in_lean4/blob/63fad08fcd5f4f3b8b0464561e2fa252671296b9/book/TPiL/InductiveTypes.lean#L1577-L1600

LF/Basics.lean
berberman: Good catch. I believe `factorial_bad` was already incorrect before this PR. Given that we are using this `Nat` until UsingLean, I think we need to define the subtraction.

berberman: Actually let's just use `pred`.

berberman:
@Robertboy18 Thanks for the careful review!

Robertboy18 (approved):
Thanks, these changes address my earlier comments and the updated build is passing. I noticed two tiny prose typos: `checked in,in which case` needs a space after the comma, and `A traveler can check in only after buying a ticket. and their bag...` should be one sentence. Otherwise this looks good to me!

LF/Basics.lean
```diff
@@ -2685,358 +2695,347 @@ theorem and_eq_or (b c : Bool) : (b && c) = (b || c) → b = c := by
...
-Suppose that a course has a grading policy based on late days,
-where a student's final letter grade is lowered if they submit too
-many homework assignments late.
-::::
 
-```lean
-namespace LateDays
-open scoped NatPlayground.Nat
+:::full
+Now that we have learned the basic features of Lean 4, let's close the chapter
```
dsainati1: Is this the first time we mention the Version number? I think this is confusing if so, we should explain earlier in the chapter that Lean 4 is the current (non-backwards compatible) version of Lean and we will just say Lean to refer to Lean 4 throughout this book.

LF/Basics.lean
```diff
@@ -2685,358 +2695,353 @@ theorem and_eq_or (b c : Bool) : (b && c) = (b || c) → b = c := by
...
 
-Niklas Halonen (xhalo32): We need to teach how to prove a goal that looks like `natural ≠ minus` for example.
-One could write `injection x` for example:
+:::exercise (rating := 1) (name := "replace_bag")
+Define `replaceBag`.
 ```lean
-example : natural ≠ minus := by
-  intro x
-  injection x
+def replaceBag (newContent : BagContent) (t : Traveler) : Traveler := solution!(
```
dsainati1: I generally like this exercise except for this part. The `replaceBag` operation feels a bit contrived to me because there isn't really a way to replace your bag after it's been inspected. If we're talking about checked bags (as I assume we are because of the battery check), the screening happens once the bag has left your possession, so you can't replace it. 

Maybe the way to clear this up is to tweak the exercise's framing slightly to apply to carry-on luggage (and change the battery restriction to e.g. liquids). In this case it is actually possible to leave the security-screened section of an airport, requiring another inspection in order to re-enter.

bcpierce00: I like this idea!

bcpierce00:
#Note: Looks great, modulo Daniel's comment about the very last part.

dsainati1 (approved):
I like this exercise. I think it will probably need a couple tweaks but it's in good enough condition IMO that we should get it out to the alpha testers and see what they have to say.

PR #214, opened 2026-08-08
Cleaning up CONTRIBUTING.md and STYLE.md

ionathanch:
This PR cleans up the contributors' guide and revamps the style guide, removing extraneous information and reorganizing as needed.


Robertboy18 (commented):
I read through the reorganized Verso section. I found three things that seem worth fixing before this is merged:

- The manual-grading path has disappeared from the guide. `:::grade` is still supported and is still used for `GRADE_MANUAL` exercises (for example in `TS/Types.lean`), while `:::gradeTheorem` is the structured form for theorem grading. Could we keep a short description and example of `:::grade`, and mention the optional `(manual := true)` / `(level := Advanced)` exercise fields?
- The solution section says the mechanisms produce two source variants, but the saver produces three: solutions, student, and terse. `solution!` is replaced with `sorry` in both student and terse output. It would help to include the terse path and behavior here.
- The `-- SOLUTION` paragraph lists “a multi-line proof” as a use case, but the guidance immediately below says not to use it for proof bodies because that leaves an empty `by` block. I think the first paragraph should instead say that it can wrap an entire top-level theorem declaration.

A few small wording fixes: “Inline code fences” would be clearer as “Inline roles” or “Inline markup”; “The below list” should be “The sections below list”; and “as describe above” should be “as described above.”

Robertboy18 (commented):
Small correction to my second bullet: there is also a separate `grading` output in addition to `student`, `solutions`, and `terse`. The underlying issue is unchanged: this section currently describes only two outputs and omits `terse`, where `solution!` is also replaced with `sorry`. The special grading output retains the solution form.

ionathanch:
I've added the `grade` directive back in. Why do we have that though? Shouldn't manually graded exercises be marked using `:::exercise (manual := true)` rather than with a separate directive?

Also to clarify so far I've only edited the text from "Directive fence depth" through to the end of "Quiz and solution directives", with the others rearranged but not yet fixed, so they're likely outdated or wrong. As TODOs to myself, these subsections remain:

* `lean` block flags
* Solution mechanisms inside `lean` blocks
* BNF grammars
* Displays: `` ```display `` and `` ```displaymath ``
* Inline roles
* Author-only annotations
* Structural and presentation blocks
* Diagrams with ASCII fallback

ionathanch:
Wait what is the `grading` variant for? It looks the same as the `solutions` variant to me

xhalo32:
I have yet to finish the documentation for the grading variant. Sorry for the confusion!

ionathanch:
I've just added that it's the same as the solutions variant except the extracted Lean has `autograding` attributes, I hope that's correct
Feel free to commit any elaborations if needed

ionathanch:
(Converted back to draft PR because I skimmed through the Lean Style section and that needs desloppifying)

ionathanch:
Ok I'm done for real now. It may be easier to look at the rendered [STYLE.md](https://github.com/plclub/sf-in-lean/blob/ionathanch/md-files/STYLE.md) and [CONTRIBUTING.md](https://github.com/plclub/sf-in-lean/blob/ionathanch/md-files/CONTRIBUTING.md) than at the diffs.

STYLE.md
```diff
@@ -308,80 +306,47 @@ def x : Nat := "str"
...
+```
 
-Guard `#check` only when its printed form is what the example is testing. For
-example, `Imp` checks its custom delaborator this way:
-
-````lean
-```lean
-/-- info: aexp {3 + X * 2} : Aexp -/
-#guard_msgs in
-#check aexp {3 + (X * 2)}
```
berberman, line 352: I think this usage of `#guard_msgs` is still worth documenting: they are not public facing (either outside of verso files or in ignore blocks) and are only for testing purpose.

bcpierce00 (approved):

PR #215, opened 2026-08-08
Slidebreaks weren't being emitted in `terse`

ionathanch:


ionathanch:
If you compare e.g. https://sf-in-lean.torus.icu/main/lf/terse/preview/ vs. https://sf-in-lean.torus.icu/pr/215/preview/lf/terse/ with Inspect Element in a browser, right after "A function on days" will now be a `<div class="slide-break"></div>` element.

mwhicks1:
How is this slide break used in HTML? Do we need more to generate slide-like views of the material?

mwhicks1 (approved):

ionathanch:
Looking in the original SF it seems like there's a JS script that turns a terse document into slides according to the slide boundaries: https://github.com/DeepSpec/sfdev/blob/master/common/slides.js
Although I've never seen this be used and I haven't tried it myself

PR #216, opened 2026-08-09
Refactor Tactics

berberman:
Summary of changes:
- Explain implicit arguments to clarify `apply trans_eq` example
- Also introduce named arguments
- Rework "Using Tactics on Hypotheses" section to "Using `apply` on Hypotheses" (which only covers `apply thm at h`), as `rw/dsimp [h] at h'` has alreay been introduced in earlier chapters
- Introduce `specialize` tactic after `replace` (`STYLE.md` now says it's part of Automation chapter, but it appears nowhere!)
- Rework "Generalizing the Induction Hypothesis" section to drop "delayed intro" material and directly introduce `induction ... generalizing` technique
- Remove the `≤?` notation and whole bunch of `beq`/`ble` theorems
- Replace `ble`/`beq` exercises
- Improve tactic list in "Review" section
- Bring back and redo some failing exercises
- Clean up the old comments
- Many style/wording polishing!


LF/Tactics.lean
```diff
@@ -1387,677 +1256,295 @@ theorem double_injective : ∀ (n m : Nat),
...
-    dsimp; cases (n == 5)
-    case false => rfl
-    case true => rfl
-  case true => rfl
+def chooseIf {α : Type} (test : α → Bool) (x y : α) : α :=
+  if test x then x else y
+
+theorem chooseIf_self {α : Type} (test : α → Bool) (x : α) :
+    chooseIf test x x = x := by
+  unfold chooseIf
```
dsainati1: We have not actually taught the `unfold` tactic yet! A very brief explanation in a new section prior to this one makes sense to me here.

xhalo32: I believe Chris and I came to the conclusion on Zulip that we should not teach `unfold`. One option is to use `rw [chooseIf]`. Should we open another discussion about this on Zulip?

dsainati1: `unfold` gets used extensively throughout later volumes (e.g. Hoare), so if we are planning on this a lot of stuff will need to change. Definitely worth making a zulip thread imo.

xhalo32: I'm working through Hoare and removing all `unfold`s at the moment.

dsainati1: I still think it's worth making explicit in a Zulip discussion; i was not aware we were removing the tactic entirely. This also changes some of the discussion we have in informal proofs around "unfolding definitions" or reasoning "by the definition of X, then Y". So it's a pretty involved change to presentation, imo.

LF/Logic.lean
```diff
@@ -2105,7 +2108,9 @@ reflexive.)
 theorem beq_eq_true (n m : Nat) :
     (n == m) = true ↔ n = m := by
   constructor
-  · apply beq_eq
+  · rw [Nat.beq_eq_true_eq]
```
xhalo32: This just feels like going in a circle. `Nat.beq_eq_true_eq` is just a version of this `beq_eq_true`. We should at least simplify the proof to just
```
theorem beq_eq_true (n m : Nat) :
    (n == m) = true ↔ n = m := by
  rw [Nat.beq_eq_true_eq]
```
But to be honest I think we either should remove this as `Nat.beq_eq_true_eq` which is `beq_iff_eq` which comes from `LawfulBEq` which we don't want to explain here.

LF/Tactics.lean
```diff
@@ -89,150 +61,141 @@ previously proved lemma.
...
 ```lean
-theorem silly1 (n m : Nat) : n = m → n = m := by
-  intro eq
-  /- Here, we could finish with `rw [eq]` as we
+example (n m : Nat) (h : n = m) : n = m := by
+  /- Here, we could finish with `rw [h]` as we
     have done several times before.  Or we can finish
     by using `apply`: -/
-  apply eq
+  apply h
```
xhalo32, line 74: Should we use `exact` here as there is nothing for lean to infer which would make `apply` more convenient?

LF/Tactics.lean
```diff
@@ -89,150 +61,141 @@ previously proved lemma.
...
 :::::
 
 ::::::
 
 ::::full
-To use the `apply` tactic, the (conclusion of the) fact
+To use the {tactic}`apply` tactic, the (conclusion of the) fact
 being applied must match the goal exactly (perhaps after
-simplification) -- for example, `apply` will not work if the left
+simplification) —bodies for example, {tactic}`apply` will not work if the left
```
xhalo32: Which bodies?

LF/Tactics.lean
```diff
@@ -89,150 +61,141 @@ previously proved lemma.
...
 and right sides of the equality are swapped.
 ::::
 
 :::slidebreak
 :::
 
 ::::terse
-The goal must match the hypothesis _exactly_ for `apply` to
+The goal must match the hypothesis _exactly_ for {tactic}`apply` to
```
xhalo32: Technically it doesn't need to match "exactly", but up to definitional equality. Not sure where this clarification belongs.

LF/Tactics.lean
```diff
@@ -89,150 +61,141 @@ previously proved lemma.
...
 
 :::solution
-The `rw` tactic is used to apply a known equality (a
+The {tactic}`rw` tactic is used to apply a known *equality* (a
 hypothesis from the context or a previously proved lemma) to
 modify the goal, replacing all occurrences of one side by the
 other.
 
-The `apply` tactic uses a known implication (a hypothesis from the
+The {tactic}`apply` tactic uses a known *implication* (a hypothesis from the
```
xhalo32: But the first example of `apply` was not an implication... That's why `exact` would be better. I think that `apply` should usually be used with implications (or foralls).

LF/Tactics.lean
```diff
@@ -89,150 +61,141 @@ previously proved lemma.
...
 context, a previously proved lemma, or a constructor) to replace a
 goal that matches the conclusion of the implication with subgoals,
 one for each premise of the implication.
```
xhalo32: This is technically incorrect as it can often infer some of the premises (that appear as variables in the goal for example). So it doesn't create a subgoal for each premise, but for a subset of premises which it couldn't infer.

LF/Tactics.lean
```diff
@@ -245,225 +208,201 @@ can also be used to modify a hypothesis rather than the goal.)
...
-```
-HIDE: Robert Rand: I found using m, n and o throughout this discussion
-super confusing -- m doesn't come between n and o! Rocq's eq_trans uses
-x, y and z, which is what I wanted to change this too anyhow.
-```
-:::
+_transitive_.
 
 ```lean
 theorem trans_eq {α : Type} (x y z : α) :
```
xhalo32, line 226: Should we make x, y, z implicit?

LF/Tactics.lean
```diff
@@ -245,225 +208,201 @@ can also be used to modify a hypothesis rather than the goal.)
...
-theorem trans_eq_example'' (a b c d e f : Nat) :
-    [a, b] = [c, d] →
-    [c, d] = [e, f] →
+example (a b c d e f : Nat)
+    (h₁ : [a, b] = [c, d])
+    (h₂ : [c, d] = [e, f]) :
     [a, b] = [e, f] := by
-  intro eq1 eq2
-  apply trans_eq [a, b] [c, d] [e, f] eq1 eq2
+  apply trans_eq [a, b] [c, d] [e, f] h₁ h₂
```
xhalo32, line 303: If `trans_eq` uses implicit variables this could be just
```
exact trans_eq h₁ h₂
```

LF/Tactics.lean
```diff
@@ -245,225 +208,201 @@ can also be used to modify a hypothesis rather than the goal.)
...
 In the previous example, we had to specify the `x` and `z` arguments
-to `trans_eq` before we could supply `[c, d]` for `y` or `eq1` and `eq2` for
+to {name}`trans_eq` before we could supply `[c, d]` for `y` or `eq1` and `eq2` for
 the premises. However, we just said that Lean was able to infer these arguments, so it's
-a bit redundant (and wordy) for us to do it. Thankfully,
-Lean allows us to use `_`s for positional arguments that it can infer.
+a bit redundant (and wordy) for us to do it.
+:::
+
+Thankfully, Lean allows us to use `_`s for positional arguments that it can infer.
```
xhalo32, line 313: Would it make sense for underscores to be explained in Poly.lean

xhalo32 (commented):
Reviewed until line 400

LF/Tactics.lean
berberman: Skimmed through Poly, I don't think there's a good spot as we already use implicit aguments there. Let's just introduce it here.

LF/Tactics.lean
berberman: I think the idea is let `trans_eq` use visible arguments and then introduce `_` and mention implicit arguments. Also it's about practicing passing named arguments.

LF/Tactics.lean
berberman: https://github.com/plclub/sf-in-lean/pull/216#discussion_r3780581682

LF/Tactics.lean
```diff
@@ -472,74 +405,65 @@ applying the constructor `.succ` to another number.  But there is more
...
+    (h : n + 1 = m + 1) :
     n = m := by
-  intros h1
-  have h2 : n = Nat.pred (n + 1) := by rfl
-  rewrite [h2, h1]
+  have : n = Nat.pred (n + 1) := by rfl
+  /- The hypothesis name defaults to `this` when unspecified. -/
+  rewrite [this, h]
   rfl
 ```
```
xhalo32, line 447:
````suggestion
We can _prove_ the injectivity of {name}`Nat.succ` by using the {name}`Nat.pred` function and its characterizing lemma {name}`Nat.pred_succ`:

```lean
example (n m : Nat)
    (h : n + 1 = m + 1) :
    n = m := by
  have : n = Nat.pred (n + 1) := Nat.pred_succ n
  /- The hypothesis name defaults to `this` when unspecified. -/
  rewrite [this, h]
  rfl
```
````

I think using `rfl` is overusing defeq since the equation contains a variable `n`, so I would prefer we use the characterizing lemma instead.

File: LF/Tactics.lean [hunk 12]
```diff
@@ -876,401 +738,381 @@ and we apply the tactic `injection h with hxy`.  What will happen?
...
 that get us closer to the main goal we're trying to prove. Often,
 though, these facts are just special cases of more general hypotheses
 we already have.
 
-If `h` is a quantified hypothesis in the current context -- i.e.,
-`h : forall (x : α), P` -- then `have h := h (x := e)` will
-change `h` so that it looks like `P` with `x` replaced by `e`.
+If `h` is a quantified hypothesis in the current context — i.e.,
+`h : ∀ (x : α), P x` — then we can use {tactic}`have` to obtain a special
+case of `h` by supplying a value for `x`. For example, `have h := h (x := e)`
```
xhalo32: Why do we use `(x := e)` when the binder is explicit?

LF/Tactics.lean
```diff
@@ -876,401 +738,381 @@ and we apply the tactic `injection h with hxy`.  What will happen?
...
+Often we don't care to keep this old hypothesis around, and so we can use the {tactic}`replace`
+tactic instead. It behaves like {tactic}`have`, except that
+it gets rid of the old hypothesis afterwards when possible:
 
 ```lean
-theorem replace_example m :
-    (∀ n, m * n = 0) → m = 0 := by
-  intro h
+example (m : Nat) (h : ∀ n, m * n = 0) : m = 0 := by
   replace h := h (n := 1)
```
xhalo32: I have never needed to use `replace`, and I think `clear` is more useful. Why do we teach `replace` but not `clear`?

LF/Tactics.lean
```diff
@@ -876,401 +738,381 @@ and we apply the tactic `injection h with hxy`.  What will happen?
...
-intro n; induction n
+::::
+
+:::dev "Yipeng Liu (berberman)"
+A single {tactic}`contradiction` can close `zero.succ` case without any rewrite as shown below —
+`h : double 0 = (m' + 1).double` gets unfolded to `0 = ((m'.double).add 1).succ`,
+and then `noConfusion` kicks in. The unfold can happen because {name}`Nat.double` is not marked as
+`irreducible`. Similarly it can close `succ.zero` case. I think this is fine, and I removed the
+{name}`Nat.double_zero`/{name}`Nat.double_succ` rewrites.
+:::
```
xhalo32: I disagree with removing the explicit rewrites. `contradiction` is doing too much automation here which will confuse the reader. Even after `rw [Nat.double_zero, Nat.double_succ] at h` the hypothesis looks like `0 = Nat.double m' + 2` where it hasn't been clearly explained that `+ 2` reduces to `.succ.succ` so the contradiction derives from no confusion of constructors.

~~I feel like this needs its own tactic (like `nomatch`) just like we have for constructor injectivity. `contradiction` can then be explained to automate both of these.~~

So this is what I would write instead
```
theorem double_injective (n m : Nat) (h : n.double = m.double) : n = m := by
  induction n with
  | zero =>
    cases m with
    | zero => rfl
    | succ m' =>
      rw [Nat.double_zero, Nat.double_succ] at h
      contradiction
  | succ n' ih =>
    cases m with
    | zero =>
      rw [Nat.double_zero, Nat.double_succ] at h
      contradiction
    | succ m' =>
      congr
```
(or `nomatch h` if we decide to use that instead)

LF/Tactics.lean
```diff
@@ -1279,785 +1121,375 @@ example (n m : Nat) :
...
+    | zero => rfl
+    | succ m' => contradiction
+  | succ n' ih =>
+    cases m with
+    | zero => contradiction
+    | succ m' =>
+      congr
+      apply ih -- now works
+      rw [Nat.double_succ, Nat.double_succ] at h
+      injections
```
xhalo32, line 1252:
```suggestion
      injections h
```
is more clear in my opinion

LF/Tactics.lean
```diff
@@ -1279,785 +1121,375 @@ example (n m : Nat) :
...
-    cases n
-    case zero => exact HS0 _ (ih _)
-    case succ => exact HSS _ _ (ih _ )
-
--- /ADMITTED
--- []
+      /- A trick here: by using `rfl` to close `(ys ++ x :: l₂).length = n`
+         we effectively choose `n` to be `(ys ++ x :: l₂).length`
+      -/
+      rw [ih rfl]
```
xhalo32, line 1446: Have we explained `rfl` term yet? It doesn't seem like Poly.lean explains it.

LF/Tactics.lean
```diff
@@ -2080,157 +1512,150 @@ get the first and second projections of `v` using this tactic:
...
   solution!
-    intro h
-    induction l generalizing l1 l2
-    case nil =>
-      injections h1 h2
-      rw [← h1, ← h2]
+    induction l generalizing l₁ l₂ with
+    | nil =>
+      rw [unzip_nil] at h
+      injections h₁ h₂
```
xhalo32, line 1592: It's not clear that this destructures `h`.

LF/Tactics.lean
```diff
@@ -2080,157 +1512,150 @@ get the first and second projections of `v` using this tactic:
...
-      injections h1 h2
-      rw [← h1, ← h2]
+    | cons x xs ih =>
+      let ⟨a, b⟩ := x
+      dsimp [unzip] at h
+      injections h₁ h₂
+      rw [← h₁, ← h₂]
       dsimp [zip]
       rw [ih]
       rfl
```
xhalo32, line 1602: Have we taught that `rfl` can do eta expansion?

xhalo32 (commented):
I've reviewed the whole file and made a couple of changes.

~~Mostly looks good, but I think the way we present `contradiction` and `injection` needs some work. As `contradiction` can prove goals using constructor injectivity and no confusion, we should introduce the `nomatch` tactic to explain the no confusion counterpart to `injection`. This can be a subject of another PR too.~~

After further testing I noticed that `contradiction` can't actually do injection, so we are using `contradiction` pretty much only as a replacement for `nomatch` rather than automation around both `injection` and `nomatch`. This seems mostly fine to me although `nomatch` lets you specify the hypothesis name so I still prefer that.

LF/Tactics.lean
berberman: For performance, I don't think there will be a noticable difference between `exact` and `apply` in this case; but for style, yes, we should prefer `exact` over `apply` as it's more stable. Since we're talking about `apply` in this section, let's add some quantified variables to the hypothesis.

LF/Tactics.lean
berberman: Me neither. I also never use `clear`. I think shadowing is pretty common in proofs, and it's not harmful. Let's just teach `replace` here.

LF/Tactics.lean
berberman: Good catch. I was proposing this in Lists:
https://github.com/plclub/sf-in-lean/blob/ba52c1c54c11a9cd2c86605f0648e41bc8474591/LF/Lists.lean#L205-L219

We'll have to destruct the `Prod` without relying on eta. Let's keep this proof and later introduce the idea in Lists.

LF/Tactics.lean
berberman: After `rw` we have `h : (l₁, l₂) = ([], [])` so `injections` splits it.

LF/Tactics.lean
berberman: That's why I wrote `by rfl`. It seems that you changed it to `rfl`...

LF/Tactics.lean
berberman: Ok let's add back the explicit rewrites.

LF/Tactics.lean
xhalo32: I changed the `by rfl` to `rfl` at https://github.com/plclub/sf-in-lean/blob/6be73180cebf6fdfc0c546b40d7d994b6b72b220/LF/Tactics.lean#L1441 to be consistent with this. I think we have converged to introducing `rfl` terms in Lists.lean.

LF/Tactics.lean
```diff
@@ -89,150 +61,138 @@ previously proved lemma.
...
-
-When Lean matches the current goal against the conclusion of `h`,
-it will try to find appropriate values for these variables.  For
-example, when we do `apply eq2` in the following proof, the
-universal variable `q` in `eq2` gets instantiated with `n`, and
-`r` gets instantiated with `m`.
+When we use `apply h`, Lean tries to match the conclusion of the type
+of `h` with the current goal. Here `h : n = m → [n, o] = [m, p]` has conclusion
+`[n, o] = [m, p]`, which matches the current goal. Lean then replaces the goal
+with the premise that is still need, `n = m`. Then we close the goal with `apply hnm`.
```
dsainati1:
```suggestion
with the premise that is still needed, `n = m`. Then we close the goal with `apply hnm`.
```

LF/Tactics.lean
```diff
@@ -89,150 +61,138 @@ previously proved lemma.
...
+
+More generally, the type of a theorem or hypothesis used with {tactic}`apply` may have
+universally quantified variables and premises. Lean tries to unify its conclusion with
```
dsainati1: I don't like the use of the word "unify" here; it feels very technical. Can we say something a bit more vague like "match"?

LF/Tactics.lean
```diff
@@ -549,118 +474,113 @@ to generate all equations that it can infer from `h` using the
...
 
+Notice that due to the way addition on naturals is defined, deriving a contradiction from `1 + n = 0` is not as trivial as it seems.
+
+```lean +error -keep
+example (n : Nat)
+    (h : 1 + n = 0) :
+    2 + 2 = 5 := by
+  contradiction -- doesn't work because `1 + n` doesn't reduce to `n.succ`.
+```
+
```
dsainati1: This is good, but let's also show another example illustrating what we'd need to do to make `contradiction` work in this case

LF/Tactics.lean
```diff
@@ -876,401 +749,390 @@ and we apply the tactic `injection h with hxy`.  What will happen?
...
+the _goal_ and works backward through implications that would prove
+it, until remaining goals are facts that are already known.
 
-The informal proofs seen in math or computer science classes tend
-to use forward reasoning.  By contrast, idiomatic use of Lean
-generally favors backward reasoning, though in some situations the
-forward style can be easier to think about.
+The informal proofs in mathematics and computer science often
+use forward reasoning.  In Lean, however, backward reasoning is often more
+idiomatic, though forward reas can sometimes be easier to follow or more natural for
```
dsainati1:
```suggestion
idiomatic, though forward reasoning can sometimes be easier to follow or more natural for
```

LF/Tactics.lean
```diff
@@ -876,401 +749,390 @@ and we apply the tactic `injection h with hxy`.  What will happen?
...
-```display
-theorem double_injective: forall n m,
-  double n = double m →
-  n = m
+```lean -keep
+theorem double_injective (n m : Nat) (h : n.double = m.double) : n = m := sorry
 ```
 
-The way we start this proof is a bit delicate: if we begin it with
+If we begin it with
```
dsainati1: "It" no longer refers to anything without the stuff before the colon. I liked the old text better.

LF/Tactics.lean
berberman: We have implication examples above.

dsainati1 (approved):

xhalo32 (approved):

PR #218, opened 2026-08-09
Advice on writing new SFL material

mwhicks1:
After going over the text and doing some polishing passes, I stepped back and ask myself: What changes am I making and why am I making them? I figured that working that out somewhat carefully and turning it into advice/guidance might be useful to to SFL writers. 

After I wrote the first draft of this file, I asked Claude to find before/after examples of the advice, which were added at the end of the note. They are not perfect or 100% on target, but there are still useful to read so I left them.

If you have other ideas you think are worth including, or example illustrations, let me know!

### DIFF
--- a/STYLE-WRITING.md
+++ b/STYLE-WRITING.md
@@ -39,6 +39,11 @@ This paragraph might go stale quickly.
 ## Examples 
 
 The pairs below are real polishing diffs pulled from the repo's git history, each illustrating one of the principles above. *(This section was compiled by Claude, from `git log`/`git show` on `LF/Basics.lean`, `LF/Induction.lean`, and `LF/Typeclasses.lean`. It has been human-reviewed; it does not perfectly illustrate the above ideas, but the examples are hopefully useful nevertheless.)*
+:::dev "Benjamin Pierce (bcpierce00)"
+Ugh. 
+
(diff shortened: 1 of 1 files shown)


bcpierce00:
THis is good and useful.  I left a few comments for discussion (either now or later).  

My one big comment is that I'm not very happy leaving material _in a chapter on writing style_ with a note that says it's not really what we ought to have written because we did it with AI...  :-|

bcpierce00:
#Note: Would like to see the AI-boosted examples polished better.

mwhicks1:
Sigh. You are right. I have failed the "tasteful and good AI-usage" test (once again)! 

How about I just delete the examples for now, and get them added later? I'm not sure how much time I'll have to work harder on this in the near term. Hopefully the rest of the note is still useful.

Robertboy18:
I went through all six examples against the cited commits and pushed a cleanup pass. I also corrected two descriptions that didn’t quite match the original diffs and removed the stale developer notes!

bcpierce00:
This looks great, @Robertboy18 !  I think it's ready to merge and will do so shortly unless you say there's anything you want to add.

bcpierce00 (approved):

PR #224, opened 2026-08-10
IndProp: fix reader-facing formatting artifacts

Robertboy18:
## Summary

I fixed several small formatting issues I noticed while reading through IndProp:

- format the chapter imports using the standard import block
- restore the missing Permutations heading
- fix malformed list notation and stray empty-list artifacts
- correct several Lean names and type examples
- clean up the palindrome exercise wording

AI use: I used GPT to help check these small fixes:)!

### DIFF
--- a/LF/IndProp.lean
+++ b/LF/IndProp.lean
@@ -311,6 +311,10 @@ this definition in a standard programming language, but it is
 rejected by Lean's termination checker, since the argument to
 the recursive call, `csf n`, is not "obviously smaller" than `n`.
 
+:::dev "Kihong Heo (KihongHeo)" NOW
+Probably `reaches1In` according to STYLE.md?
+:::
+
 ```lean
 /--
 error: fail to show termination for
@@ -357,6 +361,10 @@ while we could in principle convince Lean that `div2 n` is
 smaller than `n`, we certainly can't convince it that
 `(3 * n) + 1` is smaller than `n`!
 
+:::dev "Kihong Heo (KihongHeo)" NOW
+Probably `CollatzHoldsFor` according to STYLE.md?
+:::
+
 ```lean
 /--
 error: fail to show termination for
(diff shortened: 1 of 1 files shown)


LF/IndProp.lean
```diff
@@ -3248,7 +3248,7 @@ and
 [4, 3].
 ```
 
-Now, suppose we have a set `α`, a function `test: α→bool`, and a
+Now, suppose we have a set `α`, a function `test : α → Bool`, and a
```
chenson2018:
```suggestion
Now, suppose we have a type `α`, a function `test : α → Bool`, and a
```

The usage of the word set is different in Lean. Rocq's standard library `Ensemble := α → Prop` is called `Set` in Mathlib.

It would be worth doing a careful pass to eliminate incorrect usage of "set". I'd open an issue if this will happen in a follow up PR.

LF/IndProp.lean
KihongHeo, line 338: Probably `reaches1In` according to STYLE.md?

KihongHeo, line 371: CollatzHoldsFor

KihongHeo, line 403: I feel that this paragraph does not match the error message below. The Lean error msg refers to `div 2`.

KihongHeo (commented):
Looks good. I have added several comments.

LF/IndProp.lean
bcpierce00: Yes, let's do this in a follow-up PR.

bcpierce00: I'll create an issue.

bcpierce00 (approved):

bcpierce00:
#Note: Ready to merge once a few comments are squashed

berberman:
I think this chapter needs a careful pass to polish it, for example, applying the proper style and improving the wording, similar to what I’m doing in #216. It's on my plan to work on after #216, but anyone is welcome to help out as a volunteer. I think we can merge this PR first and continue the polishing work in subsequent PRs.

KihongHeo:
Yes. Let me make a pass over this file in another PR.

PR #229, opened 2026-08-11
Bump to 4.33.0 + structure Totalmap

xhalo32:
Bumps our toolchain plus the autograder to 4.33.0.

Here's the [changelog](https://lean-lang.org/doc/reference/latest/releases/v4.33.0/#release-v4___33___0)

This PR also depends on [this upstream fix](https://github.com/robertylewis/lean4-autograder-main/pull/22).

Another minor change is that the `def TotalMap` in Typeclasses is now `structure TotalMap` as per zulip discussion.

## Unresolved questions

- ~~How to fix `set_option backward.isDefEq.respectTransparency.types false` in Typeclasses.lean?~~


LF/Typeclasses.lean
```diff
@@ -800,34 +817,62 @@ On the other hand, if we update a map `m` at a key `a₁` and then look up a _di
...
+  ext a
+  by_cases h : "bar" = a
+  · subst h
+    rw [update_eq, update_neq (show "foo" ≠ "bar" by simp), update_eq]
+  · simp only [update_apply]
+    rw [beq_false_of_ne h]
+    simp
+```
+
+Given keys `a₁` and `a₂`, the tactic {tactic}`by_cases` `h : a₁ = a₂` splits the proof into the case where they are equal — where `subst h` then replaces one by the other — and the case where they are not, which is what {name}`update_neq` wants. Use it to prove the following theorem, which states that if we update a map to assign key `a` the same value as it already has in `m`, then the result is equal to `m`:
```
Robertboy18, line 860: Is this paragraph meant to sit just before `update_same`? It describes assigning `m[a]` back to `a`, but the exercise immediately below is `update_shadow`.

LF/Typeclasses.lean
```diff
@@ -511,6 +511,7 @@ We place no constraints on the value type `β`.
 The {ref "Lists"}[Lists] chapter introduced a partial map abstraction, `PartialMap`, with a
 `find` function for lookup, based on lists of key-value pairs.
 Here, we are going to build a map abstraction using functions instead. The advantage of this representation is that it offers a more _extensional_ view of maps, as we saw with functions in the {ref "Logic"}[Logic] chapter: two maps that respond to every query in the same way will be represented as exactly the same function, rather than just as "equivalent" list structures. This simplifies proofs that use maps.
```
Robertboy18, line 513: Now that `TotalMap` is a structure, “represented as exactly the same function” no longer seems quite right. Perhaps say that maps answering every query identically can be proved equal?

LF/Typeclasses.lean
Robertboy18, line 1: At `LF/Typeclasses.lean:1051`: This explanation seems left over from when `TotalMap` was a bare function type. Now both maps are structures. Could we explain the actual distinction: total maps always return a value, while partial maps return an `Option`?

LF/Logic.lean
```diff
@@ -676,7 +676,7 @@ Not (a : Prop) : Prop
 ```
 
 ```leanOutput not
-def Not : Prop → Prop :=
+@[implicit_reducible] def Not : Prop → Prop :=
```
dsainati1, line 679: We can't add this annotation without explaining it. Why is it necessary?

LF/Typeclasses.lean
dsainati1, line 513:
```suggestion
Here, we are going to build a map abstraction using functions instead. The advantage of this representation is that it offers a more _extensional_ view of maps, as we saw with functions in the {ref "Logic"}[Logic] chapter: two maps that respond to every query in the same way will be represented as structures containing exactly the same function, rather than just as "equivalent" list structures. This simplifies proofs that use maps.
```

LF/Typeclasses.lean
```diff
@@ -597,6 +612,11 @@ Our typeclass `MyGetElem` is a simpler version of the standard library's {name}`
 which has many instances such as {name}`Array`, {name}`List`, and {name}`Vector`.
 We develop it to illustrate the notation-as-typeclass approach.
 
+:::dev "Niklas Halonen (xhalo32)"
+A reason I can come up with why we use a notation typeclass (in library code) over a plain notation is that it makes it possible (for a downstream consumer) to write `open scoped MyGetElem` instead of writing `open scoped TotalMap` and `open scoped TotalMap` individually.
```
dsainati1: Did you mean `open scoped TotalMap` and `open scoped PartialMap`?

xhalo32: Yes!

LF/Logic.lean
xhalo32: Notice that it's `leanOutput`. It's necessary as of 4.33.0 where `@[implicit_reducible]` was added to `Not` and a lot of other definitions. [Here's the relevant part of the changelog](https://lean-lang.org/doc/reference/latest/releases/v4.33.0/#The-Lean-Language-Reference--Release-Notes--Lean-4___33___0-_LPAR_2026-08-10_RPAR_--Highlights--Breaking-Changes--Transparency)

LF/Typeclasses.lean
xhalo32: I think it's okay since it refers to the underlying functions that are mentioned earlier in the paragraph: "Here, we are going to build a map abstraction using functions instead"

I think what we should add is an explicit mention that we encapsulate these functions inside a structure.

LF/Typeclasses.lean
xhalo32: Good catch! It's now out of date indeed. Made a dev note about it

LF/Typeclasses.lean
xhalo32: I think this introductory paragraph is about the inner functions already. I've added
> Instead of using functions directly, we encapsulate them inside a `structure` which we call `TotalMap`.

after the paragraph.

xhalo32:
@Robertboy18 @dsainati1 Thank you for the reviews and improvements!

dsainati1 (approved):
I think this continues to move this file in the right direction

PR #231, opened 2026-08-11
Hoare: start going through proofs

xhalo32:
This PR reworks many of the proofs to be more idiomatic Lean (at least according to my intuition).

We also made the Imp command syntax in Imp.lean use semicolons in sequencing rather than after assignment.

## Unresolved questions

Right before we introduce the `apply_rules` automation, we show the reader that they don't need to use `rw [assertImplies_def] at himp` since `apply_rules` can work with `P ->> P'` by its definition. How do we want to explain this?

### DIFF
--- a/HL/Imp.lean
+++ b/HL/Imp.lean
@@ -1106,17 +1106,39 @@ inductive Com.EvalR : Com → State → State → Prop where
   | whileTrue (st st' st'' : State) (b : Bexp) (c : Com) (hb : b.eval st = true)
       (hc : EvalR c st st') (hloop : Com.EvalR (imp {while (~b) {~c}}) st' st'') :
       EvalR (imp {while (~b) {~c}}) st st''
+```
 
-notation:40 st0:41 " =[ " c " ]=> " st1:41 => Com.EvalR c st0 st1
+:::instructors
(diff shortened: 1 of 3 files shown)


xhalo32:
This needs to be rebased on main now that the underlying PRs have been merged. Done

HL/Imp.lean
```diff
@@ -736,39 +738,49 @@ unrecognized subcommand with the `~` escape.
 namespace Imp.Delab
 open Lean PrettyPrinter Delaborator SubExpr
 
-/-- Rebuild `imp_com` concrete syntax from a `Com` term. -/
-partial def delabComInner : DelabM (TSyntax `imp_com) := do
+/-- Rebuild `imp_com` concrete syntax from a command term whose constructors
+live in namespace `ns` — `Com` for Imp itself, and an extension chapter's own
+command type elsewhere, with its constructors beyond Imp's five handled by
+`extra`.  Constructors are matched by name (`ns ++ `seq`, …) because each
```
xhalo32:
```
`ns ++ `seq`
```
has unmatched backtics

dsainati1 (approved):

PR #232, opened 2026-08-11
Tackle dev comments in `Lists`

dsainati1:
There should be no more urgent dev comments in the file after this


LF/Lists.lean
```diff
@@ -1131,16 +1072,17 @@ which is immediate from the induction hypothesis.  _Qed_.
 In some situations, it is necessary to generalize a
 statement in order to prove it by induction.  Intuitively, the
 reason is that a more general statement also yields a more general
-(stronger) inductive hypothesis.
+(stronger) inductive hypothesis. While the following theorem
```
dsainati1:
```suggestion
(stronger) inductive hypothesis. While the following
```

ionathanch:
How do we feel about ending textual proofs with ∎? (Typed using `\qed`)
If we do use qed (the abbreviation) I think it should be _QED_ and not _Qed_, which is very Rocqesque

dsainati1:
∎ looks to me like the tofu symbol that shows up when your browser can't display a unicode character, so I would prefer to avoid that one. Using QED instead of Qed makes sense to me though.

ionathanch:
> ∎ looks to me like the tofu symbol

1950s mathematics textbooks in shambles

bcpierce00 (approved):

PR #228, opened 2026-08-11
Polish IndProp

KihongHeo:
I went over the file and fixed the following:
- Change names
- Remove duplicate sentences
- Remove guard_msgs
- Polish other parts

Let me know if you have any other suggestions.

PR #234, opened 2026-08-11
Use Lean terminology for types

### LINKED ISSUE #226: Check usage of the word "set"
The usage of the word "set" is different in Lean compared to Rocq. Rocq's standard library Ensemble := α → Prop is called Set in Mathlib.

Someone should make a careful pass through everything to eliminate incorrect usage of "set".

Robertboy18:
I read through the current sources for places where “set” was still being used in the Rocq sense of a whole datatype.

This changes the clear cases to “type” or “value,” while leaving genuine mathematical sets, simp sets, sets of rules, and ordinary uses unchanged.

A couple of these lines overlap with #213 and #228, since those PRs edit the same files.

Closes #226.

### DIFF
--- a/LF/Basics.lean
+++ b/LF/Basics.lean
@@ -101,8 +101,8 @@ A datatype definition:
 :::
 
 ::::full
-Let's start with a very simple example.  The following declaration tells
-Lean that we are defining a set of data values, i.e. a _type_.
+Let's start with a very simple example.  The following declaration introduces
+a new type and lists its possible values.
 ::::
 
 ```lean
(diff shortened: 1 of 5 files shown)


dsainati1 (changes requested):
I don't think these usages of "set" are really Rocq-specific. The intuition of a type as a set of values is useful and common beyond just Rocq, and I don't think needs to be removed as part of the switch to Lean.

bcpierce00:
I actually tend to agree with Daniel here.  Most of these rewordings are fine, but a few are a bit awkward, and I'm not sure the 100% terminological purity is worth it.

PR #235, opened 2026-08-12
Polish the Logical Foundations postscript

Robertboy18:
I gave the LF Postscript a focused pass so it reads as a finished chapter rather than a translated draft. This removes the remaining scaffolding, tightens the recap, points readers toward the next two volumes, and replaces the old Rocq links with current Lean resources.

Tested with:
- `lake build LF.Postscript`
- `make lf-student`
- `make style-check`

### DIFF
--- a/LF/Postscript.lean
+++ b/LF/Postscript.lean
@@ -9,27 +9,23 @@ htmlSplit := .never
 file := some "Postscript"
 %%%
 
-:::dev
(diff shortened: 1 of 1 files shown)


bcpierce00 (approved):

bcpierce00:
#Note: I think we want this instead of #237

PR #237, opened 2026-08-12
bib: add entry and citations

luisacicolini:
Adds a missing entry for the preface and citations in both preface and postscript

### DIFF
--- a/Bib.lean
+++ b/Bib.lean
@@ -139,6 +139,15 @@ def halpern2001: Article where
   year := 2001
   month   := none
 
+def manna1971: Article where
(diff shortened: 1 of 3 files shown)


bcpierce00 (approved):

bcpierce00:
Thank you!

We should update the Postscript's reading list a bit too...
- Add the Lean books that are recommended on the Lean website
- See if there are any more recent books on functional programming that should be included
- Remove the Rocq-specific books and maybe just point to an external list of resources for Rocq

bcpierce00:
#Note: Actually, this is superseded by #235.

PR #241, opened 2026-08-12
Remove `abbrev`s (#189)

### LINKED ISSUE #189: Convert abbreviations to definitions
Discussion on zulip converged to the following key points:
- All functions that don't eliminate into Prop, i.e. functions that return a value rather than a type, should be definitions. Having characterizing lemmas is preferred/idiomatic but not necessary for simple ad-hoc definitions.
- Functions that eliminate into Prop (or Type), i.e. those that return a type rather than a value, can be abbreviations, but we should avoid defining new API for them. For example, `abbrev Even (x : Nat) := ∃ n : Nat, x = double n` is an ad hoc Prop-valued function. If it needs API like `Even.foo`, it should be a `def` instead. This is partly because dot-notation doesn't always work intuitively with terms of a type that is an abbrev.

These points, and concepts like reducibility and transparency, are not something we should explain to the students, especially in Logical Foundations. The words "elaboration" and "reducibility" should intentionally appear nowhere in this volume. The way we introduce abbreviations (in Logic.lean?) and hand-wave around these topics is by saying that "abbreviations are used for type aliases/synonyms, but not for objects that have novel API". Here's @chenson2018 's suggestion for the student-facing explanation given after the first abbreviation

> Here we have used `abbrev`, which works very similarly to `def` but indicates a light layer of abstraction. The most practical difference in this volume is that we will not follow these up [with] characterizing lemmas, as we expect these to [be treated as synonyms for types] for which we've already written these lemmas. You'll see `abbrev` used for some ad hoc statements, especially simple propositions that we are unlikely to use often.

We should convert those abbreviations that do not conform to these points into `def`s.

ionathanch:
Re [#lean-software-foundations-contributors > Abbreviations in Logic.lean](https://leanprover.zulipchat.com/#narrow/channel/607217-lean-software-foundations-contributors/topic/Abbreviations.20in.20Logic.2Elean/with/614061239), closes #189 trivially.

### DIFF
--- a/LF/Logic.lean
+++ b/LF/Logic.lean
@@ -2775,14 +2775,14 @@ The following reasoning principle is _not_ derivable with the tools we've seen s
 ::::
 
 ```lean
-def excluded_middle := ∀ a : Prop, a ∨ ¬ a
+def ExcludedMiddle := ∀ a : Prop, a ∨ ¬ a
 ```
 
 ::::full
 To understand operationally why this is the case, recall that,
 to prove a statement of the form {lean}`a ∨ b`, we use the {tactic}`left` and {tactic}`right`
 tactics, which effectively require knowing which side of the disjunction
-holds. But the universally quantified {lean}`a` in {lean}`excluded_middle` is an
+holds. But the universally quantified {lean}`a` in {lean}`ExcludedMiddle` is an
 _arbitrary_ proposition, which we know nothing about. We don't have enough
 information to choose which of {tactic}`left` or {tactic}`right` to apply.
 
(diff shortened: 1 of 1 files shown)


LF/Logic.lean
```diff
@@ -2993,15 +2993,15 @@ why do the solutions not do this?
...
+def not_not := ∀ a : Prop, ¬ ¬ a → a
 
-abbrev de_morgan_not_and_not := ∀ a b : Prop, ¬ (¬ a ∧ ¬ b) → a ∨ b
+def de_morgan_not_and_not := ∀ a b : Prop, ¬ (¬ a ∧ ¬ b) → a ∨ b
 
-abbrev imp_or := ∀ a b : Prop, (a → b) → (¬ a ∨ b)
+def imp_or := ∀ a b : Prop, (a → b) → (¬ a ∨ b)
 
-abbrev consequentia_mirabilis := ∀ a : Prop, (¬ a → a) → a
+def consequentia_mirabilis := ∀ a : Prop, (¬ a → a) → a
```
berberman: I think we should also make them in camel case, e.g. https://github.com/lean-forward/logical_verification_2025/blob/76421700cb010aeddeb2b2a4a93e538b92b0e585/lean/LoVe/LoVe03_BackwardProofs_ExerciseSheet.lean#L114-L121

dsainati1 (approved):

PR #242, opened 2026-08-12
pass on Preface and citations

luisacicolini:
I added some projects and did a pass on the preface, let me know if we should have more/less


bcpierce00:
This is great.  My only suggestion is to try merging the specific Lean projects into the bulleted list below...

LF/Preface.lean
```diff
@@ -119,37 +116,39 @@ community of users in both research and at companies like DeepMind,
...
   obtaining the highest level of common criteria certification,
   and for formal specifications of the x86 and LLVM instruction
   sets and programming languages such as C.
 
 - As _environments for developing formally certified software
   and hardware_, they have been used, for example, to build
-  CompCert, a fully-verified optimizing compiler for C, and
-  CertiKOS, a fully verified hypervisor, for proving the
+  CompCert{citet Bib.leroy2016}[], a fully-verified optimizing compiler for C, and
+  CertiKOS{citet Bib.gu2016certikos}[], a fully verified hypervisor, for proving the
```
mwhicks1: You might also cite the Cedar effort:
- https://mhicks.me/papers/disselkoen24vgd.html (paper)
- Repo https://github.com/cedar-policy/cedar-spec/tree/main/cedar-lean

mwhicks1 (approved):

LF/Preface.lean
```diff
@@ -73,17 +73,14 @@ Logic is the field of study whose subject matter is _proofs_ --
 unassailable arguments for the truth of particular propositions.
 Volumes have been written about the central role of logic in
 computer science.  Manna and Waldinger called it "the calculus of
-computer science," while Halpern et al.'s paper _On the Unusual
-Effectiveness of Logic in Computer Science_ catalogs scores of
+computer science,"{citet Bib.manna1971}[] while Halpern et al.'s paper _On the Unusual
```
Robertboy18: These `{citet}` roles repeat authors already present in the prose and run into the preceding text in the rendered page (`CompCertLeroy et al. (2016)`, for example). Could we rephrase the Manna/Halpern sentence so the textual citation supplies the author name, and use `{citep}` after CompCert and CertiKOS?

Bib.lean
```diff
@@ -129,6 +129,44 @@ def winskel1993 : Article where
...
+def manna1971: Article where
+  title := inlines!"Toward automatic program synthesis"
+  authors := #[inlines!"Zohar Manna", inlines!"Richard J Waldinger"]
+  journal := inlines!"Communications of the ACM"
+  volume := inlines!"14"
+  number  := inlines!"3"
+  year := 1971
+  month := none
+
+def leroy2016: Article where
```
Robertboy18: Could these two entries use `InProceedings`? Both are conference papers, and representing them as `Article` currently leaves an empty volume field in the rendered citation. Also, there’s a leading space before `Zhong Shao`, and the official spelling is `Vilhelm Sjöberg`:)! Again just being a bit precise!

LF/Preface.lean
```diff
@@ -119,37 +116,39 @@ community of users in both research and at companies like DeepMind,
...
 
-Lean and its relatives have become critical enablers for a huge
-variety of work across computer science and mathematics:
-:::dev "Benjamin Pierce (bcpierce00)"
-Need more examples, including specifically things accomplished in Lean.
+Lean and its relatives have become critical enablers for a [huge
+variety of work](https://leanprover-community.github.io/papers.html) across computer science and mathematics:
+- formalizing the proof of [Fermat's last theorem](https://github.com/ImperialCollegeLondon/FLT)
+- formalizing the [Sphere packing problem](https://github.com/thefundamentaltheor3m/Sphere-Packing-Lean)
+- Deepmind's AI model for IMO problems: [Alphaproof](https://deepmind.google/blog/ai-solves-imo-problems-at-silver-medal-level/)
```
Robertboy18: Small naming fix: `DeepMind` and `AlphaProof`.

LF/Preface.lean
luisacicolini: added!

Bib.lean
luisacicolini: fixed and double checked for all the entries I added, thanks for the catch :) for whatever reason I thought the only available constructor was article

LF/Preface.lean
luisacicolini: fixed

LF/Preface.lean
luisacicolini: should be fixed now :)

luisacicolini:
thank you all for the reviews :) should be good to go now!

bcpierce00:
One last comment around line 135...

bcpierce00:
So sorry -- my git viewer was not showing me the right thing.  :-(

Yes, this is fine.  I'll merge.

bcpierce00 (approved):

PR #244, opened 2026-08-13
Document `inversion`.

rogerburtonpatel:
I adapted the paragraph from the old `INPROGRESS.md` which justifies the `inversion` tactic and put it in `CustomTactics.lean`. This seems sufficient to me for now.


LF/CustomTactics.lean
```diff
@@ -18,6 +18,18 @@ meta section
 namespace Lean.Parser
 open Tactic
 
+
+/- n Lean, "Inversion" on `h: f n ≤ 0`, where `f : nat -> nat`,
```
ionathanch: **I**n Lean

ionathanch: also "**i**nversion" on...

ionathanch:
I think the comment should be a doc comment of some sort attached to maybe the example it's talking about at the bottom? rather than floating at the top of the file, since this file doesn't *only* implement the `inversion` tactic (notably, `apply ... at ...`).

dsainati1:
While we're at it, we should write a Lean doc comment(i.e., with `/--`) for this tactic so that when we hover over it in Verso output generated from `{tactic}inversion` or in the Lean IDE, students will see a summary of how to use the tactic.

ionathanch:
There already is one! It's this doc comment here: https://github.com/plclub/sf-in-lean/blob/main/LF/CustomTactics.lean#L30
It's worded like docs for tactics in general (so not catering to our students specifically), but we can add an example usage or a usage hint (e.g. "Use when `cases` fails with "Dependent elimination failed..."")[^1]

image

[^1]: tbh I would prefer the hint to appear where the error occurs, like an additional "Did you mean...?", but idk if it's possible to modify the error messages of an existing tactic

ionathanch:
I've moved the doc comment down with some edits and also grouped tests together with brief doc comments on what they were testing, I think it's ready for review now

rogerburtonpatel:
@ionathanch @dsainati1 please check if this can't be merged; I will merge it in an hour otherwise.

dsainati1 (approved):

PR #245, opened 2026-08-14
README: On SFL build structure, expected use

mwhicks1:
Answering questions from a curious outsider made me realize that you shouldn't have to go through `CONTRIBUTING.md` just to understand how the repo is organized, what `make` produces, and how to use it. So I lifted some and adjusted some text from `CONTRIBUTING.md` to the README to explain that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ per-chapter build target: a whole volume is the smallest unit.)
 The first build compiles the whole Lean toolchain's dependencies and
 takes a while; later builds are incremental.
 
-## Orientation
+## Repository organization
 
 Each volume has its own top-level directory - `LF`, `HL`, and `TS` - each
 of which is paired with a top-level `.lean` file.
@@ -51,7 +51,9 @@ written in [Verso](https://verso.lean-lang.org/) format.
 
 Running `make` at the top level produces, for each volume, four
 different ready-for-distribution outputs in a temporary top-level
-`_out` directory, each with both `.lean` and `.html` variants.
+`_out` directory, each with both `.lean` and `.html` variants (in
+`_out/<vol>/<variant>/lean/` and `_out/<vol>/<variant>/html-multi/`
+respectively).
 
 - **student**   (full prose, solutions elided)
 - **solutions** (full prose, solutions shown)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -44,8 +44,37 @@ takes a while; later builds are incremental.
...
 ## Orientation
 
-For everything else — repo layout, conventions, PR workflow — see
-[CONTRIBUTING.md](CONTRIBUTING.md).
+Each volume has its own top-level directory - `LF`, `HL`, and `TS` - each
+of which is paired with a top-level `.lean` file.
+Within each directory there are multiple `.lean` files, one per chapter,
+written in [Verso](https://verso.lean-lang.org/) format.
+
+Running `make` at the top level produces, for each volume, four
```
Robertboy18, line 52: `make serve` depends on `all`, and each volume target also includes `grading`, so it builds four variants. Could we update the earlier quick-start sentence from "all three variants — student / solutions / terse" to include `grading` as well?

README.md
```diff
@@ -44,8 +44,37 @@ takes a while; later builds are incremental.
...
+- **student**   (full prose, solutions elided)
+- **solutions** (full prose, solutions shown)
+- **terse**     (little prose, no solutions, workinclass elided;
+                 for lecturing)
+- **grading**   (solutions variant with automated grading support,
+                 for instructors)
+
+Students are expected to work through the student `.lean` versions,
+filling in the exercises, or to go through the HTML and switch
+to the Lean just for exercises. See [ALPHATESTERS.md](AlPHATESTERS.md)
```
Robertboy18: The link target has a lowercase `l` (`AlPHATESTERS.md`), while the file is `ALPHATESTERS.md`; GitHub treats these paths as case-sensitive. Could we fix this occurrence and the same link near the top of the README?

README.md
mwhicks1: Good catch.

ionathanch:
Should we remove the _Repository organization and Makefile targets_ section from CONTRIBUTING.md then? Just so the information isn't duplicated and doesn't potentially diverge if it changes in the future

PR #247, opened 2026-08-14
DHS polishing pass over typeclasses

dsainati1:
Polishing pass over typeclasses, including:
* An exercise about monoids
* An exercise about groups
* Removing some cruft

### DIFF
--- a/LF/Typeclasses.lean
+++ b/LF/Typeclasses.lean
@@ -532,15 +532,26 @@ about monoids is that the identity element of a monoid is unique. That is,
 if we have two monoids over the same set with the same operator, their identity elements must also
 be the same:
 
-:::dev "Daniel Sainati @dsainati1"
-I don't know why but the infoview for this proof is extremely confusing. How to set
-things up to be clearer?
-:::
-
 ```lean
 theorem id_unique {α : Type} {m₁ m₂ : Monoid α} (h : m₁.op = m₂.op) : m₁.id = m₂.id := by
-  rw [←m₁.left_id m₂.id, h, m₂.right_id]
+  obtain @⟨op₁, id₁, left_id₁, right_id₁, assoc₁⟩ := m₁
+  obtain @⟨op₂, id₂, left_id₂, right_id₂, assoc₂⟩ := m₂
+  have h' : id₁ = id₂ := by
+    -- this introduces a use of m₁'s operator
+    rw [← left_id₁ id₂]
+    -- we use our hypothesis to rewrite m₁'s operator into m₂'s
+    rw [h]
+    -- then, we can use m₂'s right id
+    rw [right_id₂]
+  exact h'
 ```
+
+In the above proof, we can destructure the monoid instances `m₁` and `m₂` with the `obtain` tactic
+we saw in the {ref "Logic"}[Logic] chapter. When we do so however, because these are
+class instances instead of normal structures, we preface our tuple with the `@` symbol.
+When stepping through the above proof, if the notation is confusing to you,
+remember that you can set `set_option pp.all true` or `set_option pp.explicit true`
+to make Lean show you more clearly what is going on.
 
 A _group_ is a special kind of monoid with an _inverse_ operation `inv`, which has the property that
 `∀ x, inv x ⊗ x = id = x ⊗ inv x`. We can extend the definition of a {name}`Monoid` to capture this
@@ -583,11 +594,11 @@ This also prints weird.
 Two groups defined with the same operation over the same set must have the same inverse as well.
 
 ```lean
-theorem inv_unique {α : Type} {m₁ m₂ : Group α} (h : m₁.op = m₂.op) : m₁.inv = m₂.inv := by
+theorem inv_unique {α : Type} {g₁ g₂ : Group α} (h : g₁.op = g₂.op) : g₁.inv = g₂.inv := by
   solution!
     ext a
-    rw [←m₁.right_id (Group.inv a), ←m₁.right_inv a]
-    rw [m₁.assoc, h, m₂.left_inv a, m₂.left_id]
+    rw [←g₁.right_id (Group.inv a), ←g₁.right_inv a]
+    rw [g₁.assoc, h, g₂.left_inv a, g₂.left_id]
 ```
 ::::
 
--- a/LF/Typeclasses.lean
+++ b/LF/Typeclasses.lean
@@ -553,6 +553,7 @@ class instances instead of normal structures, we preface our tuple with the `@` 
 When stepping through the above proof, if the notation is confusing to you,
 remember that you can set `set_option pp.all true` or `set_option pp.explicit true`
 to make Lean show you more clearly what is going on.
+For example, the goal is displayed as `Monoid.id = Monoid.id` since the instances `m₁` and `m₂` are implicit arguments to `Monoid.id`. Setting `pp.explicit true` displays the goal as `@Eq α (@Monoid.id α m₁) (@Monoid.id α m₂)`.
 
 A _group_ is a special kind of monoid with an _inverse_ operation `inv`, which has the property that
 `∀ x, inv x ⊗ x = id = x ⊗ inv x`. We can extend the definition of a {name}`Monoid` to capture this
--- a/LF/Typeclasses.lean
+++ b/LF/Typeclasses.lean
@@ -1180,37 +1180,17 @@ instance  {α β : Type} : MyGetElem (PartialMap α β) α (Option β) where
 theorem getElem_def  {α β : Type} (m : PartialMap α β) (a : α) : m[a] = m.toTotal[a] := rfl
 ```
 
-:::dev "Niklas Halonen (xhalo32)" NOW
-The following few paragrahps are out-of-date because `TotalMap` is also a structure.
-:::
-
 Remember that we discussed earlier with total maps that using accessing the `inner` field
 and performing function application application exposes the implementation,
 and that's why we introduced a new notation {name}`MyGetElem`?
-Here we take extend that concept, and instead of using a `def` for partial maps, like this...
+Here we extend that concept, and instead of using a `def` for partial maps, like this...
 
 ```display
 def PartialMap (α : Type) (β : Type) := TotalMap α (Option β)`
 ```
 
 ...we define partial maps as a structure containing a total map.
-This more strongly hides the fact that it's a total map.
-
-Now, the type system doesn't consider `PartialMap α β` to be definitionally equal to `TotalMap α (Option β)`,
-so the following equality doesn't type check:
-
-```lean -keep +error (name := empty_eq)
-example  {α β : Type} : (∅ : PartialMap α β) = (∅ : TotalMap α (Option β)) := by rfl
-```
-
-```leanOutput empty_eq
-Type mismatch
-  ∅
-has type
-  TotalMap α (Option β)
-but is expected to have type
-  PartialMap α β
-```
+This more strongly hides the fact that it's implemented using a total map.
 
 Updating a partial map at a key means storing a {name}`some` value there.
 To update, we create a new partial map from `a →ₜ some b ; m.toTotal` by wrapping it in angle brackets, i.e. using the anonymous constructor syntax.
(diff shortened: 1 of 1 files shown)


LF/Typeclasses.lean
```diff
@@ -440,6 +445,81 @@ instance : HasThree Nat where
...
+* an operator (which we'll write `+`) that combines two elements of type `α` into one,
+* a particular element `id` of type `α`, which we call the "identity element", and
+* some laws about the interaction of `+` and `id`, namely that:
+    * `∀ x, id + x = x = x + id`, and
+    * `∀ x y z, x + (y + z) = (x + y) + z` (i.e., that `+` is associative)
+
+We can express these requirements in the form of a typeclass:
+
+:::dev "Daniel Sainati @dsainati1"
+Is there a way to use infix `+` notation for `op` in the definition of the typeclass?
```
chenson2018: Not directly. The way this is done is that there is a first class that has just the data you want notation for, then extend these. See [`Semigroup`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Group/Defs.html#Semigroup) as the simplest example. (Or [`Monoid`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Group/Defs.html#Monoid) too of course, it extends more classes)

LF/Typeclasses.lean
```diff
@@ -440,6 +445,81 @@ instance : HasThree Nat where
...
+  left_identity := by lia
+  right_identity := by lia
+  associativity := by lia
+```
+
+::::exercise (rating := 1) (name := "NatMonoidMul")
+However, multiplication on `Nat`s _also_ forms a monoid. What is its identity element?
+
+```lean
+instance : Monoid Nat where
```
chenson2018, line 504: Note that we're breaking a rule here about having diamonds on data carrying typeclasses, something we do explain above. The way this works in Mathlib is that there are additive and multiplicative variants of the classes, e.g. [`Monoid`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Group/Defs.html#Monoid) versus [`AddMonoid`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Group/Defs.html#AddMonoid).

Maybe just leave a dev note if this is an isolated problem for now, not sure if/how we want to talk about this.

LF/Typeclasses.lean
```diff
@@ -440,6 +445,162 @@ instance : HasThree Nat where
...
+if we have two monoids over the same set with the same operator, their identity elements must also
+be the same:
+
+:::dev "Daniel Sainati @dsainati1"
+I don't know why but the infoview for this proof is extremely confusing. How to set
+things up to be clearer?
+:::
+
+```lean
+theorem id_unique {α : Type} {m₁ m₂ : Monoid α} (h : m₁.op = m₂.op) : m₁.id = m₂.id := by
```
xhalo32, line 534: `{m₁ m₂ : Monoid α}` should be `[m₁ m₂ : Monoid α]` however it doesn't make sense to me to define `Monoid` as a class, if we are going to have multiple instances (m₁, m₂) of it. One option could be to add `op` as a type parameter to the class. Then we could write `[m₁ : Monoid α op₁]` and `[m₂ : Monoid α op₂]`.

xhalo32: I'll try this and post it in a branch. Immediately ran into the issue that it's difficult to get the `⊗` notation to work.

xhalo32: I have (tried to) addressed this (plus the diamond problem) in https://github.com/plclub/sf-in-lean/pull/248

The point boils down to the fact that for an unambiguous `⊗` notation to work, it has to be unique per carrier type. If we have multiple monoid structures it can't be, so I had to add new notation.

chenson2018: Sorry to be picky, but I have some issues with this approach... let me try fixing myself rather than just backseat driving, I have a bit of time now.

chenson2018: Okay, I've thought about this PR in general for the last hour or so. I think there is some confusion from both Niklas and Daniel that I hopefully can clear up.

First, it is *not* the case that every theorem involving typeclasses must use them as instance implicts. Usually this is the case, but examples like this are precisely when we might not, where we want to prove something about two data-carrying classes. The [Mathlib version](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Group/Defs.html#MulOneClass.ext) of this proof for referenence does this as well.

The fact that the goal prints as `Monoid.id = Monoid.id` is inherent here: you're comparing the equality of two fields of a typeclass, it is expected that they should print identically. (I'm curious, is this not the same as Rocq??) You can again point the student toward `pp.all` so that they can see `@Algebra.Monoid.id α m₁` versus `@Algebra.Monoid.id α m₂`. You can also structure the proof like:

```lean
theorem id_unique {α : Type} : ∀ {m₁ m₂ : Monoid α}, m₁.op = m₂.op → m₁.id = m₂.id := by
  rintro @⟨⟨op₁⟩, id₁, left_id₁, right_id₁, assoc₁⟩ @⟨⟨op₂⟩, id₂, left_id₂, right_id₂, assoc₂⟩ h
  have eq : id₁ = id₂ := by rw [← left_id₁ id₂, h, right_id₂]
  exact eq
```

so that they can clearly see in the infoview a differentiation between the two monoids in the hypothesis.  

Regarding the problem with duplicate additive and multiplicative instances, I think I am ultimately fine with the compromise that this PR makes. Other resources like the [MIL chapter that discusses monoids](https://leanprover-community.github.io/mathematics_in_lean/C09_Groups_and_Rings.html) just use the actual Mathlib versions, but I think that for this context, audience, etc. that it makes sense to not dwell on this.

LF/Typeclasses.lean
```diff
@@ -440,6 +445,162 @@ instance : HasThree Nat where
...
+class OpSet (α : Type) where
+  op : α → α → α
+
+infixr:70 " ⊗ " => OpSet.op
+
+class Monoid (α : Type) extends (OpSet α) where
+  id : α
+  left_identity : ∀ (x : α), id ⊗ x = x
+  right_identity : ∀ (x : α), x ⊗ id = x
+  associativity : ∀ (x y z : α), x ⊗ (y ⊗ z) = (x ⊗ y) ⊗ z
```
chenson2018: Nit: these names are so long! I'd do

```suggestion
  left_id : ∀ (x : α), id ⊗ x = x
  right_id : ∀ (x : α), x ⊗ id = x
  assoc : ∀ (x y z : α), x ⊗ (y ⊗ z) = (x ⊗ y) ⊗ z
```
which is similar to existing theorems.

LF/Typeclasses.lean
dsainati1: Turning on `pp.all` ends up being quite confusing, and `pp.explicit` still prints more information than the user probably wants to see. I really just want the goal to print as `@Algebra.Monoid.id α m₁ = @Algebra.Monoid.id α m₂`, but with `pp.explicit` I get `@Eq α (@Monoid.id α m₁) (@Monoid.id α m₂)`. It gets worse after the rewrites because `pp.explicit` also means it doesn't print using notation. 

The way that I had conceptualized this proof in my head was that we had two different instances of the Monoid typeclass, but where the `op` and type fields were equal, and then we wanted to prove the equality of the `id` fields as well. `Monoid.id = Monoid.id` as a goal not being solvable with `rfl` is going to be very confusing, I think.

chenson2018: I'm not suggesting that you use `pp.all` for the whole proof, but remind the student that they can use this to inspect the term and the underlying term. I don't think there is any option to only partially turn off normally printing like you are asking for.

If you look at the proof I've written above, in the `have` you're proving a version where it's `id₁ = id₂`. I did this so that the student is not working with `Monoid.id = Monoid.id` in the goal. If you want to change it directly in the goal, you can write:

```lean
theorem id_unique {α : Type} : ∀ {m₁ m₂ : Monoid α}, m₁.op = m₂.op → m₁.id = m₂.id := by
  rintro @⟨⟨op₁⟩, id₁, left_id₁, right_id₁, assoc₁⟩ @⟨⟨op₂⟩, id₂, left_id₂, right_id₂, assoc₂⟩ h
  change id₁ = id₂
  rw [← left_id₁ id₂, h, right_id₂]
```

or 

```lean
theorem id_unique {α : Type} : ∀ {m₁ m₂ : Monoid α}, m₁.op = m₂.op → m₁.id = m₂.id := by
  rintro @⟨⟨op₁⟩, id₁, left_id₁, right_id₁, assoc₁⟩ @⟨⟨op₂⟩, id₂, left_id₂, right_id₂, assoc₂⟩ h
  suffices id₁ = id₂ by simp [this]
  rw [← left_id₁ id₂, h, right_id₂]
```

dsainati1:
For some reason adding an `import LF.Automation` to this file breaks the build in `HL`?

```
LF/Basics.lean:1:0: unknown module prefix 'SFLMeta'
```

@bcpierce00 Seems like another issue with the Verso build process again?

EDIT: I have removed the dependency for now and repeated the definitions I was importing (from `UsingLean` and `Logic`) in this file. Ideally we would not have to do this, but for now I think it's okay to unblock this.

LF/Typeclasses.lean
```diff
@@ -1201,46 +1164,41 @@ structure PartialMap (α : Type) (β : Type) where
...
+and performing function application application exposes the implementation,
+and that's why we introduced a new notation {name}`MyGetElem`?
+Here we take extend that concept, and instead of using a `def` for partial maps, like this...
 
 ```display
 def PartialMap (α : Type) (β : Type) := TotalMap α (Option β)`
 ```
 
-...we define partial maps as a structure containing just a total map.
+...we define partial maps as a structure containing a total map.
```
xhalo32: I think all of this (up to line 1212) can be removed since `TotalMap` is a structure so there's no reason to explain why we don't make `PartialMap` a `def`.

LF/Typeclasses.lean
```diff
@@ -440,52 +443,190 @@ instance : HasThree Nat where
...
+  obtain @⟨op₁, id₁, left_id₁, right_id₁, assoc₁⟩ := m₁
+  obtain @⟨op₂, id₂, left_id₂, right_id₂, assoc₂⟩ := m₂
+  have h' : id₁ = id₂ := by
+    -- this introduces a use of m₁'s operator
+    rw [← left_id₁ id₂]
+    -- we use our hypothesis to rewrite m₁'s operator into m₂'s
+    rw [h]
+    -- then, we can use m₂'s right id
+    rw [right_id₂]
+  exact h'
```
xhalo32, line 545:
```suggestion
  -- the goal `m₁.id = m₂.id` is equivalent with `id₁ = id₂` even though it displays `Monoid.id = Monoid.id`
  exact h'
```

LF/Typeclasses.lean
```diff
@@ -440,52 +443,190 @@ instance : HasThree Nat where
...
+```
+
+In the above proof, we can destructure the monoid instances `m₁` and `m₂` with the `obtain` tactic
+we saw in the {ref "Logic"}[Logic] chapter. When we do so however, because these are
+class instances instead of normal structures, we preface our tuple with the `@` symbol.
+When stepping through the above proof, if the notation is confusing to you,
+remember that you can set `set_option pp.all true` or `set_option pp.explicit true`
+to make Lean show you more clearly what is going on.
+
```
xhalo32, line 555:
```suggestion
For example, the goal is displayed as `Monoid.id = Monoid.id` since the instances `m₁` and `m₂` are implicit arguments to `Monoid.id`. Setting `pp.explicit true` displays the goal as `@Eq α (@Monoid.id α m₁) (@Monoid.id α m₂)`.

```

LF/Typeclasses.lean
```diff
@@ -1091,19 +1210,6 @@ but is expected to have type
   PartialMap α β
 ```
```
xhalo32: I think all of this can be removed since `TotalMap` is a structure so there's no reason to explain why we don't make `PartialMap` a `def`.

xhalo32 (approved):

xhalo32:
This PR now looks good to me.

I made some clarifications about the role of encapsulation and `MyGetElem` in its own PR in https://github.com/plclub/sf-in-lean/pull/255

mwhicks1 (approved):

PR #248, opened 2026-08-15
Typeclasses: make Monoid a structure, add ⊗_[m] notation

xhalo32:
This addresses:
- [Chris's note about diamonds](https://github.com/plclub/sf-in-lean/pull/247/changes#r3787004036)
- Confusing infoView in `id_unique`

Plus I added `op_eq_of_id_eq` which I came up with as a fun exercise for working with different monoid structures on the same carrier.

Note that the code doesn't compile since I didn't rewrite the prose around `OpSet`. I think extending `Group` should be straight-forward but I didn't do that yet since we're not sure if this is a good approach for the monoids.

### DIFF
--- a/LF/Typeclasses.lean
+++ b/LF/Typeclasses.lean
@@ -467,13 +467,7 @@ structure Monoid (α : Type) where
   id : α
 
 namespace Monoid
-scoped notation:70 x:71 " ⊗_[" m "] " y:70 => Monoid.op m x y
-scoped notation:70 x:71 " ⊗ " y:70 => Monoid.op _ x y
-
-@[app_unexpander op]
-def unexpandOp : Lean.PrettyPrinter.Unexpander
-  | `($_ $m $x $y) => ``($x ⊗_[$m] $y)
-  | _ => throw ()
+scoped notation:70 x:71 " ⊗_[" m "] " y:70 => op m x y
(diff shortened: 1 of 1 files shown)


LF/Typeclasses.lean
```diff
@@ -462,17 +462,39 @@ let's use a typeclass to define a _monoid_, a simple algebraic structure that in
 We can express these requirements in the form of a typeclass:
 
 ```lean
--- first we define a notation typeclass for our operator ·
-class OpSet (α : Type) where
+structure Monoid (α : Type) where
```
chenson2018, line 465: While I do get the appeal of following the `Lawful` naming, I find diverging from Mathlib here confusing. I think this can be done also without any need for unexpanders. Give me a bit and I'll try it myself.

xhalo32: Diverging in the sense that we don't have `Add/Mul`?

You are right, it doesn't need an unexpander, it was left over from my earlier testing (I went through like four iterations). I'll remove it.

chenson2018:
Please see my comment at https://github.com/plclub/sf-in-lean/pull/247#discussion_r3789441631. Both mathematically and technically, I would not take this approach of indexing the notation by the monoid.

PR #255, opened 2026-08-16
Typeclasses: explain `MyGetElem` and encapsulation better

xhalo32:
- Added `PartialMap.toTotal_def` plus `emptyNatMap` examples
- Added `PartialMap.toTotal_eq_getElem`
- Removed mention of `def PartialMap`
- Added examples that use the `PartialMap` simp lemmas
- Made some simp lemmas use `rfl` instead of `by rfl` to be `dsimp` compatible

### DIFF
--- a/LF/Typeclasses.lean
+++ b/LF/Typeclasses.lean
@@ -693,12 +693,14 @@ def emptyNatMap : TotalMap Nat Nat := ∅
 
 ### Getting Elements
 
-While `TotalMap`s happen to be implemented as functions under the hood,
-we would prefer not to expose this
-fact in its public interface. Accordingly, we define new operations for querying and
-updating mappings. As a first attempt at a query operation, playing the role that `find` played for the {ref "Lists"}[Lists] chapter's
-list-based maps, we could define a function
-`get` for getting the value associated with a key:
+:::dev "Niklas Halonen (xhalo32)"
+I have clarified here that `MyGetElem` has nothing to do with the encapsulation (`inner` vs `get`).
+It's simply notation that expands to the public API (`get`).
+:::
+
+While `TotalMap`s happen to be implemented as functions under the hood, we would prefer not to expose this fact in its public interface.
+Accordingly, we define new operations for querying and updating mappings.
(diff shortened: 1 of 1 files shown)


LF/Typeclasses.lean
```diff
@@ -1170,35 +1175,46 @@ structure PartialMap (α : Type) (β : Type) where
...
-```display
-def PartialMap (α : Type) (β : Type) := TotalMap α (Option β)`
+@[simp]
+theorem toTotal_eq_getElem {α β : Type} (m : PartialMap α β) (a : α) :
+    m.toTotal[a] = m[a] := rfl
 ```
 
-...we define partial maps as a structure containing a total map.
-This more strongly hides the fact that it's implemented using a total map.
+Remember that we discussed earlier with total maps that accessing the `inner` field and performing function application application exposes the implementation, and that's why we introduced the API {name}`TotalMap.get`?
```
mwhicks1: Typo ("application application") and the sentence structure is a bit hard to follow. Proposed rewrite:

We previously defined {name}`TotalMap.get` so that users can retrieve elements from a `TotalMap` in a manner independent of its actual implementation, which is a function stored in `TotalMap.inner`.

mwhicks1 (approved):
Looks good.

LF/Typeclasses.lean
```diff
@@ -693,12 +693,14 @@ def emptyNatMap : TotalMap Nat Nat := ∅
 
 ### Getting Elements
 
-While `TotalMap`s happen to be implemented as functions under the hood,
-we would prefer not to expose this
-fact in its public interface. Accordingly, we define new operations for querying and
-updating mappings. As a first attempt at a query operation, playing the role that `find` played for the {ref "Lists"}[Lists] chapter's
-list-based maps, we could define a function
-`get` for getting the value associated with a key:
+:::dev "Niklas Halonen (xhalo32)"
```
dsainati1: IMO we don't need to add dev blocks for these kinds comments. As I understand it, dev comments are like "TODO" notes for us to come back to and fix. If we need documentation of something in the text, maybe :::instructors is better?

dsainati1 (approved):
Looks good to me modulo Mike's and my comments

xhalo32:
Thanks for the fast reviews! I have addressed both points.

PR #256, opened 2026-08-16
Fix infinite regress in Imp delaborator

david-christiansen:
Previously, saturated applications of `Com.asgn` that didn't match one of the cases that is expressible in the grammar would be elaborated as antiquotations. Unfortunately, the antiquotation elaboration itself was problematic, because it directly invoked `delab` which returned to the very same case.

This meant that the delaborator would hit its max recursion depth, and emit an ellipsis instead of an understandable term.

One example of a case not expressible in the grammar is when a free Lean variable stands for the Imp variable to be assigned.

An alternative solution here is to also add an explicit quotation and escape mechanism for names. Instead of using `ident` for the LHS of the assignment, you could use a custom parser like `varname` that can be `ident` or `"~" term`; this would allow the antiquotation from the inexpressible term to be smaller in some cases, but it does introduce some precedence subtlety if you use the same antiquotation operator. This PR goes with the smaller fix.

I'm not familiar enough with the repository to know where to put a regression test, but here's an example that demonstrates the fix:

    #check fun (x : Ident) (a : Aexp) => Com.asgn x a

Without this fix, that line outputs:

    fun x a => ⋯ : Ident → Aexp → Com

With the fix, it outputs:

    fun x a => (Com.asgn x a) : Ident → Aexp → Com

Without the fix,

    #check fun (x : Ident) (a : Aexp) => imp {
      if (true) {
        skip;
      } else {
        ~(Com.asgn x a)
      }
    }

emits

    fun x a =>
      imp {
        if (true) {
          skip;
        } else {
          ~⋯
        }
      } : Ident → Aexp → Com

and with it it emits:

    fun x a =>
      imp {
        if (true) {
          skip;
        } else {
          ~(Com.asgn x a)
        }
      } : Ident → Aexp → Com

### DIFF
--- a/HL/Imp.lean
+++ b/HL/Imp.lean
@@ -752,7 +752,10 @@ partial def delabComInner : DelabM (TSyntax `imp_com) := do
       | .lit (.strVal s) =>
         let a ← withAppArg delabAexpInner
         `(imp_com| $(mkIdent (.mkSimple s)):ident := $a;)
-      | _ => `(imp_com| ~$(← delab))
+      | _ =>
+        let x ← withAppFn <| withAppArg delab
+        let a ← withAppArg delab
+        `(imp_com| ~($(mkIdent ``Com.asgn) $x $a))
     | Com.seq _ _ =>
       let s1 ← withAppFn <| withAppArg delabComInner
       let s2 ← withAppArg delabComInner


xhalo32:
@meluge already fixed this in 5be71b7e as part of https://github.com/plclub/sf-in-lean/pull/231

We should compare the approaches to see if one is better than the other.

xhalo32:
Now as #231 is merged, this PR has conflicts. I don't understand the delaborators well enough to resolve the conflict, but maybe @meluge could give their opinion?

bcpierce00:
#Note: Needs conflict resolution from @meluge

david-christiansen:
I think the big difference in the approaches is really one of how strict you want to be about metalanguage/object language separation.

The test case in #231 is this:
```
variable (x : Ident) (a : Aexp)
/-- info: imp {
  x := ~a
} : Com -/
#guard_msgs in
#check imp { x := ~a }
```

In this example, a Lean variable `x` that could stand for any identifier at all is rendered identically to the specific `Imp` identifier `x`. In a paper, I might use `$x$` for the former, and `\texttt{x}` for the latter.

I tend to think of Lean variables as being metavariables here, so I'd typically want to render them differently, but there can be value in not being overly pedantic about distinctions in some contexts.

When I've done things like this in the past, I would render that as something more like `~x := ~a`, with an antiquote feature in the syntactic category of identifiers in addition to the categories of expressions and commands.

berberman:
Subsumed into #337

PR #257, opened 2026-08-17
Reimplement autograding

xhalo32:
This PR replaces the old autograder with the new implementation which has support for definition and inductive holes.
It also adds a grading script `grading_check.py` which can be used in CI by running `make grading-check`.

## TODOs

- [ ] Documentation is outdated
- [ ] `comparator` could be moved to `plclub` from `xhalo32`
- [x] How to deal with the `List.IsNil.match_1` issue? I worked around this by changing `_ :: _` to `_`. See `Match Deduplication` in `AUTOMATED-GRADING.md`

These can probably be fixed by later PRs.


ionathanch:
We should probably start being more consistent with these doc names: either `AUTOMATED-GRADING`, `STYLE-CODE`, `STYLE-WRITING`, `ALPHA-TESTERS` or `AUTOMATED_GRADING`, `STYLE_CODE`, `STYLE_WRITING`, `ALPHA_TESTERS`

xhalo32:
@ionathanch good point, AUTOMATED_GRADING is the odd one out so I'll fix it!

xhalo32:
I'll split this PR into the infra and the changes to the materials.

scripts/grading_check.py
```diff
@@ -0,0 +1,209 @@
...
+#   checks all solutions in LF, doesn't run 'make', prints out statistics and the detailed json
+
+from subprocess import check_output, Popen, STDOUT
+from pathlib import Path
+from fractions import Fraction
+import os, sys, tempfile, shutil, json, argparse
+
+def runshell(cmd):
+    p = Popen(cmd, shell=True)
+    os.waitpid(p.pid, 0)
```
Robertboy18: One thing I’m a little worried about here: `runshell` waits for the command but never checks whether it succeeded. Could we use `subprocess.run(..., check=True)`? Otherwise a failed build might leave the script grading stale output.

scripts/grading_check.py
```diff
@@ -0,0 +1,209 @@
...
+    return total
+
+def assert_results(aggregate_results):
+    for vol in aggregate_results:
+        if "student" in aggregate_results[vol]:
+            for r in aggregate_results[vol]["student"]["theoremReports"]:
+                assert r["error"] is not None and next(iter(r["error"])) == "illegalAxiom"
+        if "solutions" in aggregate_results[vol]:
+            for r in aggregate_results[vol]["solutions"]["theoremReports"]:
+                assert r["error"] is None
```
Robertboy18, line 156: I may be missing a safeguard elsewhere, but would this pass if `theoremReports` were empty? Both loops would do nothing. Maybe we should require at least one report and check that the student and solution report names match.

xhalo32: Good point! A wider concern is that if there are no reports for an individual chapter, then it automatically passes. Not sure how we should check this though... One option once we know the max points for each chapter is to use the max points as a "checksum".

scripts/grading_check.py
xhalo32: Thanks for spotting this. I tested your suggestion and seems to work as expected, so let's go with that.

scripts/grading_check.py
xhalo32: Student and solution report names will always match since the names come from the grading variant which is the same between the tests, so I don't think it's a useful assertion to add.

Robertboy18: That makes sense about the report names. I still think checking that each selected volume and variant has at least one report would be useful, since otherwise losing all grading attributes would pass silently. The max-points checksum sounds like a good follow-up for detecting individual empty chapters.

xhalo32: Let's implement the check later since the autograder is anyways tested  manually for each week's assignment by of the TAs.

AUTOMATED-GRADING.md
```diff
@@ -0,0 +1,85 @@
+# Automated Grading
+
+TODO this is outdated!
```
mwhicks1, line 3: Is this still outdated? I notice in the TODO list at the top that documentation is mentioned as not done. Let's add an issue if we are not fixing this?

mwhicks1 (approved):

AUTOMATED-GRADING.md
xhalo32: https://github.com/plclub/sf-in-lean/issues/313

PR #262, opened 2026-08-17
Survey of :::details candidates

bcpierce00:
Adds `details-candidates.md`, a survey (no chapter files touched) of regions that could be wrapped in `:::details` directives — split out of `bcp/misc` for individual review.

Each entry is a proposal with inclusive line ranges; the intended workflow is to triage the list (delete unwanted entries) and then apply the survivors verbatim in a follow-up pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/details-candidates.md
+++ b/details-candidates.md
@@ -1,0 +1,351 @@
+# `:::details` candidates — survey pass
+
+*Survey only. No chapter file was modified by the pass that produced this list.*
+
+## How to use this file
+
+Every entry below is a **proposal** to wrap a region of a chapter in a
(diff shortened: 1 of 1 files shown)


bcpierce00:
#Note: The suggestions in the new .md file should actually be implemented before this is merged!

ionathanch:
Was this meant to be included within #269?

bcpierce00:
It’s ok for it to live in the repo for the moment, I guess. Might be good
to make an issue for somebody to push it to completion — all the info is in
that .md file file.

PR #263, opened 2026-08-17
Pass over UsingLean

ionathanch:
Current status: <del>about half done, up to "A New Step Towards Automation"</del> all done

### DIFF
--- a/LF/Basics.lean
+++ b/LF/Basics.lean
@@ -1392,6 +1392,7 @@ def ofNat : _root_.Nat → Nat
   | .succ n => .succ (ofNat n)
 
 instance (n : _root_.Nat) : OfNat Nat n := ⟨ofNat n⟩
+attribute [pp_nodot] Nat.succ
 ```
 :::
 
--- a/LF/UsingLean.lean
+++ b/LF/UsingLean.lean
@@ -197,7 +197,7 @@ that you need to prove it yourself.
 ::::
 
 :::terse
-Just because {tactic}`rw?` suggests a theorem does not mean that it will be useful; 
+Just because {tactic}`rw?` suggests a theorem does not mean that it will be useful;
 choose carefully from its suggestions (if at all).
 :::
 
@@ -242,12 +242,8 @@ don't want to have the tactics like {name}`Nat.add_comm` and
 {name}`Nat.add_assoc` "guess" which subterms to rewrite.
 
 The {tactic}`calc` tactic writes down the intermediate goals of a proof, and
-allows us to specify exactly which rewrite rules to apply at each step. It is a
-powerful tool for structuring proofs, and is often more readable than long
-{tactic}`rw` chains.
-
-{tactic}`calc` is designed to mimic the style of proofs in mathematics
-textbooks, which will often look something like this:
+allows us to specify exactly which rewrite rules to apply at each step. It is designed
+to mimic the style of proofs in mathematics textbooks, which will often look something like this:
 
 ```display
 n + (m + k)
--- a/LF/UsingLean.lean
+++ b/LF/UsingLean.lean
@@ -6,14 +6,18 @@ import LF.Induction
 open Verso.Genre Manual
 open SFLMeta
 
-set_option pp.fieldNotation false
-
-#doc (Manual) "UsingLean: Using the full power of a proof assistant" =>
+#doc (Manual) "UsingLean: Using the Full Power of a Proof Assistant" =>
 %%%
 tag := "UsingLean"
 htmlSplit := .never
 file := some "UsingLean"
 %%%
+
+:::ignore
+```lean -show
+variable (n : Nat)
+```
+:::
 
 :::instructors
 This chapter is the bridge to Lean's natural numbers,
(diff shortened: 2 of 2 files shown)


LF/Basics.lean
```diff
@@ -1391,6 +1391,7 @@ def ofNat : _root_.Nat → Nat
   | .succ n => .succ (ofNat n)
 
 instance (n : _root_.Nat) : OfNat Nat n := ⟨ofNat n⟩
+attribute [pp_nodot] Nat.succ
```
mwhicks1, line 1395: Is this attribute something we should explain?

LF/UsingLean.lean
```diff
@@ -98,160 +64,158 @@ This approach is useful in a textbook for understanding the structure of
...
 ```
 
-The annotation `: Nat` tells Lean that we are using its built-in `Nat` type.
+The annotation `: Nat` tells Lean that we are using its built-in {name}`Nat` type.
 
-In this chapter we will learn how to use the built-in `Nat` and some powerful
-features for computing with and proving properties about natural numbers.
-Specifically, we will learn about `dsimp`, `calc`, and `simp` annotations,
-which enable more powerful and concise proofs.
+In this chapter we will learn how to use the built-in {name}`Nat` and some
```
mwhicks1: This chapter is entitled `UsingLean`, not `UsingNat` (for example). I think we had some higher-level goals in mind for the chapter, and we are using `Nat` to illustrate them. Can we enumerate those goals?

LF/UsingLean.lean
```diff
@@ -98,160 +64,158 @@ This approach is useful in a textbook for understanding the structure of
...
 ::::
 
-```lean
-#guard_msgs(drop all) in
-example (n m q : Nat) :
-   (n + m) + q = m + (n + q) := by
+:::terse
+However, just because {tactic}`rw?` suggests a theorem to you does
+not automatically imply that it will be useful; you will need to
+carefully look through its suggestions to see which ones seem useful.
```
mwhicks1:
```suggestion
Just because {tactic}`rw?` suggests a theorem does not mean that it will be useful; 
choose carefully from its suggestions (if at all).
```

LF/UsingLean.lean
```diff
@@ -260,77 +224,78 @@ theorem mul_three (p : Nat) :
...
-tool for structuring proofs, and is often more readable than long `rw` chains.
+The {tactic}`calc` tactic writes down the intermediate goals of a proof, and
+allows us to specify exactly which rewrite rules to apply at each step. It is a
+powerful tool for structuring proofs, and is often more readable than long
+{tactic}`rw` chains.
 
-`calc` is designed to mimic the style of proofs in mathematics textbooks, which will
-often look something like this:
+{tactic}`calc` is designed to mimic the style of proofs in mathematics
+textbooks, which will often look something like this:
```
mwhicks1:
```suggestion
The {tactic}`calc` tactic writes down the intermediate goals of a proof, and
allows us to specify exactly which rewrite rules to apply at each step. It is designed 
to mimic the style of proofs in mathematics textbooks, which will often look something like this:
```

LF/UsingLean.lean
```diff
@@ -339,8 +304,8 @@ theorem succ_mul_succ (n m : Nat) :
   rw [Nat.add_mul, Nat.one_mul, Nat.mul_add, Nat.mul_one, ← Nat.add_assoc]
 ```
 
-Given this proof with `rw`, rewrite it with `calc`. Reminder that you can use `rw?` to find
-appropriate rules to rewrite by.
+Given this proof with {tactic}`rw`, rewrite it with {tactic}`calc`.
+Reminder that you can use {tactic}`rw?` to find appropriate rules to rewrite by.
```
mwhicks1:
```suggestion
Recall that you can use {tactic}`rw?` to find appropriate rules to rewrite by.
```

mwhicks1 (approved):
Looks great. Left some small suggestions, and one slightly larger request about specifying chapter goals.

LF/Basics.lean
mwhicks1: Moreover, I think we should also add this attribute for the `NatPlayground.Nat`s, which otherwise introduce the dot notation even though theorem statements do not, and we say nothing about this notation.

ionathanch: Yes, this attribute is added to `NatPlayground.Nat` here (not `Lean.Nat`), exactly because we don't explain the dot notation until UsingLean. The global `set_option` turns it off for everything else (you can't add the attribute to existing things like `Lean.Nat`), which is turned off in Basics and Induction, but then is left back on in UsingLean, and explained at around line 500.

ionathanch: (Also this whole chunk here is in a details directive with text telling the student not to worry about these details at the moment)

LF/UsingLean.lean
ionathanch: I've added an introductory paragraph at the beginning of the chapter in the latest commit 27e69c37a85666b9d40713b6a2e368b2640a88a3, does that look good?

mwhicks1:
Oops @ionathanch I hit merge but maybe you weren't ready! I guess there can always be another PR ...

mwhicks1:
Also, my recent change moved some text from `Induction` to `UsingLean` which should not conflict with your stuff, but will need to be merged, so we can wait for that to go in first.

ionathanch:
Oh I was just resolving some conflicts (with my own other PR...), it's good to go now so I'll merge

PR #270, opened 2026-08-18
Induction: Small fixes (plus question)

mwhicks1:
- One exercise had prose below the lemma that seemed to refer to it, so I moved it. But then I was confused by it so I left a comment. Help!'
- Cleared some comments we won't address (or will but don't need the comment)
- Added line break to improve student .lean formatting


LF/Induction.lean
```diff
@@ -610,8 +593,21 @@ theorem double_add (n : Nat) : double n = n + n := by
...
+with induction.
+
+(Tip: To expand the body of {name}`even` in a proof, use `rewrite [even]` or
+`rw [even]`.)
+
+:::dev "Mike Hicks @mwhicks1" NOW
+I don't understand why this lemma is helping us with induction. Might it be better
+to just offer the lemma as an exercise, and then refer back to it when you might actually
+use it (in an exercise), to illustrate?
+:::
```
dsainati1: We do use this lemma in induction proofs, but not in this file, apparently. I think it's fine to make this an exercise.

dsainati1 (approved):

LF/Induction.lean
mwhicks1: Oops, the exact same theorem is an exercise is in `UsingLean.lean`, and that's the theorem we refer to later in the book. So I'm going to delete the exercise here and prefer the one there.

PR #273, opened 2026-08-18
Style pass over Automation

ionathanch:


LF/Automation.lean
```diff
@@ -1129,10 +1101,10 @@ theorem in_re_match {α : Type} {s : List α} {re : RegExp α} {x : α}
...
     would not allow us to reason about the case `In x ∈ s₂`. -/
-    simp at hin
-    rcases hin with hin₁ | hin₂
-    · exact ih₁ hin₁
-    · exact ih₂ hin₂
+    simp only [List.mem_append] at hin
+    cases hin with
+    | inl hin₁ => exact ih₁ hin₁
+    | inr hin₂ => exact ih₂ hin₂
 ```
```
ionathanch: @dsainati1 What part of this should be `workinclass!`? The two chunks with the comments?

dsainati1 (approved):

LF/Automation.lean
dsainati1: Ah. I think I ported this file before we had `workinclass!`, maybe? That or I just didn't know about it. But in the original Rocq this function is just stubbed after the `induction`, so every case gets worked in class. We could just `workinclass!` the whole thing to replicate that, but I also think it could make sense to just `workinclass!` the interesting branches as well.

PR #274, opened 2026-08-18
UsingLean: spit and polish (and questions)

mwhicks1:
- Removed out-of-date :::dev (we do this exact proof in Basics)
- Trimmed redundant/irrelevant text/bits
- Highlight dsimp and rw difference
- Added two (urgent) questions about field notation and code actions, which don't gel with my experience using the extracted student files so far

### DIFF
--- a/LF/UsingLean.lean
+++ b/LF/UsingLean.lean
@@ -504,13 +504,6 @@ is recommended in idiomatic Lean developments.
 As an example, observe the difference in how Lean prints the goal in the following two examples:
 ::::
 
-:::dev "Mike Hicks (@mwhicks1)" NOW
-I'm _pretty_ sure that we have seen field notation before now, but not called attention to it.
-We did set disable field notation in some cases (`Nat.succ`) but not in general. Either fix this
-so field notation never comes up, or mention it when it first appears (in the Info View
-probably), by moving this text there.
-:::
-
 ::::terse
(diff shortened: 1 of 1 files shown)


LF/UsingLean.lean
```diff
@@ -510,6 +504,13 @@ is recommended in idiomatic Lean developments.
 As an example, observe the difference in how Lean prints the goal in the following two examples:
 ::::
 
+:::dev "Mike Hicks (@mwhicks1)" NOW
+I'm _pretty_ sure that we have seen field notation before now, but not called attention to it.
```
dsainati1: I believe Basics and Induction both have `set_option pp.fieldNotation false` as part of their header, which should entirely disable field notation in those files. So I do think this file is the first place where students will encounter it.

mwhicks1: When I generate the output files and look in `_out`'s version of `Basics.lean` for `student`, I don't see a `set_option` anywhere in the file. The only places I see it are in CustomTactics, Typeclasses, and UsingLean. I do see it on line 7 of the original verso, but that is not carried over to the student file it seems?

dsainati1: Oh interesting. We should definitely fix that!

dsainati1: https://github.com/plclub/sf-in-lean/pull/292

dsainati1 (approved):

PR #275, opened 2026-08-19
Lists: Questions and some adjustments

mwhicks1:
- Relocated the introduction of structures (as `NatProd'`) to come after `NatProd` is finished. Added section header.
- Put notation stuff in `:::details`
- Clarified explanation about simplification lemmas
- Two questions: namespace for `NatProd`? And: properly introduce implicit arguments `{ ... }` ?

Didn't get that far, so still a draft.

### DIFF
--- a/LF/Lists.lean
+++ b/LF/Lists.lean
@@ -66,6 +66,14 @@ to two arguments of type {name}`Nat`."
 :::slidebreak
 :::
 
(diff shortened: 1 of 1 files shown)


ionathanch:
I think this was included in #280 so it can be closed right?

mwhicks1:
Oh that's funny. I didn't mean to do that. Yep, we can close it!

PR #276, opened 2026-08-19
Polish Smallstep

KihongHeo:
- Update variable names to follow the naming convention
- Simple cleanup
- Add a question about `absurd`

### DIFF
--- a/TS/Smallstep.lean
+++ b/TS/Smallstep.lean
@@ -677,6 +677,11 @@ theorem nf_same_as_value (t : Tm) : IsNormalForm Step t ↔ IsValue t :=
 :::dev "Kihong Heo (KihongHeo)"
 Tactic `absurd` is first introduced here. Do we want to explain it?
 :::
+:::dev "Daniel Sainati (dsainati1)"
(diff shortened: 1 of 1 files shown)


TS/Smallstep.lean
```diff
@@ -674,6 +674,9 @@ theorem nf_is_value (t : Tm) (h : IsNormalForm Step t) : IsValue t := by
 theorem nf_same_as_value (t : Tm) : IsNormalForm Step t ↔ IsValue t :=
   ⟨nf_is_value t, value_is_nf t⟩
 ```
+:::dev "Kihong Heo (KihongHeo)"
+Tactic `absurd` is first introduced here. Do we want to explain it?
```
dsainati1, line 678: I think some of these proofs were originally Claude-generated, so we probably want to redo them from scratch, in which case introducing `absurd` is likely not necessary.

KihongHeo: Thanks. Added your comments to the file.

dsainati1 (approved):

PR #286, opened 2026-08-20
Print the substitutions with braces instead

meluge:
In `hoare_asgn_example1'`, after the first bulletpoint, the goal shows as `⊢ {{True}} ->> (X = 1) [X ↦ 1]`. This is due to delabSub currently printing all substitutions without the brackets (which works fine if we're delaborating inside of an existing bracket. To make this work generally (e.g `->>` or `<<->>`), I've made `delabSub` to print with brackets by default and rather remove the brackets when we have to print substitutions inside preexisting brackets. This fixes the same example to print `⊢ {{True}} ->> {{(X = 1) [X ↦ 1]}}`.

### DIFF
--- a/HL/Hoare.lean
+++ b/HL/Hoare.lean
@@ -758,6 +758,13 @@ details.
 namespace Assertion.Delab
 open Lean PrettyPrinter Delaborator SubExpr Imp.Delab
 
+/-- Strip the braces from a printed `{{ … }}` assertion, for
+a printer that is itself inside a pair of braces and will
+supply its own. -/
+def unbrace : Term → Term
+  | `({{ $P }}) => P
+  | t => t
+
 /-- Rebuild the surface form of an assertion body, undoing the state
 threading the `assn` elaborator performs: `st[X]` prints as `X`,
 `Aexp.eval st a` as `a`, `Bexp.eval st b = true` as `b`, an applied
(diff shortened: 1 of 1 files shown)


Robertboy18 (commented):
This also needs updating to `main`, where the substitution printer is now an unexpander and `delabAexpInner` no longer exists.

HL/Hoare.lean
```diff
@@ -1648,21 +1656,25 @@ P [ X ↦ a ]
...
-picks it up inside an assertion body, and the enclosing printer supplies
-the single pair of braces. -/
+/-- Print an `Assertion.subst` back in `{{P [x ↦ a]}}` notation. -/
 @[delab app.Assertion.subst]
 def delabSub : Delab := whenPPOption getPPNotation do
-  guard <| (← getExpr).isAppOfArity ``Assertion.subst 3
+  let e ← getExpr
+  if e.isAppOfArity ``Assertion.subst 4 then
+    let P ← withAppFn delab
+    return ← `(($P) $(← withAppArg delab))
```
Robertboy18, line 1665: This breaks the applied case when the assertion is a function call like `F n`, with `F : Nat → Assertion`. It now prints `({{(F n) [X ↦ 1]}}) st`, which Lean rejects because `F n` has type `Assertion` where the notation expects `Prop`. Could we keep the old fallback here and add a test?

PR #287, opened 2026-08-20
Implement optional `with` tactics for inversion

ionathanch:
[#lean-software-foundations-contributors > inversion tactic @ 💬](https://leanprover.zulipchat.com/#narrow/channel/607217-lean-software-foundations-contributors/topic/inversion.20tactic/near/617735105)

This implements an optional sequence of tactics that goes after `with` but before inversion alternatives:

```lean
example {α x y} {l : List α} (h : NoStutter l) (hl : l = x :: y :: []) : x ≠ y := by
  inversion h with
    injections
    try subst_vars
  | nostutter2 => assumption
```

The optional tactics apply to all goals, just like `all_goals`.

Technical details: `inductionCasesOn` produces a list of `CasesSubgoals`, which previously was passed all around, since it contains the goal mvar and also the name of the constructor associated with that case. Now, the goal is assigned the constructor name as its user-facing name, so that we only have to pass around the mvars, and if the optional tactic produces additional subsubgoals, they will be named correctly:

```lean
example {x y} {l : List Nat} (h : NoStutter l) (hl : l = x :: y :: []) : x ≠ y := by
  inversion h with
    injections
    try cases x
  | nostutter2.zero => subst_vars; assumption
  | nostutter2.succ => subst_vars; assumption
```

### DIFF
--- a/LF/CustomTactics.lean
+++ b/LF/CustomTactics.lean
@@ -1,16 +1,8 @@
 module
 
 public meta import Lean.Elab.ConfigEval
-public meta import Lean.Elab.Tactic.ElabTerm
 public meta import Lean.Elab.Tactic.RenameInaccessibles
 public meta import Lean.Elab.Tactic.Induction
-public meta import Lean.Elab.Tactic.BuiltinTactic
-meta import all Lean.Elab.Tactic.BuiltinTactic
-
-public meta import Lean.Meta.Tactic.Generalize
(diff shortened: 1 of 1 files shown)


xhalo32 (approved):
Looks good to me. Didn't test locally.

dsainati1 (approved):

PR #288, opened 2026-08-20
UsingLean explains irreducible; code actions to Induction.

rogerburtonpatel:
Simple fixes to UsingLean, largely pedagogical. Better examples and more explanation. Code actions moved to Induction.


LF/UsingLean.lean
```diff
@@ -86,6 +90,11 @@ features, writing `Nat.<theorem>` to reference Lean's version
 of `<theorem>`. (By convention, theorems about a type live in the namespace of
 that type, hence the need for the `Nat.` prefix.)
 
+Definitions in {name}`Nat` are not marked `@irreducible`. This lets us use
```
berberman:
```suggestion
Definitions in built-in {name}`Nat` are not marked `@[irreducible]`. This lets us use
```

LF/UsingLean.lean
```diff
@@ -320,15 +329,13 @@ If you prefer {tactic}`rw` to {tactic}`calc`, that's fine! Each has particular
...
 
 # Definitional Simplification: {tactic}`dsimp`
 
-Often, rather than rewriting by a known equation like
-`n + succ m = succ (n + m)` using `rw [add_succ]`,
-we just want to simplify the function (here {name}`Nat.add`) automatically when we can.
+Often, rather than repeatedly rewriting by a known equation like
+`rw [Nat.mul_zero, Nat.mul_zero]` to solve a goal like
+`n * (m * 0) = 0`,
+we just want to simplify the function (here {name}`Nat.mul`) automatically when we can.
```
berberman, line 328: I think there are two use cases of `dsimp` we might want to distinguish:
1. We can feed a definition to `dsimp`, e.g. `dsimp [Nat.mul]`, which would unfold `Nat.mul` in the expression and perform reductions.
2. We can also feed a theorem, e.g. `dsimp [Nat.mul_zero]`, which would rewrite `Nat.mul_zero` many times and then perform reductions, just like `simp [Nat.mul_zero]`.

And `@[defeq]` lemmas in simp set are always used implicitly, which we shouldn't talk about here.

So the first use applies when there's no characterization lemma defined (as we have to unfold the definition and let it compute), or we're getting tired of writing many `rw`s.

However, should the second use be encouraged? Suppose that we don't want to `rw` manually and we do have characterization lemmas defined, should `dsimp [Nat.mul_zero]` be preferred over `dsimp [Nat.mul]`?

An example:

```lean
example (n : Nat) : n * (n * (n * 0)) = 0 := by
  rw [Nat.mul_zero, Nat.mul_zero, Nat.mul_zero]

example (n : Nat) : n * (n * (n * 0)) = 0 := by
  dsimp [Nat.mul]

example (n : Nat) : n * (n * (n * 0)) = 0 := by
  dsimp [Nat.mul_zero]
```

berberman (approved):
Thanks!

LF/Induction.lean
```diff
@@ -970,7 +964,92 @@ theorem mul_one (p : Nat) :
 :::
 ::::
 
+## Aside: Using Code Actions to Generate Match Skeletons
+
+Lean's language server can suggest _code actions_, which are
+small editor commands that modify the source code.
```
mwhicks1, line 972: Maybe just make these first two lines visible in terse mode, and the rest of this paragraph is `full` only? In person the lecturer will just do these things, not read them.

LF/Induction.lean
```diff
@@ -970,7 +964,92 @@ theorem mul_one (p : Nat) :
...
+  induction n with
+  | zero => exact (beq_self zero)
+  | succ n ih => rw [Nat.beq, ih]
+```
+
+Note that Lean used `_` for the induction hypothesis in the generated `succ` branch.
+At that point, Lean didn't know whether the unfinished proof would need to refer to the hypothesis.
+Since we use it in {tactic}`rw`, we replace `_` with the name `ih`.
+
+In later chapters, we will see some tactics that can make such
```
mwhicks1: Can we make this reference explicit, i.e., to the chapter where this happens?

LF/UsingLean.lean
mwhicks1: I like this example. I am going to add it. Thanks for spelling this out!

bcpierce00:
@mwhicks1 ... By "still needs attention", I meant that (1) there are some unresolved comments in this conversation and (2) I added a few :::dev comments in my last pass that should be looked at before the chapter goes live.

LF/UsingLean.lean
mwhicks1: I thought this example would be more persuasive if what we care about is multiple rewritings:
```
example (n : Nat) : (((0 * n) * n) * n) = 0 := by
  dsimp [Nat.zero_mul]
```
I thought this would be cool because you can't do `dsimp [Nat.mul]` -- reduction won't help you make progress. But when I try this with `dsimp` it doesn't work! That's because (AFAICS) `dsimp` only rewrites multiple theorems when they are equal according to definitional equality, not by (say) induction. This seems tricky to explain. It is easer to talk about `dsimp` just as a way to unfold definitions and reduce them repeatedly.

I'll add your comment to the text for later re-consideration.

PR #291, opened 2026-08-21
Lists/Poly: Revert, then introduce, implicit args in theorems

mwhicks1:
The Lists chapter used implicit arguments extensively, but did not introduce their use. For simplicity of exposition I reverted implicit arguments to explicit ones in List. Then I added a little text in Poly when they are used in theorems to explain why this is done (taking the rationale from STYLE-CODE.md).

Claude did the search/replace to explicit parameters in Lists, which I verified.

### DIFF
--- a/LF/Poly.lean
+++ b/LF/Poly.lean
@@ -184,7 +184,7 @@ What is the full type of {name}`MyList.nil`? We can read off the
 result type {lean}`MyList α` from the definition,
 but to state the full type we must also bind `α`.
 Since the type argument to the constructor is implicit,
-Lean writes its type as {lean}`{α : Type} → MyList α`.
+Lean writes its type as (the equivalent of) {lean}`{α : Type} → MyList α`.
 ::::
 
 ```lean (name := nil)
@@ -236,7 +236,8 @@ def myRepeat (α : Type) (x : α) (count : Nat) : MyList α :=
 Some simple facts about {name}`myRepeat`:
 
 ```lean
-theorem myRepeat_zero (α : Type) (v : α) : myRepeat α v 0 = MyList.nil := rfl
+theorem myRepeat_zero (α : Type) (v : α) :
+    myRepeat α v 0 = MyList.nil := rfl
 
 theorem myRepeat_succ (α : Type) (v : α) (count : Nat) :
     myRepeat α v (count + 1) = MyList.cons v (myRepeat α v count) := rfl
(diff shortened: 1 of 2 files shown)


LF/Poly.lean
```diff
@@ -1062,6 +1064,32 @@ theorem filter_cons_of_neg {α : Type} {test : α → Bool} {head : α}
    dsimp
 ```
 
+::::full
+You might have noticed that {name}`filter_cons_of_pos` and {name}`filter_cons_of_neg`
+have implicit parameters, such as `head` and `tail`, that do not have type `Type` like `α` does.
+This is a standard Lean convention for lemmas that are likely to be used by {tactic}`rw`
+or {tactic}`dsimp` when their values can be inferred by unification.
```
berberman, line 1077: IIRC we are avoiding the word "unification" as it sounds too technical cc @dsainati1

LF/Poly.lean
```diff
@@ -1062,6 +1064,32 @@ theorem filter_cons_of_neg {α : Type} {test : α → Bool} {head : α}
    dsimp
 ```
 
+::::full
+You might have noticed that {name}`filter_cons_of_pos` and {name}`filter_cons_of_neg`
+have implicit parameters, such as `head` and `tail`, that do not have type `Type` like `α` does.
```
berberman: Can we reword this sentence? Not having type of `Type` is not a cause or reason that they are implicit.

mwhicks1: Yes, but it's worth noting: It doesn't have to be of type `Type` and that's all they've seen until now. We are observing that the situation is different here, and then explaining why it's this way. Happy to consider rewording, but I want to call this out.

LF/Poly.lean
mwhicks1: Earlier in the file the word _unify_ is used to explain what's going on here. In general I'm not sure why we are Ok with terms like "implicit parameters", "parametric polymorphism", "type synthesis", but not "unification" ?

LF/Poly.lean
mwhicks1: I updated the text a little bit to hopefully help.

LF/Poly.lean
dsainati1: Oh if we actually explain unification then it's fine to use, I didn't realize this was a term we had explained already.

dsainati1 (approved):

PR #292, opened 2026-08-21
Move pp.fieldNotation out of header in Basics and Induction

dsainati1:
These were getting dropped by Verso extraction. This fixes that.


berberman (approved):

ionathanch:
Wait I think these should be in a `-show` block, it's weird that this block of code is the first thing students see in the textbook

image

dsainati1:
https://github.com/plclub/sf-in-lean/pull/294

PR #293, opened 2026-08-21
Improve elaboration performance

david-christiansen:
These changes improve elaboration performance by updating the dependencies on Verso/SubVerso. There's also a change from one highlighting API to another that is more efficient.

Before these changes, `IndProp` took 15.3s to elaborate on my machine, now it takes 10.3s. Once 4.34.0 is out, precompilation in Verso should bring this down to around 7s.

The improvements upstream are:

* Batched highlighting re-uses cached information between commands, rather than reconstructing it for each command. This is now done more consistently in Verso, and I enabled it in SFLMeta as well.

* SubVerso itself now caches more information in its own data structures, instead of re-using the compiler's, which are optimized for a very different access pattern.

Rather than backport fixes to Verso for 4.33.0, I instead made a minor change in Verso to enable it to be compatible with both versions. This means that Lake will try to edit your toolchain file when you run `lake update`. Please use `lake update --keep-toolchain` instead, until you're ready to upgrade to a new release.


bcpierce00 (approved):

lakefile.toml
```diff
@@ -46,6 +54,7 @@ name = "Credits"
 
 [[lean_lib]]
 name = "SFLMeta"
+precompileModules = true
```
ionathanch: Is this necessary for the performance improvements? This seems to be causing build errors both in the CI and locally on my machine

david-christiansen: Interesting! This is showing a difference in shared library loading between Linux and macOS.

Precompilation is not needed for the entire improvement, but it's about half of the speedup. Mostly because it by side effect causes Verso itself to be precompiled, which we plan to enable very soon by default anyway. So flushing out issues with that now would be good to do. But I think the issue here is that SFLMeta imports parts of Lake, and those aren't available at link time.

If you'd like, feel free to merge without this step, and then I can send you a bump PR once https://github.com/leanprover/verso/pull/971 is merged. Otherwise, I'll take a peek at getting the config set up for the Lake shared library too.

david-christiansen: I can solve the problem now, but only in a Lean-format Lake config file. Would you prefer that I do that, or that we just wait for leanprover/verso#971 with a follow-up PR in a few days? I know you're on a deadline, so let me know what's most useful.

Vtec234: We merged a fix for the error in CI recently - moving to Lean v4.34.0-rc2 should fix it.

david-christiansen: Let's skip the workaround in the Lake config, that's fragile compared to waiting for a release.

david-christiansen: I just updated the PR to remove the precompilation setup. It should be in Verso by the time 4.34.0 comes out, so those performance improvements will happen automatically at that update.

PR #296, opened 2026-08-21
Style pass over Poly

ionathanch:


LF/Poly.lean
```diff
@@ -1402,7 +1406,7 @@ distributed programming framework.
 def fold {α : Type} {β : Type} (f : α → β → β) (l : List α) (b : β) : β :=
   match l with
   | [] => b
-  | head :: tail => f head (fold f tail b)
+  | a :: l => f a (fold f l b)
```
berberman, line 1409: IIRC the list pattern would be `x :: xs`, but I guess we don't need to be very strict on following this naming convention.

berberman (approved):

LF/Poly.lean
ionathanch: Our style guide says that lists are `l`, which is consistent with everywhere else. ([Mathlib's style guide](https://leanprover-community.github.io/contribute/style.html) says `s`, `t`, which I think is kind of strange.) I wasn't sure about the head of the list: I think it's nice that we have `a : α` and `b : β`, but our style guide (and Mathlib) says `a`, `b`, `c` for propositions, so it should be `x : α` and `y : β`, but I feel this makes it harder to mentally keep track of which one is the list element and which one is the accumulator...

berberman: Never mind, I guess my impression was from [lean4 style](https://github.com/leanprover/lean4/blob/master/doc/std/style.md), but it does contain `a :: l`.

PR #306, opened 2026-08-24
Factor custom commands to SFLCompat library

berberman:
This PR introduces `SFLCompat` as a new library, which includes `sf_experiment` and `sf_expect_failure` in the old `SFLMeta/SFLCompat.lean` file, as well as the two new commands coming from #299 (renamed to `sf_recall` and `sf_recall_source`.) It also updates project extraction logic to let ` ```recall ` and ` ```recallSource ` verso code blocks extract to  `sf_recall` and  `sf_recall_source` commands respectively. More over it adds module system support to `sf_recall` command elaboration. Noting that previously the `SFLMeta/SFLCompat.lean` file would be copied to `vol/SFLCompat.lean` during extraction, but now entire `SFLCompat` library will get copied to the top-level of the extracted project.

### DIFF
--- a/LF/Typeclasses.lean
+++ b/LF/Typeclasses.lean
@@ -1,18 +1,4 @@
-import VersoManual
-import VersoManual.InlineLean
-import Illuminate
-import SFLMeta.Bnf
-import SFLMeta.Ignore
-import SFLMeta.Save
-import SFLMeta.Comment
-import SFLMeta.Exercise
-import SFLMeta.Grade
-import SFLMeta.Hide
-import SFLMeta.Instructors
-import SFLMeta.SlideBreak
-import SFLMeta.Solution
-import SFLMeta.Terse
-import SFLMeta.RecallBlock
+import SFLMeta
 
 set_option autoImplicit false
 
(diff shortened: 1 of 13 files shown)


LF/Typeclasses.lean
berberman, line 347: Are we using `recallSource` here because

````lean
```recall
class BEq (α : Type) where
  beq : α → α → Bool
```
````

would not work due to the universe parameter? cc @chenson2018

dsainati1 (approved):

LF/Typeclasses.lean
chenson2018: Yes, I used `recallSource` for this reason. (Some may work, I just used it across the board for this file once I hit this).

ionathanch:
How does ` ```recall +error ` ever arise? Wouldn't it have to had errored in the original declaration, which means that there wouldn't be a declaration registered in the context to compare against anyway?

berberman:
I think this is for expected failures in recall checks. For example, we might recall a definition using an alternative definition that people would expect to be equivalent, but it turns out they're not. In this case, we can see the error messages from the mismatch.

LF/Typeclasses.lean
david-christiansen: It would be possible to implement an "almost equivalent" version that e.g. ignores universes

PR #309, opened 2026-08-24
Some last polishing (and a few comments!) on Basics

bcpierce00:


dsainati1:
Is it intentional that this PR includes commits from adding the `optional` flag to exercises? The changes in those commits already appear to be in the main branch as well.

bcpierce00 (commented):
> Is it intentional that this PR includes commits from adding the `optional` flag to exercises? The changes in those commits already appear to be in the main branch as well.  

No, not intentional!  Will see if I can get rid of those...

bcpierce00:
OK, this looks better...

LF/Basics.lean
```diff
@@ -2110,17 +2110,23 @@ theorem succ_beq_succ (n m : Nat) : ((succ n) == (succ m)) = (n == m) := by rfl
...
+::::full
+Aside: Our naming convention
+for simplification rules encodes their meaning.
+For `add_zero` and `add_succ`, notice that the `zero` and
+`succ` come after the `add`; this is because they depend on `add`'s _second_ argument
 and do not care about its first.
-In the `beq` rules above, we write `zero_beq_zero` and `zero_beq_succ`
+:::dev "Benjamin Pierce (bcpierce00)"
+So they would be called `zero_add` and `succ_add` if they depended on the first argument?? This explanation isn't making complete sense to me...
+:::
```
mwhicks1, line 2121: Yes, as far as I understand it.

LF/Basics.lean
```diff
@@ -2110,17 +2110,23 @@ theorem succ_beq_succ (n m : Nat) : ((succ n) == (succ m)) = (n == m) := by rfl
...
+Also, in the `beq` rules above, we write `zero_beq_zero` and `zero_beq_succ`
 because the rules apply to both the first and second arguments of `beq`. We put
-`beq` in between the arguments because it usually written in infix.
-There are not strict style conventions for naming theorems like this in Lean, but many
+`beq` between the arguments because it usually written in infix.
+There are no strict style conventions for naming theorems like this in Lean, but many developers
 follow this approach.
+::::
+:::dev "Benjamin Pierce (bcpierce00)"
+We haven't really articulated an "approach" -- just given a couple of miscellaneous examples...
```
mwhicks1, line 2129: Do you think that it's better to leave the names as mysterious rather than give a few examples? I figured that giving some hint about naming would be useful.

mwhicks1 (approved):

LF/Basics.lean
bcpierce00: Not sure the explanation is really "actionable", especially at this point.  

Maybe move it to UsingLean, and try to broaden it a bit?

mwhicks1: I'll make a note to do that and just merge. We can get it on the next go-round.

PR #310, opened 2026-08-24
Documenting recall blocks

ionathanch:


berberman:
Could you rebase on #306 or directly commit to that branch? I've made some docs there, feel free to change them.

PR #311, opened 2026-08-25
Poly: make `unzip` a manually graded exercise about char lemmas

xhalo32:
Also adds an introduction to extensionality of pairs


dsainati1 (approved):

dsainati1:
@xhalo32 seems like this has conflicts with your PR that added the autograding support

PR #316, opened 2026-08-25
Polishing on Preface and Basics

bcpierce00:


berberman (commented):
FYI `-show` removes it from the HTML book, and `:::ignore` stops it being extracted to student projects. So it's only visible in verso source file, and we don't have to explain it or wrap it in `:::details`.

LF/Basics.lean
```diff
@@ -290,6 +319,9 @@ inductive MyBool : Type where
   | false
 ```
 
+:::dev "Benjamin Pierce (bcpierce00)"
+Maybe this next bit belongs in a :::details block?  The explanation is a bit lacking...
+:::
```
berberman, line 324: Oops, I meant to comment on this

SFLCompat/Experiment.lean
```diff
@@ -145,18 +145,18 @@ Diagnostics from the expected failure are suppressed.
...
-  "sf_expect_failure"
+  "expect_failure_in"
 
 /-
-  Like `sf_expect_failure` but reports the diagnostics.
+  Like `expect_failure_in` but reports the diagnostics.
 -/
 def expectFailureInfoTk := leading_parser
-  "sf_expect_failure?"
+  "expect_failure_in?"
```
berberman: My thought was that all commands in SFLCompat should start with `sf_` prefix to distinguish from Lean commands. I'm ok with changing them, but if we decided to rename this one, we should also rename `sf_experiment` to something else and `sf_recall` and `sf_recall_source` back to `recall` and `recall_source` for consistency.

mwhicks1: I'll change to `sf_expect_failure_in`.

mwhicks1 (approved):

PR #319, opened 2026-08-26
`nth?` characterizing lemmas and `list_ext` exercise

xhalo32:
I also cleaned up some instances of unfolding definitions with `dsimp` in favor of using the characterizing lemmas.

This has a weak dependency on https://github.com/plclub/sf-in-lean/pull/311 since the solutions use `Prod.ext`.

Also we haven't explained `Prod.ext_iff` i.e. applying extensionality in hypotheses.


LF/Tactics.lean
```diff
@@ -1902,10 +1932,11 @@ theorem unzip_zip {α β : Type}
     | nil => contradiction
     | cons y ys =>
       rw [zip_cons_cons]
-      dsimp [unzip]
-      rewrite [ih]
-      · rfl
-      · injections
+      injections h -- uses `.add` which we have to rewrite to `+` to use `Nat.add_zero`
```
mwhicks1: Have students seen `injections h` to this point?

LF/Tactics.lean
```diff
@@ -1902,10 +1932,11 @@ theorem unzip_zip {α β : Type}
     | nil => contradiction
     | cons y ys =>
       rw [zip_cons_cons]
-      dsimp [unzip]
```
mwhicks1: What is the principle for when to use `dsimp` vs. when to use rewriting? The proof is obviously much simpler when using `dsimp` here.

LF/Tactics.lean
```diff
@@ -1920,13 +1951,21 @@ theorem unzip_zip' {α β : Type}
     rfl
   | cons x xs ih =>
     let ⟨a, b⟩ := x
-    dsimp [unzip] at h
-    injections h₁ h₂
+    rw [Prod.ext_iff] at h
+    obtain ⟨h₁, h₂⟩ := h
+    rw [unzip_cons_fst] at h₁
+    rw [unzip_cons_snd] at h₂
+    dsimp at h₁ h₂
```
mwhicks1: Have we seen use of `dsimp` with no argument? Does this mean to definitionally simplify any definitions in the target?

LF/Tactics.lean
```diff
@@ -1027,27 +1027,26 @@ example (m : Nat) (h : ∀ n, m * n = 0) : m = 0 := by
...
 ```
 
-:::::exercise (rating := 3) (name := "nth?_always_none")
+:::::exercise (rating := 2) (name := "nth?_always_none")
 Use {tactic}`have`, {tactic}`replace`, or {tactic}`specialize` to prove the the following lemma,
 following the model of the examples above. Do not use {tactic}`induction`.
 
 ```lean
-theorem nth?_always_none (l : List Nat) (h : ∀ i, nth? l i = none) :
+theorem nth?_always_none {l : List α} (h : ∀ i, nth? l i = none) :
```
dsainati1, line 1039: At this point in the book we have elected to prefer standard library definitions where possible. Can we rewrite this exercise to use `List.get?`

LF/Tactics.lean
dsainati1: This is taught earlier in this chapter.

LF/Tactics.lean
xhalo32: A lot of the exercises in this file use our custom definitions. Do you want me to rewrite all of them?

dsainati1: Hmm. I can do a PR for it once this is merged, if it's a big change

LF/Tactics.lean
xhalo32: But we still haven't explained the problem with why we need `Nat.add_eq`. I have also left a dev note about a similar issue in `append_left_cancel` exercise.

LF/Tactics.lean
xhalo32: It uses the definition rather than characterizing lemmas. I call it "defeq overuse". When using `dsimp` it's better to use the characterizing lemmas, like `dsimp [unzip_nil, unzip_cons...]` which we have discussed [on zulip](https://leanprover.zulipchat.com/#narrow/channel/607217-lean-software-foundations-contributors/topic/Poly.3A.20unfolding.20with.20dsimp/with/616626886). Most of the lemmas can also later be marked as `@[simp]`.

We can reopen the conversation about this on zulip if you want others's opinions.

LF/Tactics.lean
xhalo32: This is in my opinion the primary way to use dsimp: to simplify projections and reductions. I don't remember seeing if we have explained it anywhere.

xhalo32: CC: @dsainati1

LF/Tactics.lean
mwhicks1: If that's the case, then we probably are not explaining things properly in `UsingLean.lean`. As of now, that file makes the best cause for `dsimp` being applied to _definitions_ not to theorems. For the latter, it seems more persuasive (at that moment in the course) to use `rw` for theorems. Just commented in the Zulip discussion that we need clearer guidance about when `dsimp` is appropriate.

LF/Tactics.lean
mwhicks1: > This is in my opinion the primary way to use dsimp: to simplify projections and reductions. 

This would be consistent with what we've presented in Basics and Induction, I think. Basically: Use `rw` with characterizing lemmas _or even definitions_ but then use `dsimp` to perform reductions. That's a nice 1-2 punch. We could suggest that `rw [h]; dsimp` could be replaced by `dsimp [h]` in some cases (but not all). Does that seem right?

xhalo32: Yes this sound about right. Note that the char lemma `h` needs to be an `rfl` proof to work, not `by rfl`.

LF/Tactics.lean
xhalo32: If we use the standard library definitions, the students will have access to `List.unzip_zip` for example which makes the `unzip_zip` exercise trivial. Do we want to forbid them from using the solution directly from the standard library by making it manually graded?

dsainati1: ahh okay that is a fair point. Probably in that case we need to leave it as is.

mwhicks1 (approved):

PR #320, opened 2026-08-26
Quick pass over Typeclasses

ionathanch:
Aside from style edits the major additions are:
* Two monoid and group exercises
* Formatting the Bool vs. Prop table


LF/Typeclasses.lean
```diff
@@ -14,6 +14,15 @@ file := "Typeclasses"
 tag := "Typeclasses"
 %%%
 
+:::ignore
```
dsainati1, line 17: As these show up in the generated `.lean` file for students, I think we want to adopt the convention of putting these in `:::details` blocks.

dsainati1 (approved):

LF/Typeclasses.lean
berberman: I think everything inside `:::ignore` is not extracted, see https://github.com/berberman/sf-in-lean-previews/blob/gh-pages/main/lf/student/lean/LF/Typeclasses.lean.

Please don't put them in `:::details`, beause these variables are only for verso presentation and should not affect student projects.

xhalo32:
I just noticed that all of the `List.elem_poly` definitions are incorrectly named, they should use camelCase instead.

ionathanch:
ok the definition names should be fixed now

LF/Typeclasses.lean
```diff
@@ -214,15 +228,15 @@ instance instDefaultValueNat : DefaultValue Nat where
...
+explicit {lean}`defaultValue` parameter with an instance implicit:
 
 ```lean
-def List.headOr {α : Type} [DefaultValue α] (xs : List α) : α :=
-  match xs with
+def List.headOr {α : Type} [DefaultValue α] (l : List α) : α :=
+  match l with
   | [] => DefaultValue.value
-  | hd :: _ => hd
+  | x :: _ => x
```
mwhicks1, line 227: I'm surprised that you changed `xs` to `l` but then changed the pattern variable to `x` -- if you'd left `xs` as is that would make more sense to me. I imagine you are following a naming convention in [STYLE-CODE](https://github.com/plclub/sf-in-lean/blob/main/STYLE-CODE.md)? I looked there and didn't see clear guidance about generic pattern variables. That said, that file is not even self-consistent -- it mentions `a`, `b`, `c` for propositions and `m`, `n`, `k` for natural numbers and then has an example that uses `a` and `b` for natural numbers!

LF/Typeclasses.lean
```diff
@@ -334,34 +347,38 @@ the kind of typeclass we just learned to define:
...
 natural numbers.
 
 ```lean
-theorem List.elem_poly_eq_elem_nat (xs : List Nat) (n : Nat) : xs.elem_poly n = xs.elem_nat n := by
+theorem List.elemPoly_eq_elemNat (xs : List Nat) (n : Nat) : xs.elemPoly n = xs.elemNat n := by
   solution!(
   induction xs with
   | nil =>
-    rewrite [List.elem_poly_nil, List.elem_nat_nil]
+    rewrite [List.elemPoly_nil, List.elem_nat_nil]
```
mwhicks1, line 363: The [STYLE-CODE](https://github.com/plclub/sf-in-lean/blob/main/STYLE-CODE.md) file suggests that theorem names should be all in snake case. I kind of see why you have a mixed situation here but I wonder if there are examples in Mathlib etc. that do the same thing? We should be clear in STYLE-CODE.

ionathanch: There are examples in the standard library, e.g. https://github.com/leanprover/lean4/blob/master/src/Init/Data/List/Basic.lean#L968 with `takeWhile_nil`, `dropWhile_nil`, `dropLast_singleton`, `isPrefixOf_cons_cons`, etc. I can add to STYLE-CODE explicitly stating that theorems about definitions in camel case keep the camel casing just for the definition name part

mwhicks1: > I can add to STYLE-CODE explicitly stating that theorems about definitions in camel case keep the camel casing

That would be great.

ionathanch: 👉 #342

ionathanch:
With the style update in #342 I've used `ms : List Nat` and `ys : List α` where it would make sense to

mwhicks1 (approved):

PR #321, opened 2026-08-26
Remove discussion of `dsimp` from LF

dsainati1:


LF/UsingLean.lean
```diff
@@ -324,71 +328,110 @@ uses, and both will be tools in your ever-growing toolbox of tactics.
...
+in the target expression
+  addThrice n = n + addTwice n
+
+n✝ n : Nat
+⊢ addThrice n = n + addTwice n
+```
+
+But again we encounter an error! The expression in which we are trying to rewrite
+{name}`Nat.add_assoc` isn't of the form `n + m + k`, so we can't proceed. What then, should we do?
+To proceed here, we need to reveal to Lean the underlying structure of `addThrice` and `addTwice`,
```
xhalo32: Would "we need to reveal the underlying definitions for `rw`, which operates only on syntax, to see the addition" be better?

LF/UsingLean.lean
```diff
@@ -324,71 +328,110 @@ uses, and both will be tools in your ever-growing toolbox of tactics.
...
+    rw [h, addThrice, Nat.add_assoc]
 ```
 
-:::gradeTheorem 2 dsimp1
+:::gradeTheorem 2 rwUnfold
 :::
 :::::
 
-`dsimp at h` also works on hypotheses, which {tactic}`rfl` can't touch.
+Rewriting by definitions also works in hypotheses, which {tactic}`rfl` can't touch.
```
xhalo32: Would "Rewriting can also be used in places where `rfl` can't -- like in hypotheses for instance" sound better?

LF/UsingLean.lean
```diff
@@ -397,7 +440,7 @@ example (n m : Nat) (h : 2 * n = m * 2) : n + n = m + m := by
 ```
 
 But {tactic}`rw` rewrites only one instance of a definition at a time.
-When a hypothesis mentions the same function at several different
+When a hypothesis or goal mentions the same function at several different
```
xhalo32: "mentions a function applied several times to different arguments" sounds better to me

xhalo32:
I like where this is going! Let's not forget to explain why one shouldn't rewrite definitions from other libraries, and that one should use the simplification/characterizing lemmas instead.

xhalo32 (approved):

LF/Poly.lean
```diff
@@ -862,98 +860,6 @@ checking your answers in Lean:
   print?
 :::::
 
-::::full
```
dsainati1, line 865: I moved most of this to `Logic` since it relies on `apply`, which students have not seen until `Tactics`

LF/UsingLean.lean
```diff
@@ -407,69 +449,72 @@ example (n m k : Nat) (h : square n + square m + square k = 0) :
...
 ```
-This is because `Nat.zero_mul` is true by induction, not reduction. This is a bit confusing
-to explain, and also unfortunate since one may not know why an equality holds. Thus I'd
-prefer not to include this use here.
-:::
 
-## A New Step Towards Automation
+The benefit that `dsimp only` provides is to help you understand the proof; Lean doesn't
+actually need it, because tactics like {tactic}`rw` do this automatically on the goal
+or on a hypothesis when you use them.
```
berberman: This feels like `dismp only` before `rw` is unnecessary. Consider this

```lean
example (n : Nat) : (fun x => x + 0) n = n := by
  dsimp only
  rw [Nat.add_zero]
```

`rw` would fail if the beta reduction was not performed.

dsainati1: Oh this is a much better example. I will rewrite this section.

LF/UsingLean.lean
```diff
@@ -324,71 +326,111 @@ uses, and both will be tools in your ever-growing toolbox of tactics.
...
+To proceed here, we need to reveal to Lean the underlying definitions of `addThrice` and `addTwice`,
+so that {tactic}`rw`, which only operates on syntax, can see the addition.
+We can do this by rewriting by those definitions:
 
 ```lean
-example (n m : Nat) (h : n + n = m) : triple n = m + n := by
-  -- `rw [triple]` unfolds `triple n`.
-  rw [triple, h]
+example (n : Nat) : addThrice n = n + addTwice n := by
+  -- `rw addThrice]` unfolds `addThrice`, replacing it with its definition
```
berberman:
```suggestion
  -- `rw [addThrice]` unfolds `addThrice`, replacing it with its definition
```

berberman (approved):

LF/UsingLean.lean
```diff
@@ -407,69 +449,75 @@ example (n m k : Nat) (h : square n + square m + square k = 0) :
...
-```
+## Definitional Simplification
 
-```leanOutput rwNotDone
-unsolved goals
-n✝ n : Nat
-⊢ n * n + 0 = n * n
-```
+::::full
+When dealing with unfold definitions, you may often encounter situations in which
```
mwhicks1: Proposed rewrite for all of this nearby:

Sometimes when you unfold a definition your hypothesis or goal may become hard to understand. When that happens, it can be useful to simplify it. To apply simplifications similar to those that `rfl` does, but without also checking reflexivity, you can use the tactic 'dsimp only' or 'dsimp only at h`.

mwhicks1 (approved):

PR #323, opened 2026-08-26
Make quiz options in Logic uniform

ionathanch:


LF/Logic.lean
```diff
@@ -1060,15 +1010,16 @@ besides {tactic}`intro`, {tactic}`apply`, and {tactic}`exact`?
...
 3. {tactic}`left` and/or {tactic}`right`
-4. none of the above
+4. {tactic}`cases` and {tactic}`left` and/or {tactic}`right`
+5. none of the above
 
 :::quizSolution
 ```lean
 example (a : Prop) (h : 1 = 0) : (a ∨ ¬ a) := by
-  contradiction
+  cases h
```
berberman: I think we didn't explicitly mention that `cases` can do a lot on equalities, but only vaguely said something like it works as `contradiction` in this chapter. Maybe we should adapt the following explanation of `inversion` in IndProp and then move it to this chapter:
https://github.com/plclub/sf-in-lean/blob/7b89b69a3d61e14875670e4fa97c8c144f5946e0/LF/IndProp.lean#L1556-L1571

Or to Tactics, because that's where we explained constructor disjointness and injectivity.

ionathanch: I don't think this is the right place to discuss this so I'm going to just change the `cases` back to `contradiction` (which is what the dev note said to do anyway, oops)

berberman (approved):

PR #325, opened 2026-08-27
Typeclasses: dev note about API and encapsulation

xhalo32:


LF/Typeclasses.lean
```diff
@@ -609,6 +609,35 @@ Taking suggestions for additional simple group theory theorems to prove here.
...
+  - Mention `public`, `private` keywords and that we don't use them on the course?
+  - One can mostly ignore proof terms due to proof irrelevance
+
+Here is an example where the proof term is blocking a rewrite.
+The solution is to simplify it away.
+
+```lean +error
+-- set_option pp.proofs true in
+example {n m : Nat} {a : Fin n} {b : Fin m} (h₁ : n = m) (h₂ : a.val = b.val) :
+    a = ⟨b.val, h₁ ▸ b.isLt⟩ := by
```
dsainati1, line 634: I think this example probably would need to change because this makes use of a couple things we have not explained, but we can worry about that later

dsainati1 (approved):

LF/Typeclasses.lean
xhalo32: It's just there to demonstrate that the proof term can cause problems even though we have proof irrelevance

dsainati1: Oh it's not meant to become user-facing, I see.

PR #330, opened 2026-08-27
Basics: Tweaks to terse; exercise naming typo

mwhicks1:
I updated the full/terse breakdown a little bit so more is shown in terse. The main thing missing was `structure` explanation.
I also noticed that two exercises were given the same name, so I fixed the second one's name to be different.

### DIFF
--- a/LF/Basics.lean
+++ b/LF/Basics.lean
@@ -396,7 +396,7 @@ example : (!MyBool.false) = MyBool.true := by rfl
 ```
 
 ::::full
-The technical details of how these symbolic notations work are not something you need to understand until quite a bit later in your Lean journey.  
+The technical details of how these symbolic notations work are not something you need to understand until quite a bit later in your Lean journey.
 
 :::details
 Lean has a very flexible notation system. Operators like `||` and `&&`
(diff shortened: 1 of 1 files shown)


dsainati1 (approved):

bcpierce00 (approved):

LF/Basics.lean
```diff
@@ -1916,7 +1932,7 @@ We can rewrite with these rules to expand numerals into their definitions,
 which allows us to use our {name}`add` rules.
 Here's an example of how to start a proof this way.
 
-::::exercise (rating := 1) (name := "mul_simpl_rules")
+::::exercise (rating := 1) (name := "nat_eq_rules")
```
bcpierce00, line 1935: This will mess up the autograder if a new version of Basics gets released before everybody has handed in their assignments (which *will* happen, because it the Induction chapter will need to be released before everybody has turned in this chapter).

bcpierce00 (approved):
Looks good except the exercise rename...

LF/Basics.lean
berberman: I think this name is just for display and doesn't interact with the autograder?

bcpierce00:
Ah, then it's fine!

LF/Basics.lean
xhalo32: The autograder completely ignores the `(name := ...)`. I still haven't gotten around to document this clearly...

bcpierce00:
Whew.

PR #335, opened 2026-08-28
Add devcontainer configuration

mrigankpawagi:


mwhicks1 (commented):
I'm not familiar with devcontainers in GitHub repositories. Can you explain what this will do, if we merge it? Will it cause everyone to be running within a Docker container or is that somehow locally configurable?

mrigankpawagi:
@mwhicks1 `.devcontainer/` is only used by GitHub codespaces and will not affect people working on their own machines. When a codespace is launched for the repo, the Dockerfile will be used to configure the codespaces environment.

mwhicks1 (approved):

PR #336, opened 2026-08-28
Induction: clean up resolved developer notes

Robertboy18:
Following the discussion about reviewing Basics and Induction, I went through the remaining developer notes in `LF/Induction.lean`. This removes stale planning notes, notes superseded by the current display conventions, and the hidden `cases`/`induction` quiz. It also makes a few small reader-facing wording fixes, including clarifying that the review question uses the chapter's custom `add`.

I checked the `cases` examples under Lean 4.34.0-rc2. The distinction is real, but it is too advanced for this chapter. Following review, this PR leaves the examples out of Induction; their possible placement in Logic or IndProp can be handled in a focused follow-up.

### DIFF
--- a/LF/Induction.lean
+++ b/LF/Induction.lean
@@ -13,28 +13,6 @@ htmlSplit := .never
 file := some "Induction"
 %%%
 
-:::dev
-```
-SOONER: We should also consider adding more examples to clarify
-the concepts introduced in this chapter. This could help in
-reinforcing the understanding of induction principles.
-
-LATER: In 3/22, MRC and BCP discussed "inlining" IndPrinciples
-into earlier chapters, thus eliminating it as a chapter. This
-chapter, Induction, is the first place a change would occur.  We
-would present [nat_ind] here. Then in Lists/Poly we'd present
-[list_ind], and the rest would go in IndProp and ProofObjects. The
-main wrinkle is that we'd need to introduce [apply] here instead of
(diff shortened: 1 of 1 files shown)


LF/Induction.lean
```diff
@@ -595,68 +560,6 @@ theorem double_add (n : Nat) : double n = n + n := by
 :::
 :::::
 
-:::dev "Yipeng Liu (berberman)" PotentialImprovement
```
Robertboy18, line 598: I checked these examples with Lean 4.34.0-rc2, and the distinction is real. It feels too advanced for this chapter, though. Would it be better to move this discussion to IndProp?

berberman: I think we could move these examples to Logic -- that's where we explain `injection` and `contradiction`. My opinion is to teach `cases` can do these on equalities in Logic, and then in IndProp we compare it with `inversion`. Or just completely move them to IndProp.

dsainati1 (approved):

LF/Induction.lean
Robertboy18: That makes sense. I’ll keep these examples out of Induction in this PR. We can decide in a focused follow-up whether to introduce them in Logic and compare with inversion in IndProp, or move the whole discussion to IndProp.

mwhicks1 (approved):

PR #337, opened 2026-08-28
Imp: improve delabrator and reduce antiquotation noise

berberman:
This PR combines the nonrecursive delabration fallback from #256 with the current behavior from #231. So

```lean
#check
  Com.cond (Bexp.bool b)
    (.cond (Bexp.bool true)
      (.asgn y (.num <| 3 + 4))
      (.asgn (if b then x else y) a)
    ) (.skip)
```

now prints

```lean
info: imp {
  if (~(Bexp.bool b)) {
    if (true) {
      y := ~(Aexp.num (3 + 4))
    } else {
      ~(Com.asgn (if b = true then x else y) a)
    }
  } else {
    skip
  }
} : Com
```

instead of a recursive delabration failure.

Lean variables now can appear in Imp syntax without `~`. For example, given `(b : Bool) (c : Com)`:

```lean
imp {X := a}
imp { if (b) {c} else {skip} }
```

works without writing `~a`/`~b`/`~c`. Identifiers from `Aexp` are resolved by type -- `x : Indet` becomes `Aepx.id x` and `x : Aexp` is inserted directly. Delaborator always prints out `~` for round-tripping; and arithmetic patterns such as `imp { x := ~a }` still keep `~` to be explicit about what they are binding.

Delaborators for `dcom` and `sparse_dcom` are also added, with some pretty printing rules changed.

### DIFF
--- a/HL/Hoare.lean
+++ b/HL/Hoare.lean
@@ -567,66 +567,59 @@ This has the downside that assertions need to be wrapped in `()` parentheses whe
 ```lean
 namespace Assertion
 
-section
-open Lean Elab Term
+open Lean Elab Term Meta Imp.Elab
 
-scoped syntax:max (name := assn) "assn(" ident "; " term ")" : term
+scoped syntax:max "assn(" ident "; " term ")" : term
 scoped syntax:lead "{{" term "}}" : term
-
-@[term_elab assn]
-def assnElab : TermElab := fun stx _type? => do
-  match stx with
-  | `(assn($st; $t:term)) =>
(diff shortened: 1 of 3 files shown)


xhalo32 (approved):
This works really well! It appears to also solve the Aexp delaboration issue where `ha` would have a nonsensical type in the following example:
```
example {a b : Aexp} (ha : a = b) : True := by
  induction a with
  | num n =>
    -- here `ha : n = b`
```

dsainati1 (approved):
Nice!

PR #338, opened 2026-08-28
Preface: Updated installation instructions

mwhicks1:
I added information from ALPHA-TESTERS.md to the Preface, explaining how to download and build the book. I also added information for students taking a class: Download the prepared .zip file and use that.

### DIFF
--- a/LF/Preface.lean
+++ b/LF/Preface.lean
@@ -289,12 +289,18 @@ updates. More on exercises below.
 
 If you are *reading Software Foundations on your own*, you can get the most
 up-to-date version from the [SF-in-Lean](https://github.com/plclub/sf-in-lean) GitHub
-repository. Clone that repository and then build it by typing `make` from the root
-directory. This will construct various versions of the book.
+repository. Clone that repository and then build it by typing `make lf-student` from
+the root directory. Doing so will construct the student version (full prose, with solutions elided) of the _Logical Foundations_ book.
 
-Assuming you want to start from the beginning, we recommend typing `make lf-student`
-instead. This builds the _Logical Foundations_ volume in its student form
-(full prose, with solutions elided) and writes two things to `_out/lf/student/`:
+Building the book requires that you have Lean installed. If you do not, follow the
+instructions [here](https://lean-lang.org/install/manual/) to install the Lean toolchain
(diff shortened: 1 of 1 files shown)


LF/Preface.lean
```diff
@@ -265,31 +265,72 @@ follow these steps:
...
+Note that, as the book is changing while you are taking your class, you should download
+a fresh `.zip` for each homework you do, opening it in a fresh directory. This way
+you will have access to prior solutions, and you will automatically get any Lean
+updates. More on exercises below.
+
+### Downloading and building the book from Git, for self study
+
+If you are *reading Software Foundations on your own*, you can get the most
+up-to-date version from the [SF-in-Lean](https://github.com/plclub/sf-in-lean) GitHub
+repository. Clone that repository and then build it by typing `make` from the root
```
bcpierce00: Not sure this step is correct... If people don't have Lean installed locally, the `make` won't work.  They need to do it from inside VS Code.  No?

LF/Preface.lean
```diff
@@ -265,31 +265,72 @@ follow these steps:
...
+here, but keep anything you want to survive somewhere else.
+
+If you would like to read the book HTML, start a local HTTP server and point it
+at the generated HTML files:
+
+  ```display
+  python3 -m http.server 8000 -d _out/lf/student/html
+  ```
+
+Then visit `http://localhost:8000` and start reading.
```
bcpierce00, line 339: Should we remove those instructions from where they were before?

LF/Preface.lean
mwhicks1: Good point. But opening LF/Basics in the Git repo is not what they want, since it's the verso file. Should we explain how to install Lean from the command line?

LF/Preface.lean
mwhicks1: I don't follow. Why remove these?

mwhicks1: Oh maybe I know what you mean: They are currently in `ALPHA-TESTERS.md` so it's Ok to leave them. They do not show up in the book itself.

LF/Preface.lean
mwhicks1: I've updated the file to explain command-line installation, and adjusted to talk about an easy way of doing it from VS Code.

PR #339, opened 2026-08-28
Initial Port of MoreStlc

dsainati1:
This is the initial port of MoreStlc. The proofs are `sorry`d out. 

Also, the notation does not quite work 100%. If an expert like @xhalo32 or @berberman could take a look I would appreciate it.


TS/MoreStlc.lean
```diff
@@ -0,0 +1,2567 @@
...
+scoped syntax:60 " inl " stlcTy:60 stlcTm:60 : stlcTm
+scoped syntax:50 "case " stlcTm:50 " of " "inl" stlcVar " => " stlcTm:50 " | "
+  "inr" stlcVar " => " stlcTm:50 : stlcTm
+
+scoped syntax:60 " nil " stlcTy:60 : stlcTm
+scoped syntax:60 stlcTm:61 " :: " stlcTm:60 : stlcTm
+scoped syntax:50 "case " stlcTm:50 " of " "nil" " => " stlcTm:50 " | "
+  stlcVar " :: " stlcVar " => " stlcTm:50 : stlcTm
+
+scoped syntax:60 " ( " stlcTm:60 " , " stlcTm:60 " ) " : stlcTm
```
ionathanch:
```suggestion
scoped syntax:max " ( " stlcTm:60 " , " stlcTm:60 " ) " : stlcTm
```

TS/MoreStlc.lean
```diff
@@ -0,0 +1,2567 @@
...
+#check <{ x * y * z }>
+#check <{ succ (pred x) }>
+#check <{ succ x y }>
+#check <{ inr Nat (λx : Unit . x) }>
+#check <{ nil Nat }>
+#check <{ 3 :: nil Nat }>
+#check <{ (x , y) }>
+#check <{ fst x }>
+-- why does this one not work
+-- #check <{ fst (x , y) }>
```
ionathanch: Changing the priority of the `(x, y)` syntax to `max` should make this work now since function arguments are at max level so it only expects the regular parens `(t)` and not pairs (level too low)

ionathanch (changes requested):

TS/MoreStlc.lean
ionathanch: wait you didn't uncomment the test to make sure what I said was true facts

dsainati1: Yeah I am testing locally. It does fix this test but it also breaks parsing of real actual Lean function arguments in downstream files.

ionathanch: :weary:

dsainati1: It works if I use `⟨` and `⟩` for pairs instead of parens

dsainati1: Actually I figured out a better way. Disregard.

bcpierce00 (approved):

PR #343, opened 2026-08-30
Follow up on Basics cleanup

Robertboy18:
I went through Basics once more after #341 merged and found a few small factual and explanatory issues that were easy to resolve.

This:
- corrects the explanations of namespaces, exported `Bool` names, dotted notation, and termination checking
- fixes stale comments in `and_eq_or` and a couple of inferred binders
- resolves the tuple and Airport terminology developer notes

I left the `MyBool` / `NatPlayground.Nat` question and theorem-naming notes in place since those seem to need a broader design decision.

### DIFF
--- a/LF/Basics.lean
+++ b/LF/Basics.lean
@@ -31,7 +31,7 @@ If you don't have Lean installed yet:
 * Start a new terminal session to pick up environment changes.
 * Run `make`.
 * Run `make serve`. Navigate to "http://localhost:8000/lf/student/html/" to start reading.
-* Make a copy of "\_out/lf/student/lean" to start solving as if I were a student.
+* Make a copy of "\_out/lf/student/lean" to start solving as if you were a student.
 :::
 
 This chapter introduces some of Lean's most essential features for writing functional programs
@@ -780,7 +780,7 @@ E.g., these are valid constructor expressions...
 :::
 
 We can define functions on colors using pattern matching, just as
-we did for {name}`Day` and {name}`Bool`.
+we did for {name}`Day` and {name}`MyBool`.
 
 ```lean
 def monochrome (c : Color) : Bool :=
@@ -984,14 +984,14 @@ Playground.myBar : RGB
 ```
 
 ::::full
-When a new type is declared, a `namespace` with the same name is implicitly created as well;
-definitions on that type are available inside the `namespace` without a prefix. For example,
-we can use the `blue` constructor without qualification below because
+The constructors of an inductive type are placed in the `namespace` named after the type.
+When we enter that `namespace`, we can use those constructors without a prefix.
+For example, we can use the `blue` constructor without qualification below because
 we are inside the {name}`RGB` `namespace`.
 ::::
 
 ::::terse
-Type definitions implicitly create namespaces.
+An inductive type's constructors are placed in the `namespace` named after the type.
 ::::
 
 ```lean
@@ -1099,15 +1099,13 @@ Unknown identifier `myHiddenDef`
 ```
 
 ::::full
-In fact, this is exactly what Lean does with the standard {name}`Bool` type by default.
-Since it is an important
-part of many proofs and programs, Lean implicitly `open`s many of `Bool`'s functions and
-constructors. This means we can use constructors like {name}`true` and {name}`false` and functions
-like {name}`not` without qualifying them with {name}`Bool`.
+In fact, Lean's prelude explicitly exports commonly used names from the standard
+{name}`Bool` `namespace`. This means we can use constructors like {name}`true` and
+{name}`false` and functions like {name}`not` without qualifying them with {name}`Bool`.
 ::::
 
 ::::terse
-Names from the `Bool` `namespace` are `open`ed and thus available without qualification.
+Common names from the `Bool` `namespace` are exported and available without qualification.
 ::::
 
 ```lean (name := tt)
@@ -1124,25 +1122,20 @@ Bool.true : Bool
 ```
 
 ::::full
-Finally, Lean can often automatically figure out which namespace a qualified name lives in,
-saving us the need to explicitly specify it every time we use the name. Instead of
-the fully qualified style (e.g., {name}`Day.monday`), we can opt for an implicitly qualified style,
-writing just `.monday`.
+Finally, Lean can often use an expression's expected type to fill in the missing prefix
+of a name that begins with `.`. Instead of the fully qualified style
+(e.g., {name}`Day.monday`), we can write just `.monday`.
 
-Lean tries to resolve `.monday` by checking what type is expected by the context in which this expression appears
-and inferring the namespace from that type. If there is only one such
-namespace (i.e., if it is unambiguous which constructor we're referring to), then it will
-automatically resolve to the expected value.
-:::dev "Benjamin Pierce (bcpierce00)"
-The wording is potentially confusing: From what we've said, readers might infer that every type has an associated namespace, so if we know the expected type then the namespace will always be unambiguous...
-:::
+For example, when the expected type is {name}`Day`, Lean resolves `.monday` as
+{name}`Day.monday`. If the context does not determine an expected type, dotted notation
+may be ambiguous.
 
 So, for example, we can also write {name}`nextWorkingDay` like this, using the shorter
 style for both the value being matched and the value being returned:
 ::::
 
 ::::terse
-Lean can often guess which qualified name we mean if we don't supply it explicitly:
+Lean can often use the expected type to resolve a name beginning with `.`:
 ::::
 
 ```lean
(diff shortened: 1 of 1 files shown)


LF/Basics.lean
```diff
@@ -984,14 +984,14 @@ Playground.myBar : RGB
...
-we can use the `blue` constructor without qualification below because
+The constructors of an inductive type are placed in the `namespace` named after the type.
+When we enter that `namespace`, we can use those constructors without a prefix.
+For example, we can use the `blue` constructor without qualification below because
 we are inside the {name}`RGB` `namespace`.
 ::::
 
 ::::terse
-Type definitions implicitly create namespaces.
+An inductive type's constructors are placed in the `namespace` named after the type.
```
mwhicks1: Is the implication of your change that a namespace always exists in some sense, and it's only relevant when you define something within it? Thus, "placed in the namespace" would be accurate. This take would not be immediately intuitive to me as a student, I don't imagine. Same comment on the changes for 987.

LF/Basics.lean
```diff
@@ -1099,15 +1099,13 @@ Unknown identifier `myHiddenDef`
 ```
 
 ::::full
-In fact, this is exactly what Lean does with the standard {name}`Bool` type by default.
-Since it is an important
-part of many proofs and programs, Lean implicitly `open`s many of `Bool`'s functions and
-constructors. This means we can use constructors like {name}`true` and {name}`false` and functions
-like {name}`not` without qualifying them with {name}`Bool`.
+In fact, Lean's prelude explicitly exports commonly used names from the standard
```
mwhicks1: We have not used the term "export" above. We are speaking about `open` "bringing in" particular definitions (line 1072). Does Lean use the term "export"? If so, please update the text further up to use that term too. If not, please adjust this text.

LF/Basics.lean
```diff
@@ -1124,25 +1122,20 @@ Bool.true : Bool
...
-Lean tries to resolve `.monday` by checking what type is expected by the context in which this expression appears
-and inferring the namespace from that type. If there is only one such
-namespace (i.e., if it is unambiguous which constructor we're referring to), then it will
-automatically resolve to the expected value.
-:::dev "Benjamin Pierce (bcpierce00)"
-The wording is potentially confusing: From what we've said, readers might infer that every type has an associated namespace, so if we know the expected type then the namespace will always be unambiguous...
-:::
+Finally, Lean can often use an expression's expected type to fill in the missing prefix
+of a name that begins with `.`. Instead of the fully qualified style
+(e.g., {name}`Day.monday`), we can write just `.monday`.
```
mwhicks1: Change to 

So, instead of the fully qualified style {name}`Day.monday`, we can write just `.monday`.

LF/Basics.lean
```diff
@@ -1124,25 +1122,20 @@ Bool.true : Bool
...
+
+For example, when the expected type is {name}`Day`, Lean resolves `.monday` as
+{name}`Day.monday`. If the context does not determine an expected type, dotted notation
+may be ambiguous.
```
mwhicks1: And then what happens ... ?

LF/Basics.lean
```diff
@@ -1221,24 +1214,21 @@ GRADE_MANUAL 1: custom_namespace_checks
 :::
 :::::
 
-## Constructors with Multiple Parameters (Tuple Types)
+## Constructors with Multiple Arguments
```
mwhicks1: Why did you change this?

Parameter are binders in a declaration, and arguments are values at a use. So the text below should use the name _parameter_ in most (all?) places.

LF/Basics.lean
```diff
@@ -1221,24 +1214,21 @@ GRADE_MANUAL 1: custom_namespace_checks
...
 
 ```lean
 namespace Playground
 ```
 
 ::::full
-A given constructor of an inductive type can have multiple parameters,
-not just zero or one. This feature is one way to define _tuple types_ in Lean.
-:::dev "Benjamin Pierce (bcpierce00)"
-Is there another way to define tuple types?  (The discussion below is about an analog of record types, not tuples.)
```
mwhicks1, line 1234: I think that Benjamin's comment was not suggesting you get rid of reference to Tuple types, but that you get rid of "one way to" since the alternative (structures) requires you have named parameters. Please restore the mentions of tuple types.

LF/Basics.lean
```diff
@@ -1391,10 +1381,10 @@ namespace NatPlayground
...
 
 ::::full
-All the types we have defined so far — both enumerated types
-such as {name}`Day`, {name}`MyBool`, and {name}`Playground.Bit`, and tuple types such as
-{name}`Playground.Nibble` built from them — are finite. The natural numbers, on
-the other hand, are an infinite set, so we'll need to use a
+All the types we have defined so far — both enumerated types such as {name}`Day`,
+{name}`MyBool`, and {name}`Playground.Bit`, and types carrying a fixed amount of data
+such as {name}`Playground.Nibble` — are finite. The natural numbers, on the other hand,
+are an infinite set, so we'll need to use a
```
mwhicks1: In general, restoring tuple types probably means you just revert these (and subsequent) text edits.

LF/Basics.lean
```diff
@@ -2046,9 +2035,9 @@ theorem two_mul_two : (two * two : Nat) = four := by
 ### Equality and Ordering
 
 ::::full
-When we say that Lean relies on almost nothing that's truly built-in, we really mean it: even
-testing equality is not a primitive operation, but an ordinary function that we could reimplement
-ourselves as users.
+When we say that Lean relies on almost nothing that's truly built-in, we really mean it:
+even a Boolean equality test is not a primitive operation, but an ordinary function that
```
mwhicks1: Change "Boolean equality test" to "testing equality" since I think it could be confusing given our own introduction of `MyBool` vs. the built-in `Bool`

LF/Basics.lean
```diff
@@ -2516,21 +2505,20 @@ def even' (n : Nat) : Bool :=
   | succ (succ n') => even' n'
 ```
 
-When Lean checks this definition, it verifies that the recursion
-terminates.  Specifically, it checks that one of the parameters
-is _structurally decreasing_ — each recursive call made in the body of the
-definition is made on an argument that is smaller than the original input.
+When Lean checks this definition, it verifies that the recursion terminates.
+For this definition, it does so by checking that the recursive argument is
```
mwhicks1: Change back to "Specifically" -- the first sentence already refers to "this definition".

LF/Basics.lean
```diff
@@ -2751,10 +2739,6 @@ theorem and_eq_or (b c : Bool) : (b && c) = (b || c) → b = c := by
 :::suppressPreviousHeaderWhenTerse
 :::
 
-:::dev "Benjamin Pierce (bcpierce00)"
-In this section, the same artifact is called a simplification rule, a rewrite rule, and a characterization lemma.  I think we're using "simplification rule" here.
-:::
```
mwhicks1, line 2756: We should never use "rewrite rule". But have said "characterization lemma" and "simplification rule" throughout.

mwhicks1 (commented):
Thanks this pass. I am not comfortable with all of the text changes, so some requests here to make changes and/or revert.

LF/Basics.lean
chenson2018: There is a command with the literal syntax `export` that is being described here. It is correctly distinct here from `open`. The reason this paragraph is here is to answer why we can use `true` *without* having an `open Bool`. It is because there is a line `export Bool (false true)` in core Lean brining them into the root namespace.

mwhicks1: The above text never mentions `export` and only talks about explicitly including definitions in `open`, i.e., with
```
open MyOtherNamespace (myVisibleDef)
```
When the sentence starts with "In fact, Lean's prelude explicitly exports commonly used names ..." the reader is (or at least, I am) led to believe that it does so using the mechanism just shown. 

If we want to talk about `export` doing this thing, I think we should show an example like the one you've given. Or, at the least, we should be a little more careful with the text. Instead of writing "In fact, ..." to start off, you might say: "Lean also has an `export` command for selectively _exporting_ names (rather than selectively importing them with `open`). This is what Lean uses to ..." etc. explaining the `true` case.

chenson2018: I think the point of the way this paragraph was written (I think by @dsainati1?) is that we don't want an aside about `export` (it's very rarely used) but we do want to clear up why we can always just say `true` and `false`. If there's a way to write this without explaining `export`, I think that would be preferable. One way is to make the analogy that when Lean starts, it essentially does `open` for a few common types. This isn't literally true, but the end effect is the same. Would this be a good compromise?

mwhicks1: Maybe reword the paragraph, thus:

You might be wondering why you can use constructors like {name}`true` and {name}`false` and functions like {name}`not` without qualifying them with {name}`Bool`, and without explicitly `open`ing the `Bool` namespace. Lean provides a way to _export_ unprefixed names from a namespace, with the same effect as selectively `open`ing that namespace downstream, and the Lean prelude does that for commonly used names from the standard library. We don't explain this mechanism here because it's rarely used.

Robertboy18:
Thanks for the careful review, and sorry I changed so much at once. I pushed a follow-up that restores the parameter and tuple terminology, uses the suggested `export` and dotted-notation wording, and keeps “characterization lemma” where intended while removing “rewrite rule” from Basics. I’ll keep the next PRs much more focused.

LF/Basics.lean
Robertboy18: I used Mike’s latest wording here. It distinguishes `export` from `open` without introducing the command syntax.

LF/Basics.lean
Robertboy18: You’re right. I restored the parameter and tuple terminology and removed only “one way to,” which was the part raised by the developer note.

mwhicks1 (approved):

PR #344, opened 2026-08-30
Polishing pass on Induction

bcpierce00:
... and some more cross-cutting formatting work.  The most significant is that I replaced most of the blank lines between -- comments with lines containing just --.  This reads significantly better for me.

I'm done with this chapter.  @Robertboy18 there are a few things for you to look at,


LF/Induction.lean
KihongHeo, line 254: Do we have a good solution to use the `lean` tag for this block? `-keep` does not help ignore previous definitions. `+error` would be misleading.

KihongHeo, line 548: The hidden block below appears rather abruptly. I think it would be helpful to add a brief note indicating what kind of details are hidden there.

Also, the hidden blocks currently show only a triangle icon. How about adding a message such as “details are hidden” to all folded hidden blocks?

KihongHeo: My mistake. I overlooked that this is already explained in the Basics section.

.vscode/settings.json
```diff
@@ -7,11 +7,101 @@
...
     // Lean input abbreviation for the type-colon glyph ⦂ (U+2982) used in the
     // typing judgment `⊢ t ⦂ T` (see TS/Types.lean): type `\tc` then space.
-    // Mirrored into every generated student project by `vscodeSettingsJson`
-    // in SFLMeta/Save/Project.lean — keep the two in step.
+    // Exported to every generated student project (see `studentSettingKeys`
+    // in SFLMeta/Save/Project.lean); this file is the source of truth.
     "lean4.input.customTranslations": {
         "tc": "⦂"
     },
+    // Tint Lean comments brown. This file is the source: generated
```
mwhicks1, line 15: Nice! I was wondering if we could get the student Lean files to render more nicely.

LF/Induction.lean
mwhicks1: We should be using `recall` here, but when I do it doesn't work. I think it's a bug in the `recall` implementation; investigating now.

proofread/rounds/Induction-r01.diff
```diff
@@ -0,0 +1,191 @@
+--- LF/Induction.lean (before this round)
```
mwhicks1, line 1: I don't think we want this file and the next couple of files in the repo, do we? These seem like maybe your own proofreading infrastructure which uses Claude?

mwhicks1: Same with `PROOFREADING.md` and others.

mwhicks1 (approved):

LF/Induction.lean
mwhicks1: Actually, we cannot use `recall` because the theorem was proved before `add` was made irreducible. So I think that `display` is the best solution. I was able to recall the `add` definition above this, though, so that's now properly rendered.

mwhicks1:
I have handled all of the `:::dev` comments in this file, so @Robertboy18 there is nothing for you to do on this one. @bcpierce00 please see my comments about the proofreading stuff, which I would not have thought we'd include in the repo (perhaps I failed to notice when it showed up before).

proofread/rounds/Induction-r01.diff
mwhicks1: Update: We will leave this stuff here and others can use it. @bcpierce00 might add a little more to it to make the approach easier to follow.

LF/Induction.lean
```diff
@@ -908,17 +922,23 @@ example (n : Nat) : Nat.beq n n := by
...
-
+::::terse
+Let's do the proof!
+::::
+::::full
+One possible proof is the following.
+::::
 ```lean
 example (n : Nat) : Nat.beq n n := by
-  induction n with
```
Robertboy18, line 919: Could we keep this consistent with `beq_self` and write `(n == n) = true` here? As written, `Nat.beq n n` relies on Lean’s `Bool`-to-`Prop` coercion, which I don’t think we have introduced yet, and that feels like an unnecessary distraction in an example about code actions.

mwhicks1: Will do.

PR #345, opened 2026-08-31
Polish Types

KihongHeo:
This PR changes the following:
- use unicode subscript
- hide notation definition in detail blocks
- explicit type annotation for `value_is_nf'`

### DIFF
--- a/TS/Types.lean
+++ b/TS/Types.lean
@@ -304,18 +304,18 @@ Here is the single-step relation, informally.
 
 ```
                    -----------------------------                 (ifTrue)
-                   if true then t1 else t2 ⟶ t1
+                   if true then t₁ else t₂ ⟶ t₁
 
                    ------------------------------                (ifFalse)
-                   if false then t1 else t2 ⟶ t2
+                   if false then t₁ else t₂ ⟶ t₂
 
-                              t1 ⟶ t1'
+                              t₁ ⟶ t₁'
             -----------------------------------------------      (ifStep)
-            if t1 then t2 else t3 ⟶ if t1' then t2 else t3
+            if t₁ then t₂ else t₃ ⟶ if t₁' then t₂ else t₃
 
-                              t1 ⟶ t1'
+                              t₁ ⟶ t₁'
                          -------------------                     (succStep)
-                         succ t1 ⟶ succ t1'
+                         succ t₁ ⟶ succ t₁'
 
                            -----------                           (predZero)
                            pred 0 ⟶ 0
@@ -324,9 +324,9 @@ Here is the single-step relation, informally.
                         ------------------                       (predSucc)
                         pred (succ v) ⟶ v
 
-                              t1 ⟶ t1'
+                              t₁ ⟶ t₁'
                          -------------------                     (predStep)
-                         pred t1 ⟶ pred t1'
+                         pred t₁ ⟶ pred t₁'
 
                           ----------------                       (isZeroZero)
                           iszero 0 ⟶ true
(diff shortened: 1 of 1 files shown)


TS/Types.lean
```diff
@@ -809,20 +812,26 @@ local macro_rules
...
 
+```lean
 -- The actual definition, written in the notation above.
 inductive Tm.HasType : Tm → Ty → Prop where
   | tru : <{ ⊢ true ⦂ Bool }>
   | fls : <{ ⊢ false ⦂ Bool }>
-  | ite (t1 t2 t3 : Tm) (T : Ty)
-      (h1 : <{ ⊢ t1 ⦂ Bool }>) (h2 : <{ ⊢ t2 ⦂ T }>) (h3 : <{ ⊢ t3 ⦂ T }>) :
-      <{ ⊢ if t1 then t2 else t3 ⦂ T }>
+  | ite (t₁ t₂ t₃ : Tm) (T : Ty)
```
dsainati1: While we are changing the variable names in this file: types should use the greek letter tau, rather than a capital t, imo

KihongHeo: Fixed.

TS/Types.lean
```diff
@@ -921,41 +933,41 @@ understand the parts we've given of the informal proof in the following
 exercise before starting — this will save you a lot of time.)
 
 ```lean
-theorem progress (t : Tm) (T : Ty) (hT : <{ ⊢ t ⦂ T }>) : Tm.IsValue t ∨ ∃ t', t ⟶ t' := by
+theorem progress (t : Tm) (τ : Ty) (hT : <{ ⊢ t ⦂ T }>) : Tm.IsValue t ∨ ∃ t', t ⟶ t' := by
```
Robertboy18: I think this rename missed the type in `hT`: it still uses `T`, so Lean introduces a separate implicit `{T : Ty}` and leaves `τ` unused. Could this be `⦂ τ` here, and likewise in `preservation` below?

TS/Types.lean
```diff
@@ -809,20 +812,26 @@ local macro_rules
...
+      (h₁ : <{ ⊢ t₁ ⦂ Bool }>) (h₂ : <{ ⊢ t₂ ⦂ τ }>) (h₃ : <{ ⊢ t₃ ⦂ τ }>) :
+      <{ ⊢ if t₁ then t₂ else t₃ ⦂ τ }>
   | zero : <{ ⊢ 0 ⦂ Nat }>
-  | succ (t1 : Tm) (h : <{ ⊢ t1 ⦂ Nat }>) : <{ ⊢ succ t1 ⦂ Nat }>
-  | pred (t1 : Tm) (h : <{ ⊢ t1 ⦂ Nat }>) : <{ ⊢ pred t1 ⦂ Nat }>
-  | isZero (t1 : Tm) (h : <{ ⊢ t1 ⦂ Nat }>) : <{ ⊢ iszero t1 ⦂ Bool }>
+  | succ (t₁ : Tm) (h : <{ ⊢ t₁ ⦂ Nat }>) : <{ ⊢ succ t₁ ⦂ Nat }>
+  | pred (t₁ : Tm) (h : <{ ⊢ t₁ ⦂ Nat }>) : <{ ⊢ pred t₁ ⦂ Nat }>
+  | isZero (t₁ : Tm) (h : <{ ⊢ t₁ ⦂ Nat }>) : <{ ⊢ iszero t₁ ⦂ Bool }>
 end
```
Robertboy18: Since the matching `section` is inside the collapsed details block, this `end` is shown by itself in the main code block. Would it make sense to move the `end` into the second details block too?

TS/Types.lean
KihongHeo: Nice catch. fixed.

TS/Types.lean
KihongHeo: Yes. Fixed.

mwhicks1 (approved):

PR #347, opened 2026-08-31
Imp: proof polishing, remove EvalR unexpanders

xhalo32:
Cleans up some AI generated proofs and text.

I got until `no_whiles_eqv`.


HL/Imp.lean
```diff
@@ -1131,13 +1070,62 @@ example :
...
+  · sorry
+```
+
+However, it's possible to use `simp` as long as we have assigned `st'` ourselves:
+
+```lean
+example : ∅ =[ X := 2; Y := 3 ]=> {Y ↦ 3, X ↦ 2} := by
+  apply EvalR.seq (st' := {X ↦ 2})
+  · apply EvalR.asgn
+    simp
```
berberman, line 1254: Let's use `simp only [Com.evalR_eq]` instead of a non-terminal `simp`.

HL/Imp.lean
```diff
@@ -1198,9 +1186,9 @@ Is the following proposition provable?
 
 ```display
 ∀ (c₁ c₂ : Com) (st st' : State),
-  st =[ ~c₁ ~c₂ ]=> st' →
-  st =[ c₁ ]=> st →
-  st =[ c₂ ]=> st'
+  st =[ ~c₁; ~c₂ ]=> st' →
+  st =[ ~c₁ ]=> st →
+  st =[ ~c₂ ]=> st'
```
berberman: Those `~` are no longer needed.

HL/Hoare.lean
```diff
@@ -4742,10 +4734,9 @@ inductive Com.EvalR : Com → State → Result → Prop where
 instance : HasEval Com State Result where
   Eval := Com.EvalR
 
-@[app_unexpander Com.EvalR]
-def Com.unexpandEvalR : Lean.PrettyPrinter.Unexpander
-  | `($_ $c $st0 $st1) => ``($st0 =[ ~$c ]=> $st1)
-  | _ => throw ()
+@[simp]
+theorem Com.evalR_eq {c : Com} {st : State} {res : Result} :
+    EvalR c st res ↔ st =[ ~c ]=> res := by rfl
```
berberman, line 4742: This is fine, and we need to run `simp` if the internal structure gets reveled after `apply`. Alternately, we could make a set of theorems such as

```lean
theorem eval_skip :
    st =[ skip ]=> st :=
  EvalR.skip
```

and apply them instead of the constructors.

xhalo32: What about the induction principle for EvalR, do we want to make a custom one that uses the notation?

meluge: Isn't the `~(imp {…})` noise just a limitation of unexpanders, which only see the syntax and so can only put a term in the `imp_com` slot via ~? The =[ ]=> notation itself prints fine because `HasEval.Delab.delabTriple` looks at the delabed command and splices the inner `imp_com` when it is `imp { … }`. Maybe same can be done on `Com.EvalR` with `@[delab app.Com.EvalR]`?

xhalo32: This simp lemma is here mostly for cosmetic reasons. I don't think we should delaborate Com.EvalR as =[ ]=> just like Nat.add doesn't delaborate to the + syntax. When doing induction in =[ ]=> we lose the notation, and this is usually acceptable (one can use simp to bring back the notation if it's neede by another lemma that uses the notation).

meluge (approved):
I've added the two instructor notes on the macro design as #337 rewrote them on main.

HL/Imp.lean
meluge: Removed!

HL/Imp.lean
meluge: This is a terminal simp. I'm not sure if there was some change.

PR #353, opened 2026-09-01
UsingLean: Full pass

mwhicks1:
I've gone over the entirety of UsingLean, making several changes
- Adjusted section headings to mention introduced tactics; helps the TOC
- Put exercises in `:::full` blocks
- Added uses of `:::terse` where needed
- Reorganized a little content to hopefully make the narrative simpler
- Changed one exercise from optional to required

### DIFF
--- a/LF/UsingLean.lean
+++ b/LF/UsingLean.lean
@@ -207,7 +207,7 @@ this to be a slow and frustrating way to write proofs. Instead, we
 suggest figuring out what you would like your next step to be,
 conceptually, and then using {tactic}`rw?` to search for a theorem
 that implements it. If no such theorem exists, you may need to prove it
-yourself, or you may want to rethink your approach.
+yourself.
 
 ::::exercise (rating := 1) (name := "mul_three_beq")
 Prove the following theorems about {name}`Nat`s.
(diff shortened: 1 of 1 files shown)


LF/UsingLean.lean
```diff
@@ -203,9 +200,19 @@ example (n m k : Nat) :
...
 ```
 
+:::::full
+We strongly recommend against blindly using {tactic}`rw?` and
+accepting its suggestions without due consideration! You will find
+this to be a slow and frustrating way to write proofs. Instead, we
+suggest figuring out what you would like your next step to be,
+conceptually, and then using {tactic}`rw?` to search for a theorem
+that implements it. If no such theorem exists, you may need to prove it
+yourself, or you may want to rethink your approach.
```
dsainati1: I'd worry that saying "you may want to rethink your approach" if the rewrite readers need isn't in the standard library might be too discouraging? I feel like this tends to happen fairly often, so I wonder if this sets up too high an expectation that the theorems readers need will be available often.

LF/UsingLean.lean
```diff
@@ -388,11 +388,12 @@ n✝ n : Nat
 ⊢ addThrice n = n + addTwice n
 ```
 
-But again we encounter an error! The expression in which we are trying to rewrite
-{name}`Nat.add_assoc` isn't of the form `n + m + k`, so we can't proceed. What then, should we do?
-To proceed here, we need to reveal to Lean the underlying definitions of `addThrice` and `addTwice`,
-so that {tactic}`rw`, which only operates on syntax, can see the addition.
-We can do this by rewriting by those definitions:
+The reason is that the expression in which we are trying to rewrite
```
dsainati1, line 391: I think the part of the old explanation that pointed out that `rw` is operating syntactically is important here, it explains why you need to reveal the underlying structure of `addThrice` for it to work. Otherwise the new one reads better.

mwhicks1: > pointed out that `rw` is operating syntactically

That's fair. I was thinking that was obvious given that definitions are not automatically reduced. Do you have a suggested edit? Go ahead commit to the branch.

LF/UsingLean.lean
mwhicks1: Ok, fair enough, happy to back that out. Do you mind doing it (commit to this branch, and also see below)?

dsainati1 (approved):

PR #355, opened 2026-09-01
Add normalize tactic using `solve_by_elim`.

dsainati1:
This introduces a new automation tactic for use with small step evaluation derivations. It also teaches use of `apply_rules` to solve goals that are repeated applications of constructors. 

This will come especially in handy for the proofs in https://github.com/plclub/sf-in-lean/pull/349

### DIFF
--- a/TS/Smallstep.lean
+++ b/TS/Smallstep.lean
@@ -1,6 +1,7 @@
 import SFLMeta
 
 import TS.Slang
+import TS.AttributeDecls
 
 open Verso.Genre Manual
 open SFLMeta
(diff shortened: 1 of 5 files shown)


TS/Smallstep.lean
```diff
@@ -1861,3 +1862,126 @@ theorem compiler_is_correct (a : Aexp) :
...
+
+We can package all this up into a dedicated tactic for solving reduction sequences,
+which we'll call `normalize`:
+
+```lean
+syntax "normalize" " using " ident,+ : tactic
+
+macro_rules
+  | `(tactic| normalize using $xs,*) =>
+    `(tactic| repeat apply Multi.step <;> try solve_by_elim (maxDepth:=10) using $xs,*)
```
berberman: I think there are two design choices here. Firstly, `solve_by_elim` tries `constructor` by default if no rule applies. We might want to disable it with `(constructor := false)` and explicitly add the constructors we care about to the rules for our purpose.

Secondly, there's a consideration of whether to let `normalize` build `Multi`. I think the answer is yes, given the current implementation. Then we want to try `rfl` first, so that goals like `<{ x }> ⟶* <{ x }>`  can be closed. And we should also take care of generated subgoals (as `repeat` only works on the main goal).

We can do something like

```lean
macro_rules
  | `(tactic| normalize using $xs,*) =>
    `(tactic|
      first
      | rfl
      | (apply Multi.step
         · solve_by_elim (maxDepth := 10)
            (constructor := false) only using $xs,*
         · normalize using $xs,*))
```

and add `Tm.IsValue.abs Tm.IsValue.tru Tm.IsValue.fls` to `StlcEval`.

TS/Smallstep.lean
```diff
@@ -1861,3 +1862,126 @@ theorem compiler_is_correct (a : Aexp) :
...
+::::exercise (rating := 1) (name := "normalize_ex")
+Use the `normalize` tactic to prove the following. You will need to supply the
+term `e'` yourself.
+
+```lean
+theorem normalize_ex : exists e', (.p (.c 3) (.p (.c 2) (.c 1))) ⟶* e' ∧ IsValue e' := by
+  solution!
+    exists (.c 6); constructor
+    . normalize using SimpleArith
+    . constructor
```
berberman: Use `\.` instead of `.`

```suggestion
    · normalize using SimpleArith
    · constructor
```

dsainati1: We have been using regular `.` everywhere else so far, no?

berberman: I think we are mostly using `\.`, which is conventional. There are some `.` left in IndProp but I will change them. IIRC our style guide says using `·`(`\.`) to select a subgoal, but it doesn't clarify it's `\.` not `.`.

berberman (approved):

PR #360, opened 2026-09-02
Linter performance: pre-index each command's info trees

david-christiansen:
This PR replaces repeated `InfoTree` re-traversals in the SF linters with an explicit pass that produces an index of the relevant data. On `MoreSTLC`, this reduces lint time for me from 2.7s to 0.15s.


berberman (approved):
Thank you!

PR #375, opened 2026-09-08
Tactics: More tidying

mwhicks1:
- Update forward reasoning with `apply` to make sudden appearance of `rw` tactic flow better. Updated `apply at` custom tactic to help
- Removed bits I added to `cases`, realizing they did not make sense; moved back to Logic
- Showed Review in terse build
- Hide exercises from terse build
- Added “Additional Exercises” section, since final exercises were not associated with the final section
- A little bit of polishing text

### DIFF
--- a/LF/Tactics.lean
+++ b/LF/Tactics.lean
@@ -167,6 +167,7 @@ example (n m o : Nat)
     rw [h₁, h₃]
 ```
 
+::::::full
 :::::exercise (rating := 3) (name := "injection_ex3")
 ```lean
 theorem injection_ex3 {α : Type} (x y z : α) (l j : List α)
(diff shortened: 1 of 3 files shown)


LF/Tactics.lean
```diff
@@ -1208,25 +1188,33 @@ use forward reasoning.  In Lean, however, backward reasoning is often more
...
-particular proof style, and so the developers of the [Mathlib](https://github.com/leanprover-community/mathlib4) library
-defined the `apply ... at ...` tactic to
-better support forward reasoning. Mathlib is a very large development,
-so we will not import the whole thing here, but we have
-made `apply ... at ...` available because it is quite useful.
+Note that the {tactic}`apply` tactic, and indeed any tactic, can target
+the goal explicitly rather than implicitly:
+you can name it using the turnstile symbol `⊢`, written `\|-`, `\goal` or `\vdash`.
+Since the goal is exactly what ordinary {tactic}`apply` already works
+on, `apply t at ⊢` behaves exactly like `apply t`.
```
dsainati1: Not sure how I feel about this bit. Unless I am misunderstanding the change you made to CustomTactics, `apply t at ⊢` is not actually supported out of the box by Mathlib's `apply at`, right? If not, why are we creating a custom tactic just for this when it is no different than `apply`?

mwhicks1: Oh, I thought it _was_ supported by Mathlib's `apply` but you are right, it is not.

What was there, which was to switch to talking about `rw` abruptly, seems wrong too, though. I will cut it entirely and we can figure out where to put it later (maybe Logic)

dsainati1 (approved):

PR #382, opened 2026-09-08
Make build stamps reproducible

berberman:
This PR replaces per-volume timestamps with git HEAD commit's short hash with its UTC time as the build stamp, regardless of whether there are uncommitted changes in the workspace.


mwhicks1 (commented):
When I look at the built book it seems Ok. When I run `make release` and look at the released book HTML, I see timestamps like this:
```
Source 8a44be0fd99b5bf56eaf649c16c5ac9a66129462, committed 2026-09-08 23:15 UTC; uncommitted changes; built 2026-09-10 22:59 UTC
```
I don't know why it has the "uncommitted changes" part (there aren't any), and the corresponding "built" timestamp. How are these detected? Is there an artifact with the way releases are constructed?

bcpierce00:
Can we just always use the last commit signature?

berberman:
> I don't know why it has the "uncommitted changes" part (there aren't any), and the corresponding "built" timestamp. How are these detected? Is there an artifact with the way releases are constructed?

I'm not able to reproduce this behavior, and actually the stamp was computed before the release script started modifying chapter files. Maybe there were some overlooked files in your workspace?



> Can we just always use the last commit signature?

Yes, using only the commit hash seems to be a good trade-off. I think it's fine to not detect for uncommitted changes.

mwhicks1:
> Yes, using only the commit hash seems to be a good trade-off. I think it's fine to not detect for uncommitted changes.

I still would like to see the time of that commit (not worrying about uncommitted files, as you say). Will we see that too, or just the hash (which is less human-interpretable)?

berberman:
Now it shows `Source revision: 6dc98cd, committed 2026-09-17 20:33 UTC`

mwhicks1 (approved):

PR #387, opened 2026-09-10
Logic: Tidying and a few :::dev notes

bcpierce00:


LF/Logic.lean
mwhicks1, line 2159: Given the Claude comment, should we just remove `solution!` here? What happens in the Rocq version?

LF/Logic.lean
```diff
@@ -2213,15 +2217,12 @@ we can say either
 
 Again, these two notions are equivalent:
```
mwhicks1, line 2218: Why is there a colon at the end of this sentence?

LF/Logic.lean
mwhicks1, line 2284: Basically this is saying that computation is a good proof tactic. But this is a little confusing to me because we seem to want to eschew computation in favor of "simplification rules", which imply a preference for the Prop version, despite the downside shown here.

mwhicks1 (approved):

LF/Logic.lean
mwhicks1: Looking at the Rocq original, this should be an exercise. Made it so.

LF/Logic.lean
mwhicks1: Looks like the reason was a parenthetical that was dropped. Changed to a period.

LF/Logic.lean
mwhicks1: Rather than litigate here, I added this as a `:::dev` comment.

PR #390, opened 2026-09-14
Polishing imp

dsainati1:


HL/Imp.lean
```diff
@@ -919,8 +916,10 @@ imp {Z := X; Y := 1; while (Z ≠ 0) {Y := Y * Z; Z := Z - 1}}
 
 ## Desugaring Notations
 
-Even though the notations are useful for getting the high-level picture, it's sometimes helpful to turn off the notation to see the parsed structure as a plain term.
-This can be done with `set_option pp.notation false` (which we briefly mentioned in the Typeclasses chapter) as follows:
+Even though the notations are useful for getting the high-level picture,
+it's sometimes helpful to turn off the notation to see the parsed structure as a plain term.
+This can be done with `set_option pp.notation false`
+(which we briefly mentioned in the Typeclasses chapter) as follows:
```
mwhicks1, line 877: Add chapter reference to Typeclasses

HL/Imp.lean
```diff
@@ -1661,15 +1648,15 @@ Marc Bezem 2022:
...
+The following function yields {name}`true` just on programs with no while
 loops. Using `inductive`, write a property `Com.NoWhilesR` that holds
 exactly when `c` is while-free, then prove it equivalent to `Com.no_whiles`.
 
 ```lean
 def Com.no_whiles (c : Com) : Bool :=
   match c with
   | imp {skip} => true
-  | imp {_x := ~_a} => true
+  | imp {x := ~a} => true
```
mwhicks1, line 1681: Are we able to avoid such antiquoting now?

HL/Imp.lean
```diff
@@ -1992,6 +1965,10 @@ evaluating an expression might diverge, the short-circuiting `and`
 would _not_ be equivalent to the original, since it would make more
 programs terminate.)
 
+:::dev PotentialImprovement
+This exercise turned out to be easier than we intended!
```
mwhicks1, line 1991: Should we change the difficulty level?

mwhicks1 (approved):

HL/Imp.lean
dsainati1: This exercise is optional, so I don't know that it's super necessary, but we could adjust it down.

HL/Imp.lean
dsainati1: These ones appear to be necessary. Unsure why.

HL/Imp.lean
dsainati1: Our build infrastructure does not support cross-book references, apparently!

mwhicks1: Hm, maybe that's something we should try to fix ...

HL/Imp.lean
dsainati1: cc @berberman any idea?

HL/Imp.lean
dsainati1: https://github.com/plclub/sf-in-lean/issues/391

HL/Imp.lean
```diff
@@ -156,7 +151,30 @@ inductive Aexp where
...
+-- … then `Bexp V`, `Com V`, `abbrev State (V) [DecidableEq V] :=
+-- TotalMap V Nat`, and `[DecidableEq V]` wherever a lookup/update is
+-- performed.
+```
+:::
+
+The `Bexp` definition is unchanged, except that it now refers to the new {name}`Aexp`.
+=======
 The `Bexp` definition is unchanged, except that it now refers to the new `Aexp`.
+>>>>>>> 90ba887100aa05a137f9792c5a46c0a0147c6074
```
berberman, line 177: Heads up: an unresolved merge conflict gets introduced here

PR #397, opened 2026-09-15
Logic: better Nat.zero_or_succ

ChengZ3:
I think this makes more sense because the right-hand side is no longer always true, and it also matches the theorem name better.


mwhicks1 (approved):
Great improvement!

PR #399, opened 2026-09-17
Imp: Ported Rest and Polish Pass

meluge:
This is follow up to PR 347.

- The case study and the throw section is now ported.
- Dropped some stale dev notes and polished AI generated proofs.


mwhicks1 (approved):
I made a few tweaks, and then I did the work to make exercises visible/not as per the Rocq, and to add the grading annotations. I had Claude help with these things -- no text was changed, just metadata.

PR #401, opened 2026-09-17
IndProp polishing (mostly low level)

bcpierce00:


LF/IndProp.lean
```diff
@@ -1052,13 +1046,17 @@ example (n : Nat) (hEven : Even n) (h : n = 1) : true = false := by
 :::
 ::::
 
+:::dev "Claude"
+None of this quiz's options matches its solution. In the `succ_succ` case the hypothesis is `n + 2 = 1`, and `contradiction` alone does not close it (checked); `injection` is needed first, as the solution shows. So (C) is not sufficient as stated and (D) is the literally correct answer, which is presumably not intended. Options: add `injection` to option (C) or as a new option, or change the hypothesis to one that `cases` and `contradiction` refute directly.
+:::
+
```
mwhicks1: I prefer leaving the answer as (D) -- if we don't have "insufficient tactics" occasionally, it's just a waste to include it.

mwhicks1 (approved):

LF/IndProp.lean
mwhicks1: (fixed)

mwhicks1 (approved):

PR #402, opened 2026-09-17
Quick round of polishing on Automation.lean

bcpierce00:


LF/Automation.lean
```diff
@@ -54,22 +54,22 @@ theorem Perm3_In_old (α : Type) (x : α) (l₁ l₂ : List α)
...
 
 ::::full
-The {tactic}`lia` tactic implements a decision procedure for integer linear
-arithmetic, a subset of propositional logic and arithmetic. {tactic}`lia`
-is also a decision procedure for first-order logic.
-:::dev "@rogerburtonpatel"
+The {tactic}`lia` tactic implements a decision procedure for _linear integer
+arithmetic_: propositional formulas whose atoms are linear
+constraints over the natural numbers and integers.
+:::dev "Roger Burtonpatel (rogerburtonpatel)"
```
mwhicks1, line 66: We refer to first-order logic earlier, but I don't think we explain it there either. The only time we talk about it is in the introductory lecture slides, IIRC. Maybe Preface?

dsainati1 (approved):

mwhicks1 (approved):

LF/Automation.lean
mwhicks1: I'll address this in another PR.

PR #408, opened 2026-09-18
Fix confusing text in Tactics

dsainati1:
Student pointed out that the text here is misleading, given that the subsequent definition is `sorry`.


LF/Tactics.lean
```diff
@@ -640,8 +640,8 @@ let ⟨a, b⟩ := v
 
 ::::::full
 :::::exercise (rating := 3) (name := "zip_unzip'")
-Here is an implementation of the {name}`unzip` function from
-chapter {ref "Poly"}[Poly]:
+Recall the {name}`unzip` function from chapter {ref "Poly"}[Poly];
```
chenson2018, line 643: If I were copy pasting from `unzip` into `unzip'`, I would definitely forget to replace the recursive call in the second branch. 

It's a shame we have to copy over at all, I assume this is a grading/import limitation? If so, I'd at least leave a comment.

dsainati1: I believe based on https://github.com/plclub/sf-in-lean/pull/394 that direct use of Poly's `unzip` was removed for grading reasons. @ChengZ3 how much work would it be grading-wise to allow using the old function instead of renaming it?

mwhicks1 (approved):

LF/Tactics.lean
dsainati1: I am going to merge this for now to fix the narrative issue with the text, but I'll make a ::dev comment explaining the problem.

